PR #62, opened 2014-11-17
Add the Spring IO plugin to spring-session and spring-session-data-redis

wilkinsona:
This change paves the way for Spring Session's inclusion in Spring IO 1.1 (see spring-io/platform#58). 

Note that I've added the plugin to `spring-session-data-rest` which does not yet contain any code. I'm not sure what the plans are for this module, but validating its dependencies and checking that they're covered by the Platform seemed like the right thing to do.


rwinch:
@wilkinsona This is now merged into master

PR #99, opened 2015-01-09
Small improvement to WebSocket guide

rstoyanchev:
Simply move content around so that we first show what needs to be
done and then explain what is done internally. Arguably this is an
easier way to digest since the former is required knowledge while
the latter is optional.


rwinch:
Thanks for the PR This is now merged into master

PR #123, opened 2015-01-25
pump spring-data-redis version

danveloper:
App fails to start when specifying `@EnableRedisHttpSession` with a `ClassNotFoundException` looking for `RedisSentinelConfiguration`. Confirmed that pumping the spring-data-redis version resolves this issue.

Workaround:
> build.gradle
```groovy
  compile('org.springframework.session:spring-session-data-redis:1.0.0.RELEASE') {
    exclude module: 'spring-data-redis'
  }
  compile 'org.springframework.data:spring-data-redis:1.4.1.RELEASE'
```

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -10,6 +10,6 @@ jstlVersion=1.2.1
 seleniumVersion=2.44.0
 servletApiVersion=3.0.1
 spockVersion=0.7-groovy-2.0
-springDataRedisVersion=1.3.0.RELEASE
+springDataRedisVersion=1.4.1.RELEASE
 springSecurityVersion=4.0.0.RC1
 springVersion=4.1.3.RELEASE


rwinch:
# tl;dr

Thanks for reporting this and providing a PR! This is now merged into master.

You could also work around this issue using the following:

```groovy
compile 'org.springframework.session:spring-session-redis:1.0.0.RELEASE'
compile 'org.springframework.boot:spring-boot-starter-redis'
```

The advantage is that this allows Boot to manage as many of the dependency versions as possible

# Details of the Problem...

Digging into this a bit deeper so I could figure out how this issue got past our build...

## Version Mismatch

The issue happens because Spring Boot 1.2.1's [RedisAutoConfiguration](https://github.com/spring-projects/spring-boot/blob/v1.2.1.RELEASE/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/redis/RedisAutoConfiguration.java#L76) requires `RedisSentinelConfiguration` which is not available until Spring Data Redis 1.4.0

## Why does Boot allow a version mismatch?

Spring Boot is suppose to manage my versions for me. So how did this happen?

### Boot Gradle only manages unspecified versions

This happens because Spring Boot's Gradle plugin currently only manages unspecified versions (not transitive versions). That means when using Spring Boot 1.2.1 that a configuration like this:

```groovy
compile 'org.springframework.session:spring-session-data-redis:1.0.0.RELEASE'
```

will bring in Spring Data Redis 1.3.0 resolved from Spring Session Data Redis 1.0.0.RELEASE.

### Additional Workaround

You could also work around this issue using the following:

```groovy
compile 'org.springframework.session:spring-session-redis:1.0.0.RELEASE'
compile 'org.springframework.boot:spring-boot-starter-redis'
```

Instead of using spring-session-data-redis, we only include spring-session and allow Spring Boot's starter pom to bring in all our Redis dependencies. I created #125 to improve our sample and documentation.

### Spring Boot Gradle 1.3+ Adding Dependency Management

Spring Boot 1.3 is currently scheduled to include the dependency management plugin which will ensure transitive dependencies are also managed. [See spring-boot/issues/2133](https://github.com/spring-projects/spring-boot/issues/2133).

## How did it get past our Build?

So the question is..."How did this get past our build?"

### Samples use Spring Boot 1.10

We do have integration tests and a CI build that verify all of the Spring Session samples work. This includes a Spring Boot sample. However, the Spring Boot sample is on 1.10 version of Spring Boot. Since you are using [Spring Boot 1.2.1](https://twitter.com/danveloper/status/559439175737282560) with spring-session-data-redis you are getting conflicting versions.

### Spring IO Checks

We also certify that Spring Session works with Spring IO 1.1.1 versions (which means it should work with Spring Boot 1.2.1). More concretely, this means we re-run all the tests with the versions found in Spring IO 1.1.1. The reason this works is because the Spring IO plugin changes ALL of the versions to match the versions found in Spring IO. This means that even though we declare a dependency on version Spring Data Redis 1.3.0 within our build when we certify against Spring IO we ensure all the versions are Spring IO versions including Spring Data Redis. This means that our tests will pass.

# Changes in Spring Session?

So what changes should we make to Spring Session? Initially we used Spring Data Redis 1.3.0 as we viewed this as the minimum version of Spring Data that could be used. This ensured that our tests ran against Spring Data 1.3.0 and then re-ran (when using Spring IO Checks) against the latest version.

However, this does seem cause some headaches in transitive versions. It also does not make it easy to follow the best practice of being on the latest and greatest.

Ultimately, I think we should update all of our versions to match the Spring IO Versions #126 and then create additional testing to verify our minimum versions as described in #120

PR #136, opened 2015-02-02
Ensure SessionRepositoryFilter always sends cookie if its value changes

dsyer:
In some cases clients might also like to or need to get a Set-Cookie header on
every request. This is really useful if you want to implement a stateless
repository, for instance, because the session ID can change every time the
session is mutated.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,5 +7,6 @@ bin
 .settings
 .project
 target
+pom.xml
 out
 .springBeans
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -7,5 +7,6 @@ bin
 .settings
 .project
 target
+pom.xml
```
martin-g: why ignore the pom.xml

dsyer: Because this is a gradle build.

dsyer:
Actually I think always sending the cookie is probably not the best idea for the reasons you say. I was just throwing out ideas really - some stuff I was playing with and it seemed to work so I thought it would be good to discuss it here. The flag to always send cookies is a separate commit, so it ought to be easy to get rid of.

rwinch:
Thanks I'm closing this in favor of #154 since we do not want the always send cookie flag. I'll cherry-pick the commit and apply it to #154.

PR #158, opened 2015-02-23
Spring security sessionmanagement example

domdorn:
Sample showing Spring-Security Integration with XML Config and Session-Management enabled.


rwinch:
In my opinion, we don't want to have a sample for every permutation of Spring Security (or other projects). If we did this, we would have way too many samples.

Instead, I believe we need to highlight deltas in Spring Session. So the question I'd ask is "How does *Spring Session* use change between the basic Spring Security / Spring Session example and this sample?"

Thoughts?

domdorn:
I agree, maybe we should just call it security-xml .. people can use it as a bootstrap example or compare their existing XML setups and I can add corner cases from our prod-environment to it.

rwinch:
Thanks for the fast response. I like the idea of having a basic XML based example of using Spring Security and Spring Session. If you can update the sample to leverage only the basics of Spring Security and to rename the project to security-xml I can merge in the project.

rwinch:
Closing due to lack of feedback. Please reopen and resolve the conflicts if you are still interested in contributing

pivotal-issuemaster:
@domdorn Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-session&pullRequestId=158)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-session&pullRequestId=158) to manually synchronize the status of this Pull Request.

Issue #141, opened 2015-02-09
Secure flag of CookieHttpSessionStrategy 


cemo:
CookieHttpSessionStrategy is currently fixing secure flag of cookie by request. If user tries to redirect after a `https login page` -> `http page`, this will cause an unwanted surprise. 

Please provide an option to be set by user.

For those who are interested in this issue, @rwinch explained and gave some wonderful insights and please be careful such configurations at production. see SEC-2799 for more.

cemo:
There is another todo for `set domain` as well. Maybe a customizer for cookie would be ideal.

rwinch:
Thanks for the great request! I'm trying to aggregate all the feature requests for cookies, so we can solve this problem more holistically. Therefore, I'm closing this in favor of #299 which takes this feature into account.

rwinch:
This is now resolved via #299 See http://docs.spring.io/spring-session/docs/current-SNAPSHOT/reference/html5/guides/custom-cookie.html for details.

Issue #112, opened 2015-01-19
CookieHttpSessionStrategy should provide a possibility to further define the cookie


tux4ever:
At the moment it is only possible to configure the name of the cookie, but not further properties. e.g. context path. It would be great if the CookieHttpSessionStrategy.createSessionCookie is refactored to a new bean which could be replaced or the CookieHttpSessionStrategy should not be final.

Relates to #106

domdorn:
#155 is more or less a specific version of this... maybe it might be better to just make createSessionCookie protected and allow users to override it in a subclass?

tsachev:
:+1:. Either create `createSessionCookie` protected and `CookieHttpSessionStrategy` non final or allow customization of the cookie with something like CookieCustoizer.
Please advice which approach you think is better and maybe I could provide  a pull request for this feature.

tsachev:
#243 is also related. This seems to be common request.

praptin:
+1 I hope in the next release it will provide more customization like to set domain. For my own need I create my own implementation.

zxkane:
+1 I hope we can specify the `domain` to share it among all subdomains.

rwinch:
Thanks for the great request! I'm trying to aggregate all the feature requests for cookies, so we can solve this problem more holistically. Therefore, I'm closing this in favor of #299 which takes this feature into account.

rwinch:
This is now resolved via #299 See http://docs.spring.io/spring-session/docs/current-SNAPSHOT/reference/html5/guides/custom-cookie.html for details.

praptin:
Thanks Rob, when will the release version will be released?

rwinch:
@praptin We plan to release a milestone next week and go GA Q1 of next year.

Issue #106, opened 2015-01-14
Allow configuring cookie expiration in CookieHttpSessionStrategy


rwinch:
While a secure default is important (i.e. having a cookie expire when browser is closed), some applications may want to customize this. We should allow customizing the cookie expiration.

Relates to #112

rwinch:
I'm trying to aggregate all the feature requests for cookies, so we can solve this problem more holistically. Therefore, I'm closing this in favor of #299 which takes this feature into account.

rwinch:
This is now resolved via #299 See http://docs.spring.io/spring-session/docs/current-SNAPSHOT/reference/html5/guides/custom-cookie.html for details.

PR #162, opened 2015-02-24
Allow to customize the way the session cookie is created. Fixes #155 #141 #112 #106

### LINKED ISSUE #155: Allow overriding of cookie path and/or cookiePathGenerationStrategy in CookieHttpSessionStrategy
Allow to set the cookie path to a fixed value and/or allow to override the strategy on how to derive the cookie path.

domdorn:
This allows to customize all parameters of the cookie to be generated and also allows to provide a strategy to manually calculate the contextPath (in case it needs a more complex logic to calculate it then just setting it from the request or setting it to a static value). 

It includes test-cases and documentation in form of javadoc

### DIFF
--- a/spring-session/src/main/java/org/springframework/session/web/http/CookieHttpSessionStrategy.java
+++ b/spring-session/src/main/java/org/springframework/session/web/http/CookieHttpSessionStrategy.java
@@ -103,6 +103,7 @@ import org.springframework.session.Session;
  * attributes. An example is provided below:
  * </p>
  *
+ * <pre><code>
  * {@code
  *      HttpSessionManager sessionManager =
  *              (HttpSessionManager) req.getAttribute(HttpSessionManager.class.getName());
@@ -145,7 +146,7 @@ import org.springframework.session.Session;
  *      req.setAttribute("addAccountUrl", sessionManager.encodeURL(contextPath, newSessionAlias));
  *      req.setAttribute("accounts", accounts);
  * }
- *
+ * </code></pre>
  *
  * @since 1.0
  * @author Rob Winch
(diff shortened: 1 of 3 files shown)


cemo:
I need to customize secure flag of Cookie as well. 

+1

domdorn:
@rwinch is there anything missing preventing this to be merged? I would really like to see this in 1.0.1 as right now, we have to provide our own CookieHttpSessionStrategy just to override how certain things in the cookie are set. Thx

rwinch:
@domdorn Thanks for the quick responses. I am rather busy with getting Security 4.0 GA out right now so haven't had time to look at Spring Session. 

After Security 4.0 GA I will be focusing on a 1.0.1 release. We will make sure something gets in 1.0.1 to resolve your issue.

cbeattie-miovision:
I needed this as well. I needed to set the context path to a fixed value so I am glad to see that this is now possible in this PR. 

The `CookiePathCalculationStrategy` is a nice touch but I wonder if it could be made a little more broad (and thus more useful). Clearly the purpose of the `CookieHttpSessionStrategy` class is to manage HTTP sessions via cookies and therefore the `createSessionCookie()` method is of critical importance. Wouldn't it be better to use the strategy pattern on the creation of the cookie itself rather than on only the path? For example:
```
public interface CookieCreationStrategy {
    Cookie createSessionCookie(HttpServletRequest, Map<String, String> sessionIds);
}
```

The default implementation (ie. `DefaultCookieCreationStrategy()`) could be exactly the same code that is currently in `CookieHttpSessionStrategy.createSessionCookie()` but using a strategy pattern would allow one to customize the way the cookie is created including things like setting the cookie domain or setting a custom path. One complication would be that `createSessionCookie()` currently uses some member variables but that isn't insurmountable.

Just a suggestion...but this approach seems a bit more flexible.

OR. Another far simpler approach would be to remove the `final` qualifier from `CookieHttpSessionStrategy` and allow for inheritance (noting the somewhat questionable value of doing so it certainly would be pragmatic).

Youmoo:
:+1: 
I need to customize cookie.domain , to allow all sub domains share same sessions.

leonty:
Hi,

Any progress on this PR ? It's 8 months since it's been proposed while this functionality is required by many. I for instance have the same issue as @Youmoo.

rwinch:
Thanks for the great request! I'm trying to aggregate all the feature requests for cookies, so we can solve this problem more holistically. Therefore, I'm closing this in favor of #299 which takes this feature into account.

PR #197, opened 2015-04-21
Remove duplicate `var headers = {};`

izeye:


rwinch:
Thanks for the report and the PR! This is now merged into master

PR #211, opened 2015-05-24
Fix <type> to </type> in Maven section of guides

asaikali:
Cutting and pasting the maven dependencies from the guides was failing
due to an xml element that was not closed.


rwinch:

PR #215, opened 2015-05-28
KeepAliveCondition for SessionRepository implementations

plajko:
It makes possible to control the session keep-alive based on a condition. It can be useful for long-polling solutions, when you don't want the continuous polling to keep the session alive.

### DIFF
--- a/spring-session/src/main/java/org/springframework/session/MapSessionRepository.java
+++ b/spring-session/src/main/java/org/springframework/session/MapSessionRepository.java
@@ -16,6 +16,7 @@
 package org.springframework.session;
 
 import org.springframework.session.events.SessionDestroyedEvent;
+import org.springframework.session.keepalive.KeepAliveCondition;
 
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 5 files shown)


rwinch:
Thanks for the PR. Can you explain your use case in a little more detail? Also, can you explain why this behavior should change? How does the current behavior differ from the standard servlet containers behavior?

plajko:
We have a custom implementation for long-polling (http://www.pubnub.com/blog/http-long-polling). Also there is a requirement: the http session must expire, and the user must be logged out. The problem there was, the long polling kept the session alive. With this modification we can define a custom condition for the session manager. In our case this is a pattern matcher for the request url (Implementation included in the PR).

Standard servlet containers behavior cannot be modified to get this work and the solution IMHO is generic enough for being helpful in other cases (although I'm not aware of any :)).

rwinch:
Thanks again for the PR.

I'm not sure that something like this is something that belongs in the framework. It seems if this was something that users were struggling with this would be solved within the servlet specification. Additionally, it seems weird to me that if a client is making a request that the session last accessed time is not updated (despite that being the reason the attribute exists).

For this reason, I am closing this as won't fix. If others feel that there is a need please comment and we can reconsider.

PR #218, opened 2015-05-31
Make the Redis hash key prefix configurable

fstegmann:
The hash key prefix can now be configured via @EnableRedisHttpSession


rwinch:
Thanks for the PR! There is a variation of this merged into master that takes into account the find by username functionality too.

PR #220, opened 2015-06-04
Add projects subdomain in project page URL

ractive:


rwinch:

NOTE: The previous URL did work, but it required an extra redirect so I agree it is better to have this change.

PR #221, opened 2015-06-08
Fix Hazelcast Sample attributeValue -> attributeName

master-slave:
Just a typo fix, changing the ID attribute for Attribute Name input from attributeValue to attributeName


rwinch:
Thanks for the PR! This is now fixed in master

PR #225, opened 2015-06-23
CONFIG_NOTIFY_KEYSPACE_EVENTS is not being used in EnableRedisKeyspac…

ggtucker:
…eNotificationsInitializer

There's a duplicate of this field in org.springframework.session.data.redis.config.ConfigureNotifyKeyspaceEventsAction which is where it appears to be used.


rwinch:
Thanks for the report and the PR! this is now merged into master

PR #236, opened 2015-07-08
Use the latest version of the Spring IO Plugin

wilkinsona:
This change is necessary to allow Spring IO Platform 2.0 to remove the platform-versions properties file that was deprecated in 1.1. To allow Spring Session's Platform compliance to be verified as part of Spring IO Platform 2.0's build and release process, I'd like this to be included in Spring Session 1.0.x which is the version that's currently in Spring IO Platform 2.0. If you have any questions, please let me know.


rwinch:
Thanks for the PR. This is merged in via 141aba95262ffa365ddeeaf809d1b1b79e4f3946

Issue #257, opened 2015-08-04
Add SessionExpiredEvent and SessionDeletedEvent


manderson23:
I have a use case where I need to audit log in reaction to login/logout/session timeout etc.

Currently you do not know whether SessionDestroyedEvent was caused by a delete or expiry. Would you be open to extending SessionDestroyedEvent to support this? Two options I can think of are

1. Add a Cause enum to SessionDestroyedEvent with values DELETE/EXPIRED
2. Add SessionExpiredEvent that extends SessionDestroyedEvent

I'm implementing this myself for now and I'm happy to do so going forward but through I would raise it in case it was viewed as a useful feature to go into the library itself.

rwinch:
We would love to see a PR that provides this functionality! I think that option 2 sounds the best to me. If you can get a PR that does this, we will could include it in 1.1 (which I just started work on).

manderson23:
Option 2 is what I am currently implementing. I'll look at creating a PR.

manderson23:
@rwinch Do you need Redis running locally for the `./gradlew check` integration tests to pass?

rwinch:
Yes you do

PR #258, opened 2015-08-10
Add SessionExpiredEvent and SessionDeletedEvent

manderson23:
Implements #257 

Let me know if I've missed anything particularly documentation and tests.

### DIFF
--- a/docs/src/docs/asciidoc/index.adoc
+++ b/docs/src/docs/asciidoc/index.adoc
@@ -284,7 +284,7 @@ Instead, developers should prefer interacting with `SessionRepository` and `Sess
 
 `RedisOperationsSessionRepository` is a `SessionRepository` that is implemented using Spring Data's `RedisOperations`.
 In a web environment, this is typically used in combination with `SessionRepositoryFilter`.
-The implementation supports `SessionDestroyedEvent` and `SessionExpiredEvent` through `SessionMessageListener`.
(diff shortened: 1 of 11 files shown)


rwinch:
Thanks for the PR! It seems we should come up with a similar event for deleting sessions (i.e. SessionDeletedEvent). Thoughts?

manderson23:
So SessionDestroyedEvent would be a base class (abstract?) with specific expired and deleted subclasses?

rwinch:
@manderson23 That is correct. This would allow receiving either specific event. To me it seems if we allow receiving only expired events users should also be able to only receive deleted events.

manderson23:
Sounds reasonable. Shouldn't be much more work.

manderson23:
Further doc updates still to follow.

manderson23:
@rwinch that should be SessionDeletedEvent complete. Let me know if there are any issues.

rwinch:
Thanks for the PR! This is merged into master and will be a part of 1.1. I did some very minor polish in https://github.com/spring-projects/spring-session/commit/7518306975f99e5d3095fc6ab6b19aefd0f790b8 but other than that it looked good.

You will note I have also added a `SessionCreatedEvent` #261 to support `HttpSessionListener` #4. I had to do some merging to get that to work with your changes. So you may want to give the snapshot a try and see if it is working for you to double check me :)

Issue #249, opened 2015-07-24
Add SessionContextHolder


rwinch:
Add the Session to a SessionContextHolder similar to SecurityContextHolder. The SessionContextHolder would store the Session in a ThreadLocal. This will allow users to access the current Session anywhere within their stack. For example:

```java
Session session = SessionContextHolder.getSession();
```

PR #267, opened 2015-08-14
Add SessionContextHolder

franciscospaeth:
Implemented SessionContextHolder;
Introduced SessionContextHolderStrategy;
Implementation of three modes: 
* MODE_THREADLOCAL;
* MODE_INHERITABLETHREADLOCAL;
* MODE_GLOBAL. (?)
Added necessary documentation.

Implements gh-249

### DIFF
--- a/spring-session/src/main/java/org/springframework/session/context/SessionContextHolder.java
+++ b/spring-session/src/main/java/org/springframework/session/context/SessionContextHolder.java
@@ -1,0 +1,162 @@
+/*
+ * Copyright 2002-2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.session.context;
+
+import java.lang.reflect.Constructor;
+
+import org.springframework.session.Session;
+import org.springframework.util.Assert;
+import org.springframework.util.ReflectionUtils;
+import org.springframework.util.StringUtils;
+
+/**
+ * Associates a given {@link SessionContext} with the current execution using configured strategy.
+ * <p>
+ * This class provides a series of static methods that delegate to an instance of {@link SessionContextHolderStrategy}. The purpose of the class is to provide a
+ * convenient way to specify the strategy that should be used for a given JVM. This is a JVM-wide setting, since everything in this class is <code>static</code>
+ * to facilitate ease of use in calling code.
+ * <p>
+ * To specify which strategy should be used, you must provide a mode setting. A mode setting is one of the three valid <code>MODE_</code> settings defined as
+ * <code>static final</code> fields, or a fully qualified classname to a concrete implementation of {@link SessionContextHolderStrategy} that provides a public
+ * no-argument constructor.
+ * <p>
+ * There are two ways to specify the desired strategy mode <code>String</code>. The first is to specify it via the system property keyed on
+ * {@link #SYSTEM_PROPERTY}. The second is to call {@link #setStrategyName(String)} before using the class. If neither approach is used, the class will default
+ * to using {@link #MODE_THREADLOCAL}, which is backwards compatible, has fewer JVM incompatibilities and is appropriate on servers (whereas
+ * {@link #MODE_GLOBAL} is definitely inappropriate for server use).
+ *
+ * @author Francisco Spaeth
+ * @since 1.1
+ * 
+ */
+public class SessionContextHolder {
+
+	public static final String MODE_THREADLOCAL = "MODE_THREADLOCAL";
+	public static final String MODE_INHERITABLETHREADLOCAL = "MODE_INHERITABLETHREADLOCAL";
+	public static final String MODE_GLOBAL = "MODE_GLOBAL";
+	public static final String SYSTEM_PROPERTY = "spring.session.strategy";
+	private static String strategyName = System.getProperty(SYSTEM_PROPERTY);
+	private static SessionContextHolderStrategy strategy;
+	private static int initializeCount = 0;
+
+	static {
+		initialize();
+	}
+
+	/**
+	 * Explicitly clears the context value.
+	 */
+	public static void clearContext() {
+		strategy.clearContext();
+	}
+
+	/**
+	 * Obtain the current <code>Session</code>.
+	 * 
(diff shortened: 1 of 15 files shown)


rwinch:
Thanks for the PR @franciscospaeth! At quick glance this looks like an outstanding PR! I will take a look at this sometime next week and get it merged into master. Thanks again!

franciscospaeth:
Great! Please let me know if something doesn't make sense or could be improved.

spring-session/src/main/java/org/springframework/session/context/SessionContextHolder.java
```diff
@@ -0,0 +1,162 @@
...
+		} else {
+			instantiateCustomStrategy();
+		}
+
+		initializeCount++;
+	}
+
+	private static void instantiateCustomStrategy() {
+		try {
+			Class<?> clazz = Class.forName(strategyName);
```
martin-g: I foresee OSGi users crying because of this but this is something normal for them :-)

spring-session/src/main/java/org/springframework/session/web/http/SessionRepositoryFilter.java
```diff
@@ -132,6 +136,22 @@ public void setServletContext(ServletContext servletContext) {
 		this.servletContext = servletContext;
 	}
 
+	private final class HttpRequestSessionContext implements SessionContext {
```
martin-g: As far as I can see this class could be `static`, so it won't keep to its outer. This is preferable most of the time.

spring-session/src/main/java/org/springframework/session/context/SessionContextHolder.java
rwinch: Thanks for the comment. Spring Security does something very similar and it works within OSGi, so this shouldn't be an issue.

spring-session/src/main/java/org/springframework/session/web/http/SessionRepositoryFilter.java
rwinch: Thanks for your comment. You are right that in many instances it is preferable to have a static class. However, this won't work since the `SessionRepositoryRequestWrapper` is non-static. The reasons `SessionRepositoryRequestWrapper` is non-static is because it needs a reference to the type argument of `SessionRepositoryFilter` to be type safe.

rwinch:
@franciscospaeth Thanks again for this PR! 

I took a look at the PR and have made some [minor changes](https://github.com/rwinch/spring-session/tree/gh-267) in my branch.

I also have a few concerns / comments about the implementation that we could perhaps discuss.


# Possible Inconsistent Session

My concern is that if someone sets the Session on the SessionContextHolder, then it is not reflected on the HttpSession. This seems like it could lead to inconsistencies. For example, if in the service tier someone did the following:

```java
SimpSessionContext context = new SimpSessionContext();
context.setSession(session);
SessionContextHolder.setContext(context);
```

Then later on in the web tier someone attempted to access the `HttpSession`:

```java
HttpSession session = request.getSession();
```

The two sessions would not be the same id (thus have different instances).

I'm thinking it may be best to do some refactoring to account for this.

# SessionContext necessary?

With the above refactoring I don't think that SessionContext is necessary. Instead the Session can be set and accessed directly. The HttpSessionWrapper would look up the value in the SessionContextHolder rather than save as a request attribute. I believe there may be instances where it needs to fall back to obtain the Session from the request attribute, but generally it could access the SessionContextHolder.

# TBD

What are your thoughts? If you agree I'm happy to do the cleanup or I am happy to have you send an updated PR.

franciscospaeth:
Hi @rwinch,

The changes makes sense :+1:.

Regarding the inconsistency, this is a real problem.

About the `SessionContext`, it was introduced to abstract the `SessionContextHolder` from the responsible to provide/create the session instance. Within a http request for example, session should be created only if requested. If `SessionContextHolder.getSession()` is implemented, from my point of view, it should do what `HttpServletRequest.getSession()` does, covering the case in which `getSession()` is first invoked from service tier using `SessionContextHolder`.

I had tried out some things, among them to use `SessionContextHolder` from `ExpiringSessionHttpSession`, but once `SessionContextHolder.getSession()` returns `Session` some ugly unchecked cast to `S` (`<S extends ExpiringSession>`) are required. I will think further about it, but would be happy if there are any ideas/suggestions to solve the problem, maybe I'm just not seeing the big picture.

rwinch:
Thanks for the response.

> If SessionContextHolder.getSession() is implemented, from my point of view, it should do what HttpServletRequest.getSession() does, covering the case in which getSession() is first invoked from service tier using SessionContextHolder

You bring up a good point. It would be nice to be able to create a new session from anywhere within the application. However, I think we would also need a way to inspect to see if there is a session already.

> but once SessionContextHolder.getSession() returns Session some ugly unchecked cast
> S (<S extends ExpiringSession>) are required.

Given it is a static method there is no way to and the type short of adding a generic type capture to the method. This would be consistent with the way `<T> T Session.getAttribute(String)` works too.

For example, `SessionContextHolder` could be something like this:

```java
public static <S extends Session> getSession() {
    ...
}
```

Obviously this would still require a cast, but it would be more isolated.

Thoughts?

franciscospaeth:
Hi @rwinch,

Using `SessionContextHolder` as a central point to hold the session is a good idea, but I'm still trying to understand when should the `Session` be set to `SessionContextHolder`, considering `SessionContext` is removed. 
If it is done within the `doFilter` (`SessionRepositoryFilter.doFilterInternal(HttpServletRequest, HttpServletResponse, FilterChain)`), we need to try to fetch the session from the repository even if it is not used during the request. And if it is set when it is retrieved from the request object (`SessionRepositoryRequestWrapper.getSession(boolean)`), we wouldn't have the session available before executing `HttpServletRequest.getSession(boolean)`.

Suggestions?

rwinch:
Sorry this is just a quick response as I won't have time to give this a ton of thought tonight (hopefully holes with this). I was thinking that:

* `SessionRepositoryRequestWrapper.getSession().getSession()` would always look up in the `SessionContextHolder`. If that is null (i.e. on processing an Error), then it would retrieve from the member variable and then populate the `SessionContextHolder`
* `SessionRepositoryRequestWrapper.getSession().setSession(Session)` would write to the `SessionContextHolder`
* If `SessionContext.getSession()` is invoked and the value is null, it would create a new one by invoking `SessionRepositoryRequestWrapper.getSession().getSession()` which would update the `SessionContextHolder`

franciscospaeth:
I explored the idea to have the session stored at `SessionContextHolderStrategy`, here you will find a possible solution (but is still not mature enough): rwinch/spring-session/pull/1

rwinch:
@franciscospaeth I am looking into this again. It seems that the PR you have is now gone (due to the fact I had to delete the repository). Do you mind sending it again?

I'm also thinking we should keep SessionContextHolder simple and remove the strategies and just use an `InheritableThreadLocal`. `RequestContextHolder` in spring-web doesn't need the strategies and since we could passively introduce strategies, I'd prefer to wait until we need them.

PS: Sorry for my long delay on this.

rwinch:
Closing due to lack of feedback.

PR #274, opened 2015-08-19
Add Hazelcast Spring (w/ Session*Event) Sample

manderson23:
This updates the Hazelcast sample to demonstrate firing SessionCreatedEvent, SessionDeletedEvent and SessionExpiredEvent.

Please let me know if anything else needs to be done. 

There are a couple of places where indentation etc. could be fixed. Do you have an Eclipse format XML file for your coding standard? Or can you point me to a coding standard.

PR #277, opened 2015-08-24
Add @EnableHazelcastHttpSession and HazelcastHttpSessionConfiguration

### LINKED ISSUE #276: Add @EnableHazelcastHttpSession annotation and corresponding @Configuration
Similar to the existing `@EnableRedisHttpSession` annotation, I would like to see one for Hazelcast. This will make configuration easier when using Hazelcast, and will help pave the way for auto-configuration in Spring Boot. I am willing to work on a pull request for both.

shakuzen:
This provides support for Hazelcast as a backing datastore for SpringHttpSession. This makes it easier for the user to get started with Hazelcast in a similar fashion to the equivalent Redis configuration.

fixes #276
related to #274

Still need to add the configuration for listeners to generate the Session*Events, which can mostly be adapted from @manderson23's Hazelcast sample. @rwinch I assume you don't want yet another sample for this, so if this does get merged, perhaps @manderson23's sample could use the new `@EnableHazelcastHttpSession` annotation to showcase it.

I wasn't sure which, if any, part of the documentation you want updated with this. Please let me know, and I'd be happy to add anything.

Lastly, and maybe it is a long shot, but I would ideally like to get Hazelcast auto-configuration into Spring Boot 1.3 including Spring Session auto-configuration for Hazelcast, which would use the annotation and consequently the configuration added in this pull request. Is there a timeline for Spring Session 1.1 that would allow it to make it into Spring Boot 1.3? Or would it be possible to add this to a 1.0.x release for inclusion in Spring Boot 1.3? Of course all of this depends on me getting the pull request together and it being accepted by the Spring Boot team in the first place. See spring-projects/spring-boot#3811 for the session auto-configuration.


spring-session/src/main/java/org/springframework/session/data/hazelcast/config/annotation/web/http/EnableHazelcastHttpSession.java
```diff
@@ -0,0 +1,66 @@
...
+ * @since 1.1
+ * @see EnableSpringHttpSession
+ */
+@Retention(value=java.lang.annotation.RetentionPolicy.RUNTIME)
+@Target(value={java.lang.annotation.ElementType.TYPE})
+@Documented
+@Import(HazelcastHttpSessionConfiguration.class)
+@Configuration
+public @interface EnableHazelcastHttpSession {
+	int maxInactiveIntervalInSeconds() default 1800;
```
manderson23: Would it be better to have a default of -1 (to indicate it should be ignored) for the case where you would like to use the value defined in an external Hazelcast configuration? Otherwise Spring Session will always overwrite.

shakuzen: I wanted to maintain the functionality in a very similar way to the Redis equivalent in order to lower the barrier to entry. You are right that Spring Session will always overwrite this value, which can be confusing if you configure the same thing externally. However, Spring Session would only be configuring the specific `Map` used for the `SessionRepository` which seems like it might acceptable and the same as the Redis configuration does, from what I can tell.

If we do want to avoid overwriting, I think we would have to change to an `Integer` and make the default for this `null` (I think -1 would be potentially confusing and could be mistaken for a valid value when it is not). Then only overwrite the value (Hazelcast's default `maxIdleSeconds` is 0) if this is not null.
What do you think?

rwinch: > Would it be better to have a default of -1 (to indicate it should be ignored)

I don't like the idea of -1 because in terms of HTTP Servlet specification that means do not expire. From the http://docs.oracle.com/javaee/5/api/javax/servlet/http/HttpSession.html#getMaxInactiveInterval()[Javadoc]:  "A negative time indicates the session should never timeout."

> change to an Integer and make the default for this null

Unfortunately Annotations cannot have a type of Integer and they cannot have a default value of null. Are we able to read the configuration that was provided and if it was specified in the configuration, then do not use the value in the annotation?

rwinch:
Thanks for the PR!

> Still need to add the configuration for listeners to generate the Session*Events

Right this needs to be done before we can claim we support Hazelcast.

> I assume you don't want yet another sample for this, so if this does get merged, perhaps @manderson23's sample could use the new @EnableHazelcastHttpSession annotation to showcase it.

I think this would be a good idea.

> I wasn't sure which, if any, part of the documentation you want updated with this. Please let me know, and I'd be happy to add anything.

* We need to ensure that the reference contains a EnableHttpHazelcastHttpSession section similar to http://docs.spring.io/spring-session/docs/current-SNAPSHOT/reference/html5/#api-enablespringhttpsession[EnableSpringHttpSession]. It should contain details about storage similar to http://docs.spring.io/spring-session/docs/current-SNAPSHOT/reference/html5/#api-redisoperationssessionrepository[RedisOperationsSessionRepository]
* We would want a guide that walks users through setting up Spring Session with the new support.

> Lastly, and maybe it is a long shot, but I would ideally like to get Hazelcast auto-configuration into Spring Boot 1.3 including Spring Session auto-configuration for Hazelcast, which would use the annotation and consequently the configuration added in this pull request.

This will not happen since Spring Session has not even done an RC and Spring Boot is nearly completion. Once we get the support for configuration in Spring Session, we can discuss auto configuration.

Also we need some integration tests of the configuration before we can merge the request. They would likely be similar to https://github.com/spring-projects/spring-session/blob/db2759c49731fba78689bb9330f400ea49244e2e/spring-session/src/integration-test/java/org/springframework/session/data/redis/RedisOperationsSessionRepositoryITests.java[RedisOperationsSessionRepositoryITests] 

Thanks again for the PR!

spring-session/src/main/java/org/springframework/session/data/hazelcast/config/annotation/web/http/EnableHazelcastHttpSession.java
manderson23: I can see the value in consistency across repository implementations. And I also agree that -1 isn't a reasonable value.

I just raised it as a possible concern as I suspect my use case may have the hazelcast configuration externalised from the Spring configuration.

What about if the `maxIdleSeconds` is 0 (the default) we override with the value from the annitation otherwise we leave it and don't use it (like @rwinch suggested).

shakuzen: @rwinch I'm not sure how we would be able to read the provided configuration. The user could put the configuration in any xml file and include that configuration. Alternatively, they could configure it in Java, reading the values from a properties file, perhaps.

@manderson23 The problem is, as far as I can tell, we have no way to know if the user configured `maxIdleSeconds` as 0 or if they did not configure it and it is just 0 by default. In case they configured it as 0, I think we would want to respect that and not overwrite it.

I'm leaning more towards having to change the default attribute value to 0 to match to default of `maxIdleSeconds`. Then the user can configure it themselves and we will only override if a value is passed to `maxInactiveIntervalInSeconds`. Though this is different than the Redis configuration, I believe. So we have to choose to play it safe and avoid overwriting the user's configuration, or maintain consistency with the equivalent Redis configuration.

shakuzen:
@rwinch Understood. I'll get to working on the remaining points and add them to this pull request. Thank you for the guidance. I'm always glad to help.

spring-session/src/main/java/org/springframework/session/data/hazelcast/config/annotation/web/http/EnableHazelcastHttpSession.java
rwinch: @shakuzen 

> I'm leaning more towards having to change the default attribute value to 0 to match to default of maxIdleSeconds. 

Hazelcast can be used for many things, so setting it to 0 may make sense when there is no context. However, when considered for an HttpSession we should be conscious that the default needs to be secure. A default that prevents expiration is not what I would consider secure.

> Then the user can configure it themselves and we will only override if a value is passed to maxInactiveIntervalInSeconds. Though this is different than the Redis configuration, I believe. So we have to choose to play it safe and avoid overwriting the user's configuration, or maintain consistency with the equivalent Redis configuration.

I'm wondering why this needs to be configured externally vs using the annotation. Can you elaborate?

One option is we could make the maxInactiveIntervalInSeconds attribute a String. If the value was "", then we would use the value configured externally. The default would still be 1800 to preserve a secure by default setting. Thoughts?

manderson23: @rwinch as you say Hazelcast can be used for many things and my thoughts were that we should support Hazelcast being configured completely outwith spring (this will probably be the case for me) and by default don't override any of these values.

In this case I can probably handle ensuring this is configured via Spring instead of assuming the Hazelcast configuration value is used but maybe that wouldn't be the case for everyone.

rwinch: @manderson23 Does my proposal sound like it will work for you? Specifically something like this would ensure the value configured externally would be used:

```java
@EnableHazelcastHttpSession(maxInactiveIntervalInSeconds="")
```

Another advantage to using a String is that users can leverage Spring properties (i.e. `${session.timeout}`) for configuring the timeout as well. Of course the disadvantage is that a String is not necessarily a number.

manderson23: @rwinch that would work for me.

spring-session/src/main/java/org/springframework/session/data/hazelcast/config/annotation/web/http/EnableHazelcastHttpSession.java
```diff
@@ -0,0 +1,66 @@
...
+ * backed by Hazelcast. In order to leverage the annotation, a single {@link HazelcastInstance}
+ * must be provided. For example:
+ * <pre>
+ * <code>
+ * {@literal @Configuration}
+ * {@literal @EnableHazelcastHttpSession}
+ * public class HazelcastHttpSessionConfig {
+ *
+ *     {@literal @Bean}
+ *     {@literal @ConditionalOnMissingBean}
```
rwinch: `@ConditionalOnMissingBean` only exists in Spring Boot, so we likely shouldn't include that in our Javadoc (that will confuse users who do not leverage Boot).

shakuzen: :+1: I agree. I meant to remove this in the next commit I made and I just haven't made another one yet.

spring-session/src/main/java/org/springframework/session/data/hazelcast/config/annotation/web/http/HazelcastHttpSessionConfiguration.java
```diff
@@ -0,0 +1,124 @@
...
+				enableAttrMap = AnnotationUtils
+						.getAnnotationAttributes(enableHazelcastHttpSessionAnnotation);
+				enableAttrs = AnnotationAttributes.fromMap(enableAttrMap);
+			}
+		}
+		
+		if (DO_NOT_CONFIGURE_INACTIVE_INTERVAL_STRING.equals(enableAttrs.getString("maxInactiveIntervalInSeconds"))) {
+			this.maxInactiveIntervalInSeconds = null;
+		} else {
+			this.maxInactiveIntervalInSeconds = enableAttrs.getNumber("maxInactiveIntervalInSeconds");
```
shakuzen: If something other than a `Number` is given, the application will fail on startup with a straightforward message, which I figured was sufficient exception handling. Let me know if you want anything else.

spring-session/src/main/java/org/springframework/session/data/hazelcast/config/annotation/web/http/EnableHazelcastHttpSession.java
```diff
@@ -0,0 +1,77 @@
...
+	/**
+	 * This is the session timeout in seconds. By default, it is set to 1800 seconds (30 minutes).
+	 * This should be a non-negative integer.
+	 * <p>If you wish to use external configuration (outside of this annotation) to set this value, you can
+	 * set this to {@link HazelcastHttpSessionConfiguration#DO_NOT_CONFIGURE_INACTIVE_INTERVAL_STRING},
+	 * which will prevent this configuration from overriding the external configuration for this value.</p>
+	 * 
+	 * @return the seconds a session can be inactive before expiring
+	 */
+	String maxInactiveIntervalInSeconds() default "1800";
```
shakuzen: @rwinch and @manderson23 I implemented my understanding of the previous discussion on the default/overriding the session timeout. Let me know what you think. Currently it only mentions the String constant I added `DO_NOT_CONFIGURE_INACTIVE_INTERVAL_STRING`, but perhaps the JavaDoc for this can/should mention the value of that constant as `""`?
Also let me know if you think of a better name for the constant.

spring-session/src/main/java/org/springframework/session/data/hazelcast/config/annotation/web/http/HazelcastHttpSessionConfiguration.java
shakuzen: Sorry, this actually fails right now. I thought I tested it. I will push a fix after testing...

spring-session/src/main/java/org/springframework/session/data/hazelcast/config/annotation/web/http/EnableHazelcastHttpSession.java
rwinch: At least for now, lets not provide a public constant. Stating empty String in the javadoc should be sufficient since it is very easy to reproduce.

shakuzen: Updated now as requested.

rwinch:
Thanks for the updates! I have officially added the [hazelcast-spring](https://github.com/spring-projects/spring-session/tree/master/samples/hazelcast-spring) sample into master.

At the moment the following things are necessary to get this PR merged.

* Add integration tests for the configuration
* Include the listeners from hazelcast-spring sample in Spring Session distribution. Perhaps merging them into a single class that implements all interfaces.
* Update the provided configuration to leverage the listeners
* Update the hazelcast-spring sample to use the new configuration
* Add the documentation I mentioned. Specifically:
  * We need to ensure that the reference contains a EnableHttpHazelcastHttpSession section similar to [EnableSpringHttpSession](http://docs.spring.io/spring-session/docs/current-SNAPSHOT/reference/html5/#api-enablespringhttpsession). It should contain details about storage similar to [RedisOperationsSessionRepository](http://docs.spring.io/spring-session/docs/current-SNAPSHOT/reference/html5/#api-redisoperationssessionrepository)
  * We would want a guide that walks users through setting up Spring Session with the new support.

manderson23:
@rwinch @shakuzen with the Hazelcast annotation should we consider removing the listeners from the map when the context is destroyed? This could be an issue is a webapp context was redeployed while the underlying Hazelcast instance wasn't destroyed? We don't want to end up with orphaned listeners.

rwinch:
Good call @manderson23 I've added that to the list of things that needs done.

spring-session/src/main/java/org/springframework/session/data/hazelcast/SessionEntryListener.java
```diff
@@ -0,0 +1,72 @@
...
+	private ApplicationEventPublisher eventPublisher;
+
+	public SessionEntryListener(ApplicationEventPublisher eventPublisher) {
+		Assert.notNull(eventPublisher, "eventPublisher cannot be null");
+		this.eventPublisher = eventPublisher;
+	}
+
+	public void entryAdded(EntryEvent<String, ExpiringSession> event) {
+		logger.debug("Session created with id: " + event.getValue().getId());
+		eventPublisher.publishEvent(new SessionCreatedEvent(this, event.getValue()));
```
shakuzen: I should probably change these to `this.eventPublisher` to match style.

shakuzen:
@rwinch @manderson23 I apologize for my inactivity lately. I have pushed the first set of integration tests for `@EnableHazelcastHttpSession`. I extracted out the `SessionEventRegistry` class that was in `RedisOperationsSessionRepositoryITests` because it seemed generally useful in integration tests handling SessionEvents regardless of the implementation. Furthermore, I wrote the tests so far in a completely agnostic manner to whether it is Redis or Hazelcast underneath. It might be nice if we can have the same set of integration tests run against a config with Redis and Hazelcast without having to duplicate the code (which I did do a little bit). I thought this task might be a bit out of scope for this pull request, though.

Next I will write some IT specific to the Hazelcast implementation. Any hints on how to test that the entry listener is actually removed upon shutdown?

Of course let me know if I've missed anything.

**TODO List**
- [ ] Add integration tests that check underlying Hazelcast directly
- [ ] Update the hazelcast-spring sample to use the new configuration
- [ ] Add the documentation:
  - [ ] EnableHttpHazelcastHttpSession section similar to EnableSpringHttpSession.
  - [ ] A guide that walks users through setting up Spring Session with the new support.

manderson23:
@rwinch @shakuzen 

If it were OK testing shutdown and listener removal in unit tests rather than integration tests you could mock HazelcastInstance and IMap since they are both interfaces. Then on shutdown verify that removeEntryListener was called on the IMap mock.

spring-session/src/main/java/org/springframework/session/data/hazelcast/config/annotation/web/http/HazelcastHttpSessionConfiguration.java
```diff
@@ -0,0 +1,156 @@
...
+	
+	private String sessionListenerUid;
+	
+	private IMap<String, ExpiringSession> sessionsMap;
+
+	@Bean
+	public SessionRepository<ExpiringSession> sessionRepository(HazelcastInstance hazelcastInstance, SessionEntryListener sessionListener) {
+		configureSessionMap(hazelcastInstance);
+		this.sessionsMap = hazelcastInstance.getMap(sessionMapName);
+		this.sessionListenerUid = this.sessionsMap.addEntryListener(sessionListener, true);
```
manderson23: Are we happy that this is using a deprecated method?

manderson23:
@shakuzen 

I was integrating this into my code today and ran into an issue where when I had multiple nodes in the cluster I wasn't seeing expiry events.

In my Hazelcast XML I didn't include max-idle-seconds so the default value of 0 was used. I was then overriding in the Java code to set this to 60 seconds. However, I didn't see expiry events when I done this only creation events.
 
To get it to expire consistently I had to set max-idea-seconds to 60 in the Hazelcast XML. Is there an issue with overriding this in Hazelcast when the map already exists? Have you seen anything like this?

shakuzen:
@manderson23 I have not tested that specifically and it sounds like it will be a good idea for me to add a test that verifies that scenario automatically. I'll take a look into why the overridden value via the annotation isn't being applied. I'm on vacation until Friday and will probably be busy getting caught up at work Friday, so I'll try to get to it Saturday if I can.

manderson23:
@shakuzen No problems. I'm using my own copy of `HazelcastHttpSessionConfiguration` for now so it isn't anything to do with the value from the annotation being applied. I suspect it is one of two issues

1. I'm not configuring Hazelcast correctly.
2. After creating a map in the XML you cannot override the max-idle-seconds property programatically and have it work correctly.

The strange thing is that JMX seems to report the MapConfig was changed but the expiry events don't happen. I'll try and look into it some more if I get the chance. I just wondered if you had come across any similar issues.

manderson23:
@shakuzen you can ignore my previous comment. It was a local Spring issue.

shakuzen:
@rwinch I believe the pull request now has all of the requested items. Please take a look when you get a chance, and let me know any feedback you have.

rwinch:
Thanks for the PR @shakuzen! I merged this into master after I squashed your commits (leaving you as the author) d1c00c60800a91b1bcb5d08378a9cc29c52f6c42 and performed some (mostly cosmetic) polish on it 76d341d6ca1d591650648f259fe3072e283771d7. Please feel free to review my changes and provide any feedback you might have.

Thanks again!

shakuzen:
@rwinch Thank you for taking a look and merging. Sorry for all the formatting issues (stray spaces and tabs) - I'll make sure to clean those up in any future PRs. I'm looking forward to this being included in a release so hopefully I can try it out at work.

manderson23:
@shazuken thanks for this. Much appreciated.

rwinch:
@shakuzen Not a problem. Thank *you* for the PR. In regards to the "issues", please don't worry about it. This has been one of the best PR's I have seen in a while!

Issue #306, opened 2015-11-12
Remove Unnecessary Code from RedisHttpSessionConfiguration


rwinch:
There is [code](https://github.com/spring-projects/spring-session/blob/1.0.2.RELEASE/spring-session/src/main/java/org/springframework/session/data/redis/config/annotation/web/http/RedisHttpSessionConfiguration.java#L124-L136) that was unnecessarily copied from Spring Security's [WebSecurityConfiguration](https://github.com/spring-projects/spring-security/blob/4.0.3.RELEASE/config/src/main/java/org/springframework/security/config/annotation/web/configuration/WebSecurityConfiguration.java#L209-L225). The code is no longer necessary, but was at the time to work around [SPR-11251](https://jira.spring.io/browse/SPR-11251)

Related: [SEC-3148](https://jira.spring.io/browse/SEC-3148)

PR #307, opened 2015-11-15
Remove unnecessary code in HazelcastHttpSessionConfiguration

shakuzen:
I copied this code from `RedisHttpSessionConfiguration` without really understanding why it was needed. I should have asked then, which would have prevented it from being introduced in the first place. As this workaround is no longer needed, I am removing this unnecessary code.

Related to #306


rwinch:
Thanks for the PR! This is merged into master via f1d0350356719597e8cbdca3002df52ea5895e21

Issue #301, opened 2015-11-10
Use Spring Data Redis Index Support


rwinch:
Instead of implementing an index on the username, we should leverage Spring Data Redis support [DATAREDIS-425](https://github.com/spring-projects/spring-data-redis/tree/issue/DATAREDIS-425) for this (once it is available).

Related #7

vpavic:
I think this falls in the same category as #684 i.e. it doesn't fit our requirements due to similar reasons. Spring Data Redis secondary indexes are a feature of Redis Repositories and (like `HashMapper` API) doesn't fit our write scenarios where we maintain a delta of changes in order to optimize writes.

Issue #148, opened 2015-02-14
Add GemFire Support


rwinch:
There has been [some](https://spring.io/blog/2014/07/08/spring-session-1-0-0-m1-released#comment-1853090512) interest in GemFire base repository. A partial solution exists on [github](https://github.com/viniciusccarvalho/spring-session-data-gemfire)

tcesenthil:
After reading through the Spring Session Documentation, here is my understanding. Will these 3 steps (at a high level) be sufficient to allow Spring Session to write to Gemfire repository instead of Redis ? I am interested in getting this component developed and tested for public & my private use.

Step 1: Implement just a Configuration class which provide all functions as the annotation

 1. Allow spring to Load the configuration class 
 2. Register Spring Session Filter in Container 
 3. Establish Repo Connection Factory
 4. Repo connection configuration

we will continue to re-use the Spring Session’s springSessionRepositoryFilter

Step 2: Need to develop an equivalent GemfireOperationsSessionRepository implementing the interface SessionRepository<S Session>

Step 3: SessionMessageListener.java 
   3.1. Need to decide a technique to identify and save delta changes in Session to underlying repository
   3.2. Need to see how session expire notification from underlying repository can be captured to invoke SessionDestroyEvent and cleanup operations -

PR #308, opened 2015-11-18
DO NOT MERGE - SGF-373 - Implement a Spring Session Adapter for GemFire backing a Ht…

jxblum:
…tpSession similar to the Redis support.


spring-session-data-gemfire/build.gradle
```diff
@@ -0,0 +1,18 @@
+apply from: JAVA_GRADLE
+apply from: MAVEN_GRADLE
+
+apply plugin: 'spring-io'
+
+description = "Aggregator for Spring Session and Spring Data GemFire"
+
+dependencies {
+    compile project(':spring-session'), "org.springframework.data:spring-data-gemfire:$springDataGemFireVersion"
```
martin-g: Is it time to split the project in -core and modules like -redis, -hazelcast, -gemfire ?
A user of Redis doesn't need Gemfire jars in her application (scope `compile`).
Another way would be to use scope `provided` and then each application should provide the respective dependency.

rwinch:
Thanks for sending this for my early review :)

spring-session/src/main/java/org/springframework/session/data/gemfire/AbstractGemFireOperationsSessionRepository.java
```diff
@@ -0,0 +1,284 @@
+/*
```
rwinch: Is there good reason to have a base class right now since we only have a single repository? If not, could this be folded into a single implementation or made package private for now? This will allow us to more flexibility when updating. Of course if there is good reason for this being made available I'd like to understand why.

spring-session/src/main/java/org/springframework/session/data/gemfire/GemFireOperationsSessionRepository.java
```diff
@@ -0,0 +1,352 @@
...
+ * @see org.springframework.session.ExpiringSession
+ * @see org.springframework.session.Session
+ * @see org.springframework.session.data.gemfire.AbstractGemFireOperationsSessionRepository
+ * @since 1.1.0
+ */
+@SuppressWarnings("all")
+public class GemFireOperationsSessionRepository extends AbstractGemFireOperationsSessionRepository {
+
+	// GemFire OQL query used to look up Sessions by principal name.
+	protected static final String FIND_SESSIONS_BY_PRINCIPAL_NAME_QUERY =
```
rwinch: A note that we will need to update how we do the principal when we fix #301. The reason is that I think we will be making it so that any attribute can be indexed, so interface for finding by a session attribute will be more generic. Perhaps something like `findBySessionAttr(String name, String value)`

spring-session/src/main/java/org/springframework/session/data/gemfire/GemFireOperationsSessionRepository.java
```diff
@@ -0,0 +1,352 @@
...
+	}
+
+	/**
+	 * Saves the specified {@link ExpiringSession} to GemFire.
+	 *
+	 * @param session the {@link ExpiringSession} to save.
+	 * @see org.springframework.session.ExpiringSession
+	 */
+	@Override
+	public void save(ExpiringSession session) {
```
rwinch: If we fold into a common base class, we can use GemFireSession directly rather than having to wrap the value passed in. Alternatively, the base class could be made generic.

This also allows for customizing when the session gets persisted. For example, we might have the ExpiringSession object itself perform the writes if users want immediate writes.

spring-session/src/main/java/org/springframework/session/data/gemfire/GemFireOperationsSessionRepository.java
```diff
@@ -0,0 +1,352 @@
...
+			writer.writeInt("maxInactiveIntervalInSeconds", getMaxInactiveIntervalInSeconds());
+			writer.writeString("principalName", getPrincipalName());
+
+			Set<String> attributeNames = new HashSet<String>(nullSafeSet(getAttributeNames()));
+
+			writer.writeObject("attributeNames", attributeNames);
+
+			for (String attributeName : attributeNames) {
+				writer.writeObject(attributeName, getAttribute(attributeName));
+			}
```
rwinch: I'm wondering if this is writing every attribute every time. If so, is there a way to only write the updates? This will increase performance a lot and will handle race conditions better. For example:

* Assume we have a Session with the attribute of username
* Thread1 - starts to process a request to get a protected resource. The username attribute is present in the session.
* Thread2 - processes a request to log the user out (remove the username attribute).
* Thread2 - The username attribute is removed from the session and Redis is updated to reflect the username is removed
* Thread1 - the request finishes and the Session is updated to include all of the original Session attributes. This means the username is now set in the Session which indicates the user is logged in. However, the user should be logged out.

If we were to write only the delta, this would not be an issue. However, writing every attribute every time alleviates problems like #311

spring-session/src/main/java/org/springframework/session/data/gemfire/GemFireOperationsSessionRepository.java
jxblum: Noted

spring-session/src/main/java/org/springframework/session/data/gemfire/GemFireOperationsSessionRepository.java
jxblum: This is something the GemFire team did with the HTTP Session Management module and something we could implement on behalf of the user.  In GemFire, it is a simple "flag" to flush on write.

spring-session/src/main/java/org/springframework/session/data/gemfire/GemFireOperationsSessionRepository.java
jxblum: Delta has been implemented in the latest commit.

spring-session/src/main/java/org/springframework/session/data/gemfire/AbstractGemFireOperationsSessionRepository.java
jxblum: Yes.  There are several reasons:

1. I wanted to have a clean separation between the basic CRUD operations implemented in the concrete class (e.g. GemFireOperationsSessionRepository) and the event handling/translation logic that would be common (literally the same) for all concrete classes.  The GemFire cache event to Spring Session event translation is, or rather should be, consistent across all GemFire SessionRepository implementations.  The concrete implementation should only have to focus on implementing the CRUD operations IMO.

2. Also, the AbstractGemFireSessionRepository handles "registering" the GemFireOperationSessionRepository as a "CacheListener" on the appropriate GemFire component without the user having to know which component (i.e. the GemFire Region storing Session state).

3. Next, I have just moved the GemFireSession class (along with a new GemFireSessionAttributes supporting class) to the AbstractGemFireSessionRepository class to simplify future SessionRepository implementations based on GemFire.  For instance, another possible implementation might be to use SD's Repository abstraction (i.e. GemfireRepository) to back the SessionRepository based on GemFire.  This was what Vinicius did in his implementation of Spring Session GemFire support, but his approach was quite brittle in how he registered the repo.

There are subtle other reasons, but the 3 points above capture my main intent.  I had thought about naming this class `GemFireSessionRepositorySupport`, but I have not renamed it yet.

rwinch:
Thank you for the PR! Here is the current check list of what we need done. 

- [ ] Documentation. This is really thing that *must* be done before we merge. Everything else is cosmetic and is something I can manage for you if necessary.
- [ ] Squash commits. We should strive for a single commit per committer. If there is real good reason to split into multiple commits we can make exceptions.
- [ ] Rename the commit comment to refer to Issue #148 
- [ ] Remove `@Override` annotations from interface overrides so that it compiles in Eclipse
- [ ] While good cleanup, we should probably move the changes of MapSessionRepository to a different commit https://github.com/spring-projects/spring-session/pull/308/files#diff-3e486055b7d5bb484429c172627aba11L68

jxblum:
All check list items resolved.

* Documented all Spring SessionGemFire support and features including: Javadoc, updated Spring Session User Guide and Samples (4 in total for client/server and p2p using both Java Config and XML).
* Squashed/fixed up all commits into a single commit summarizing all the development work to add GemFire support to Spring Session.
* Renamed the commit to refer to PR #148.
* Removed all @Override annotations on interface overrides; still used @Override on class methods.
* Reverted changes to the MapSessionRepository class.

If there are any issues, let me know.

rwinch:
Thank you for the updates John!

Here is my first round of comments. I will wait until we get these sorted out until we continue.

# Javadoc Failures

The first issue I run into is Javadoc failures. If I run the following with JDK 8 I get a failure:

```
~/code/spring-session$ ./gradlew clean check
...
:docs:compileTestJava

FAILURE: Build failed with an exception.

* What went wrong:
Could not resolve all dependencies for configuration ':docs:testCompile'.
> Could not find com.gemstone.gemfire:gemfire:8.1.0.
  Searched in the following locations:
      https://jcenter.bintray.com/com/gemstone/gemfire/gemfire/8.1.0/gemfire-8.1.0.pom
      https://jcenter.bintray.com/com/gemstone/gemfire/gemfire/8.1.0/gemfire-8.1.0.jar
  Required by:
      spring-session-build:docs:1.1.0.BUILD-SNAPSHOT > org.springframework.data:spring-data-gemfire:1.7.2.RELEASE

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

BUILD FAILED
```

# Tests Failures

* If I run `./gradlew build.` from the spring-session-build project I get failures. The first is:

   ```
:docs:compileTestJava

FAILURE: Build failed with an exception.

* What went wrong:
Could not resolve all dependencies for configuration ':docs:testCompile'.
> Could not find com.gemstone.gemfire:gemfire:8.1.0.
  Searched in the following locations:
      https://jcenter.bintray.com/com/gemstone/gemfire/gemfire/8.1.0/gemfire-8.1.0.pom
      https://jcenter.bintray.com/com/gemstone/gemfire/gemfire/8.1.0/gemfire-8.1.0.jar
  Required by:
      spring-session-build:docs:1.1.0.BUILD-SNAPSHOT > org.springframework.data:spring-data-gemfire:1.7.2.RELEASE
   ```
* If I resolve that I am then getting:

   ```
The Server is running at http://localhost:64851
:samples:httpsession-gemfire-clientserver-xml:jar
:samples:httpsession-gemfire-clientserver-xml:integrationTest

sample.AttributeTests > first visit no attributes FAILED
    org.codehaus.groovy.runtime.powerassert.PowerAssertionError at AttributeTests.groovy:32

2 tests completed, 1 failed, 1 skipped
   ```

   Looking at the test output I see:

   ```
Assertion failed: 

assert driver.title == 'Session Attributes'
       |      |     |
       |      null  false
       org.openqa.selenium.htmlunit.HtmlUnitDriver@56f318d8

	at sample.pages.HomePage._clinit__closure1(HomePage.groovy:27)
	at sample.pages.HomePage._clinit__closure1(HomePage.groovy)
	at geb.Page.verifyThisPageAtOnly(Page.groovy:187)
	at geb.Page.verifyAt(Page.groovy:146)
	at geb.Browser.doAt(Browser.groovy:367)
	at geb.Browser.at(Browser.groovy:298)
	at geb.Browser.to(Browser.groovy:450)
	at geb.Browser.to(Browser.groovy:426)
	at geb.spock.GebSpec.methodMissing(GebSpec.groovy:54)
	at sample.AttributeTests.first visit no attributes(AttributeTests.groovy:32)
   ```

   It seems like this is caused by the fact that when starting the application (to perform the tests) the following error is caused:

   ```
org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'gemfireCache': Invocation of init method failed; nested exception is org.springframework.beans.factory.BeanInitializationException: no bean of type 'com.gemstone.gemfire.cache.client.Pool' having name 'null' was found; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'DEFAULT': Initialization of bean failed; nested exception is java.lang.IllegalArgumentException: GemFire Server failed to start
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.initializeBean(AbstractAutowireCapableBeanFactory.java:1574)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:539)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:476)
	at org.springframework.beans.factory.support.AbstractBeanFactory$1.getObject(AbstractBeanFactory.java:303)
	at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.getSingleton(DefaultSingletonBeanRegistry.java:230)
	at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:299)
	at org.springframework.beans.factory.support.AbstractBeanFactory.getBean(AbstractBeanFactory.java:194)
	at org.springframework.beans.factory.support.DefaultListableBeanFactory.preInstantiateSingletons(DefaultListableBeanFactory.java:736)
	at org.springframework.context.support.AbstractApplicationContext.finishBeanFactoryInitialization(AbstractApplicationContext.java:757)
	at org.springframework.context.support.AbstractApplicationContext.refresh(AbstractApplicationContext.java:480)
	at org.springframework.web.context.ContextLoader.configureAndRefreshWebApplicationContext(ContextLoader.java:403)
	at org.springframework.web.context.ContextLoader.initWebApplicationContext(ContextLoader.java:306)
	at org.springframework.web.context.ContextLoaderListener.contextInitialized(ContextLoaderListener.java:106)
	at org.apache.catalina.core.StandardContext.listenerStart(StandardContext.java:5016)
	at org.apache.catalina.core.StandardContext.startInternal(StandardContext.java:5524)
	at org.apache.catalina.util.LifecycleBase.start(LifecycleBase.java:150)
	at org.apache.catalina.core.ContainerBase$StartChild.call(ContainerBase.java:1575)
	at org.apache.catalina.core.ContainerBase$StartChild.call(ContainerBase.java:1565)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
Caused by: org.springframework.beans.factory.BeanInitializationException: no bean of type 'com.gemstone.gemfire.cache.client.Pool' having name 'null' was found; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'DEFAULT': Initialization of bean failed; nested exception is java.lang.IllegalArgumentException: GemFire Server failed to start
	at org.springframework.data.gemfire.client.ClientCacheFactoryBean.resolvePool(ClientCacheFactoryBean.java:211)
	at org.springframework.data.gemfire.client.ClientCacheFactoryBean.initializePool(ClientCacheFactoryBean.java:182)
	at org.springframework.data.gemfire.client.ClientCacheFactoryBean.createCache(ClientCacheFactoryBean.java:172)
	at org.springframework.data.gemfire.CacheFactoryBean.resolveCache(CacheFactoryBean.java:303)
	at org.springframework.data.gemfire.CacheFactoryBean.init(CacheFactoryBean.java:244)
	at org.springframework.data.gemfire.CacheFactoryBean.afterPropertiesSet(CacheFactoryBean.java:220)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.invokeInitMethods(AbstractAutowireCapableBeanFactory.java:1633)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.initializeBean(AbstractAutowireCapableBeanFactory.java:1570)
	... 21 more
Caused by: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'DEFAULT': Initialization of bean failed; nested exception is java.lang.IllegalArgumentException: GemFire Server failed to start
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:547)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:476)
	at org.springframework.beans.factory.support.AbstractBeanFactory$1.getObject(AbstractBeanFactory.java:303)
	at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.getSingleton(DefaultSingletonBeanRegistry.java:230)
	at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:299)
	at org.springframework.beans.factory.support.AbstractBeanFactory.getBean(AbstractBeanFactory.java:217)
	at org.springframework.beans.factory.support.DefaultListableBeanFactory.getBean(DefaultListableBeanFactory.java:350)
	at org.springframework.beans.factory.support.DefaultListableBeanFactory.getBean(DefaultListableBeanFactory.java:331)
	at org.springframework.data.gemfire.client.ClientCacheFactoryBean.resolvePool(ClientCacheFactoryBean.java:206)
	... 28 more
Caused by: java.lang.IllegalArgumentException: GemFire Server failed to start
	at org.springframework.util.Assert.isTrue(Assert.java:65)
	at sample.GemFireCacheServerReadyBeanPostProcessor.postProcessBeforeInitialization(GemFireCacheServerReadyBeanPostProcessor.java:65)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.applyBeanPostProcessorsBeforeInitialization(AbstractAutowireCapableBeanFactory.java:408)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.initializeBean(AbstractAutowireCapableBeanFactory.java:1566)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:539)
	... 36 more
Error listenerStart
   ```

   It appears that simply running either gemfire xml sample results in the same error:

   ```
 $ ./gradlew :samples:httpsession-gemfire-clientserver-xml:tomcatRun
   ```

# Use AssertJ not Hamcrest

Spring Session uses AssertJ not Hamcrest for assertions. It would be best if the end result uses AssertJ. I can take care of this for you or you can do it (just let me know what you decide).

**NOTE**: This is a minor thing that I can take care of myself if necessary

Quite a few of the tests are using Hamcrest. Spring Session uses AssertJ for assertions

# Use Available Port for Gemfire server

We can use a static port for a client (i.e. Redis samples) if the CI server instances all have a server running on that port. However, we cannot guarantee a port will be available for the server to run on. This is particularly true since Spring Session might build two branches at the same time which would mean that one branch would fail since the port isn't available.

Instead, we should ensure that any servers should use an available port. You can check this out with the Hazelcast example which uses Spring's `SocketUtils.findAvailableTcpPort()`. If you want you can make the port static for the sample, but the integration tests should be dynamic.

jxblum:
@rwinch It looks like the [FEST Assertion API (1.4 & 2.x) has been replaced by AssertJ](https://github.com/joel-costigliola/assertj-core#migrating-from-fest-assertions) and that _Spring Session_ is still [using the old FEST Assertion API, v1.4](https://github.com/spring-projects/spring-session/blob/master/spring-session/build.gradle#L31) and not _AssertJ_, as indicated in the dependencies as well as not imports with...

   import static org.assertj.core.api.Assertions.*;

As outlined [here](http://joel-costigliola.github.io/assertj/assertj-core-quick-start.html).

rwinch:
Thanks @jxblum! I didn't realize I had not updated Spring Session to assertj yet. I've fixed this via #338

jxblum:
@rwinch Ah crap!  I just went through and reworked all the GemFire tests with the existing Fest Assertions.  With any luck, though, this should be a simple change the "import" statement and done (fingers crossed).  Most of the GemFire test assertions are pretty simple.

rwinch:
@jxblum Oops...sorry about that! It should be as easy as changing the import statement with possibly a few minor changes. Again, I don't mind doing this.

jxblum:
@rwinch 

The following has been addressed:

* Javadoc errors resolved
* Test failures resolved (due to missing GemFire dependency)
* All Test JUnit/Hamcrest-based assertions replaced with AssertJ.
* All GemFire client/server integration tests now randomly generate an available TCP port over which the client and server can communicate.

spring-session/build.gradle
```diff
@@ -25,8 +26,13 @@ dependencies {
 	integrationTestCompile "redis.clients:jedis:2.4.1",
 			"org.apache.commons:commons-pool2:2.2"
 
+	integrationTestRuntime "org.springframework.shell:spring-shell:1.0.0.RELEASE"
+
 	testCompile "junit:junit:$junitVersion",
-			'org.mockito:mockito-core:1.9.5',
+			'org.mockito:mockito-core:1.10.19',
+			"org.hamcrest:hamcrest-core:1.3",
```
shakuzen: Can the Hamcrest dependencies be removed now that everything has been changed over to AssertJ?

jxblum: Yes, and in fact I have just updated the PR again with a couple more changes including the removal of the _hamcrest-core_ testCompile dependency.

rwinch:
Thanks for the PR! I merged this via 019e0083b0b9fffdfe746fa2e17eb3e536d544b8 I provided a little bit of polish in 22f6f9dd727227ccd57a322a92e56c217ffc7e9e This fixes gh-148

Issue #311, opened 2015-11-24
Save session on Redis RedisSession.saveDelta


oassuncao:
I'm using Spring Session version 1.0.2.RELEASE and I found a problem when RedisSession save delta

If you add a new value on session (RedisSession) the session is saved correctly on Redis, because we have this code below:
```
		public void setAttribute(String attributeName, Object attributeValue) {
			cached.setAttribute(attributeName, attributeValue);
			delta.put(getSessionAttrNameKey(attributeName), attributeValue);
		}
```

But if in other request I get a object in this session, change some property of this object, the session is not updated.

The problem is when we create RedisSession passing MapSession, this map is setter on variable "cached" but the values of "cached" doesn't pass to "delta"

Example to reproduce the problem:
POJO
```
public class NamePOJO implements Serializable {
    private String name;

    public String getName() {
        return name;
    }

    public void setName(String name) {
        this.name = name;
    }
}
```

Controller
```
    ...
    @RequestMapping(value = "/changeSession/{value}")
    public void changeOnSession(@PathVariable(value  = "value") String value, HttpServletRequest request) throws IOException {
        final String sessionName = "SESSION_NAME";

        NamePOJO oldVersion;
        if (request.getSession().getAttribute(sessionName) == null) {
            oldVersion = new NamePOJO();
            oldVersion.setName(value);
            request.getSession().setAttribute(sessionName, oldVersion);
        }

        oldVersion = (NamePOJO)request.getSession().getAttribute(sessionName);

        log.info(String.format("Old name %s new name %s", oldVersion.getName(), value));
        oldVersion.setName(value);
    }
    ...
```

1* Request (/changeSession/Test)
Output: Old name Test new name Test
2* Request (/changeSession/Changed)
Output: Old name Test new name Changed
3* Request (/changeSession/Changed)
Output: Old name Test new name Changed

On the 3* request, the old name doesn't change

rwinch:
Hello @oassuncao and thank you for reporting this issue!

You are right that this is an issue, but it is deliberate. The reasoning is

* We want to limit the number of writes to the data store
* More importantly, we need to reduce race conditions. Consider the following:

** Assume we have a Session with the attribute of username
** Thread1 - starts to process a request to get a protected resource. The username attribute is present in the session.
** Thread2 - processes a request to log the user out (remove the username attribute).
** Thread2 - The username attribute is removed from the session and Redis is updated to reflect the username is removed
[...]

PR #313, opened 2015-11-24
Add cached values to delta

oassuncao:
One model to fixing the problem on issue https://github.com/spring-projects/spring-session/issues/311


rwinch:
Thanks again for taking the time to send a PR! Closing due to the feedback I provided on #311 Specifically we would want this to be a distinct strategy.

Issue #315, opened 2015-11-26
RedisSessionExpirationPolicy sends unnecessary del events when expire time is not changed.


tsachev:
When multiple requests are sent for the same session for example on page load
`RedisSessionExpirationPolicy` sends unnecessary del events for expirations.

They are not needed because the expiration time is rounded up and in on page load
scenario often the original expire time and the new expire time are rounded up to
the same value.

PR #316, opened 2015-11-26
Don't send expirations del events when expiration time does not change.

tsachev:
Expiration time is rounded up, so in case of many request per second from a
single session we can skip sending of delete event to reduce the traffic.
This is good because actually at the moment the redis repository is subscribed
for this events, even though it ignores them.

Fix gh-315


rwinch:
Thanks for the PR @tsachev! This is now merged into master

I love all the work you have been doing with Spring Session lately! Keep up the great work :)

PR #319, opened 2015-11-28
Upgrade to Gradle Wrapper 2.9

izeye:
BTW, I'm curious why `wrapper` task isn't used in `build.gradle` as follows:

```
task wrapper(type: Wrapper) {
    gradleVersion = '2.9'
}
```

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.9-rc-1-bin.zip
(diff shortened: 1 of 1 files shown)


rwinch:
Thanks for the PR! This is now merged into master. 

> BTW, I'm curious why wrapper task isn't used in build.gradle as follows:

I just find it easier to generate the wrapper using my current version of Gradle (it doesn't need the task actually defined in newer versions of Gradle).

I typically update the wrapper by modifying the properties file directly to avoid changing any updates that may have been made to things like the [default java options](https://github.com/spring-projects/spring-session/blob/3df40dd8a9ada28e3217c15482cfe1f74812fdf1/gradlew#L10)

Is there a reason you think it would be better defined in the build.gradle?

izeye:
@rwinch I just thought it's easier to change a version but didn't know the `gradlew` is customized. Thanks for the explanation :smile:

rwinch:
@izeye Not a problem on the explanation. Thanks again for the PR and for the extra set of eyes on things :)

PR #352, opened 2016-02-01
Remove accessing Hazelcast MapConfig

manderson23:
Fixes #339 

@rwinch and @shakuzen please review.

Issue #339, opened 2016-01-26
Support External Hazelcast instance (i.e. non-embedded)


sc-moonlight:
I could be missing something, but here's what I'm seeing.  I'm not sure if this can really be fixed, or if it just needs to be documented better that HazelcastHttpSessionConfiguration does not work for out of process Hazelcast server.

I enabled HazelcastHttpSessionConfiguration, and it works fine in a local profile that uses an embedded Hazelcast Instance.  However, when I switch to a profile that connects to a standalone Hazelcast instance, I get an exception thrown.

I'm setting up my hazelcast instance with this api:

```
    public static HazelcastInstance newHazelcastClient(ClientConfig config);
```

The first line in HazelcastHttpSessionConfiguration.configureSessionMap()  is 

```
		MapConfig sessionMapConfig = hazelcastInstance.getConfig().getMapConfig(sessionMapName);
```

then getConfig() throws this exception:

```
Caused by: java.lang.UnsupportedOperationException: Client cannot access cluster config!
          at com.hazelcast.client.impl.HazelcastClientInstanceImpl.getConfig(HazelcastClientInstanceImpl.java:278) ~[hazelcast-client-3.5.4.jar:3.5.4]
        at com.hazelcast.client.impl.HazelcastClientProxy.getConfig(HazelcastClientProxy.java:78) ~[hazelcast-client-3.5.4.jar:3.5.4]
        at org.springframework.session.hazelcast.config.annotation.web.http.HazelcastHttpSessionConfiguration.configureSessionMap(HazelcastHttpSessionConfiguration.java:98) ~[spring-session-1.1.0.M1.jar:na]
        at org.springframework.session.hazelcast.config.annotation.web.http.HazelcastHttpSessionConfiguration.sessionRepository(HazelcastHttpSessionConfiguration.java:70) ~[spring-session-1.1.0.M1.jar:na]
```

rwinch:
@sc-moonlight Why are you needing to access the the `MapConfig`? Can you post your entire Spring Session Configuration?

cc @shakuzen @manderson23

artembilan:
@rwinch , @sc-moonlight tells exactly about this code: https://github.com/spring-projects/spring-session/blob/master/spring-session/src/main/java/org/springframework/session/hazelcast/config/annotation/web/http/HazelcastHttpSessionConfiguration.java#L98.

rwinch:
@artembilan Sorry my mistake! I see it now. This appears to be a bug.

[...]

PR #356, opened 2016-02-08
Add integration tests for Hazelcast client use case

vpavic:
See #339, #352.

I've signed the CLA.


shakuzen:
:+1: Excellent work, @vpavic 
LGTM

rwinch:
Thank you for stepping up and providing an integration test @vpavic! I have merged this into master via 7c53558454a160e0443d73366ecc9e0d32211c3f

spring-session/src/integration-test/java/org/springframework/session/hazelcast/HazelcastClientRepositoryITests.java
```diff
@@ -0,0 +1,74 @@
...
+	private static final int PORT = SocketUtils.findAvailableTcpPort();
+
+	@BeforeClass
+	public static void setup() {
+		HazelcastITestUtils.embeddedHazelcastServer(PORT);
+	}
+
+	@AfterClass
+	public static void teardown() {
+		Hazelcast.shutdownAll();
```
artembilan: Uh! Just have noticed that.
From my perspective it should not be so bad at all:
```
 /**
     * Shuts down all running Hazelcast Instances on this JVM.
     * It doesn't shutdown all members of the
     * cluster but just the ones running on this JVM.
     *
     * @see #newHazelcastInstance(Config)
     */
    public static void shutdownAll() {
```
But from other side I would be consistent and close only that `HazelcastInstance` which we have started with the `HazelcastITestUtils.embeddedHazelcastServer`.

spring-session/src/integration-test/java/org/springframework/session/hazelcast/HazelcastClientRepositoryITests.java
```diff
@@ -0,0 +1,74 @@
...
+ * Integration tests that check the underlying data source - in this case
+ * Hazelcast Client.
+ *
+ * @author Vedran Pavic
+ */
+@RunWith(SpringJUnit4ClassRunner.class)
+@ContextConfiguration
+@WebAppConfiguration
+public class HazelcastClientRepositoryITests<S extends ExpiringSession>
+		extends AbstractHazelcastRepositoryITests<S> {
```
artembilan: I don't like all this `I` notations in the class names. Feels like the C++ artifacts, but that is just IMHO.
Looks like Hazelcast doesn't feel itself bad to use it in its types like `IMap` :smile:

artembilan:
Otherwise, yes, of course, the good job!
And thank you everyone for contribution!

spring-session/src/integration-test/java/org/springframework/session/hazelcast/HazelcastClientRepositoryITests.java
rwinch: @artembilan This is a convention that is used in Spring Session to distinguish between integration tests and unit tests. Otherwise we will get collisions in the class names.

artembilan: Thank you @rwinch . Got it! Makes sense: `ITest` == Integration Test.
Never mind - the matter of opinion :smile:

spring-session/src/integration-test/java/org/springframework/session/hazelcast/HazelcastClientRepositoryITests.java
rwinch: @artembilan Thanks for the feedback! That is a good point :) I created #358 Any chance you could send a PR?

rwinch:
@artembilan Thanks for the review!

spring-session/src/integration-test/java/org/springframework/session/hazelcast/HazelcastClientRepositoryITests.java
rwinch: @artembilan Not a problem at all! All feedback is welcome...keep it coming :)

Issue #358, opened 2016-02-08
HazelcastClientRepositoryITests should shutdown single instance not all


rwinch:
We should close a [single Hazelcast instance rather than all of them](https://github.com/spring-projects/spring-session/blob/7c53558454a160e0443d73366ecc9e0d32211c3f/spring-session/src/integration-test/java/org/springframework/session/hazelcast/HazelcastClientRepositoryITests.java#L55-L58)

artembilan:
I'd be glad to contribute @rwinch :smile: .

Will consider to PR this evening or tomorrow's morning.

rwinch:
@artembilan Thank you so much!

PR #360, opened 2016-02-09
GH-358 Tests: Close only used `HazelcastInstance`

artembilan:
Fixes GH-358 (https://github.com/spring-projects/spring-session/issues/358)

* It might be dangerous to `shutdownAll()` `HazelcastInstance`s from the test-case, especially when we are ran in with the integration test and on the CI environment.
Therefore extract the `instance` variable in the test and close only it from the `@AfterClass`
* Fix typo in the `AbstractHazelcastRepositoryITests` test method name
* Increase lock wait timeout to the 10 seconds in the `SessionEventRegistry`. Looks like 3 seconds isn't enough for my Windows machine.


vpavic:
:+1:

rwinch:
Thanks for the PR! This is now merged into master via bf7729b I added very minor polish in 5f23b3c to ensure that the hazelcastInstance wasn't null in the teardown method.

Issue #364, opened 2016-02-11
Spring Session Relational Database Support


rwinch:
It would be great to have Spring Session support with relational databases.

Most likely session attributes would be saved as serializable blobs since session attributes are unlikely to provide JPA mappings related to the session. For this reason, it may make sense to use straight JDBC rather than JPA. This part is up for discussion.

What we would typically look for:

* [ ] Code (obviously)
* [ ] Unit tests
* [ ] Integration tests (fewer than the unit tests for speed but still ensure things are working)
* [ ] Hello World Sample / Guide for Web Application (i.e. [HttpSession Sample](http://docs.spring.io/spring-session/docs/current/reference/html5/guides/httpsession.html))
* [ ] Sample Tests
* [ ] Optional `@EnableJdbcHttpSession` for making things easier
* [ ] Optional spring-session-jdbc pom that aggregates the dependencies to make dependency management for users easier

vpavic:
@rwinch this is a really great proposal! :+1:

Having support for relational database as session store would surely aid more widespread adoption of Spring Session, especially in conservative environments which heavily rely on a RDBMS.

I'm willing to work on a PR to provide this functionality. Are there any specific milestones/timelines for this feature?

chrylis:
Is there a particular reason to focus this on JPA instead of the (usual) underlying JDBC, since the session will end up as a blob anyhow? The only advantage that comes to mind is supporting nontraditional JPA providers (e.g., BigTable), and the management and performance overhead of JPA would be noticeable.

rwinch:
@vpavic That is great news!

> I'm willing to work on a PR to provide this functionality. Are there any specific milestones/timelines for this feature?

Given this would be contributed by the community (sounds like you) it will make it in when you get it done. I've updated the comment to include things that we would look for for this to be considered done. Of course if you only have time for one item, we can always divide the work up :)

rwinch:
@chrylis Thank you for your feedback!

> Is there a particular reason to focus this on JPA instead of the (usual) underlying JDBC, since the session will end up as a blob anyhow

This is a good point! I updated the description 

cc @vpavic

vpavic:
@rwinch I agree with @chrylis's suggestion that the implementation should be based on ```spring-jdbc``` instead of ```spring-data-jpa```. Why force users into JPA if they don't use/want it? That could present a deal breaker for some.

[...]

PR #372, opened 2016-02-20
Extract PrincipalNameResolver

vpavic:
While working on a PR for #364, I've noticed that I also required the same functionality that ```RedisOperationsSessionRepository``` implements in its inner ```PrincipalNameResolver```. Further inspection showed that ```AbstractGemFireOperationsSessionRepository``` has the same functionality implemented internally.

To avoid code duplication this PR extracts ```PrincipalNameResolver``` as a strategy interface and provides the default implementation based on Spring Security (effectively the same to the current implementation) and a test class.

I've signed the CLA.

### DIFF
--- a/spring-session/src/main/java/org/springframework/session/PrincipalNameResolver.java
+++ b/spring-session/src/main/java/org/springframework/session/PrincipalNameResolver.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2002-2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.session;
+
+/**
+ * Strategy interface used for resolving principal name from the {@link Session}.
+ *
+ * @author Vedran Pavic
+ */
+public interface PrincipalNameResolver {
+
+	/**
+	 * Resolve the principal name from the given {@link Session}
+	 * @param session the session
+	 * @return the principal name
+	 */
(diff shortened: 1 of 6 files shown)


vpavic:
Few related observations:

Even though the ```FindByIndexNameSessionRepository``` javadoc claims that _Spring Session is not aware of the authentication mechanism being used_, concrete implementations are still somewhat tied to Spring Security - I've made it possible to set a custom ```PrincipalNameResolver``` in ```RedisOperationsSessionRepository```, however ```saveDelta``` method still relies on ```SPRING_SECURITY_CONTEXT``` key.

```AbstractGemFireOperationsSessionRepository``` requires some refactoring to make it possible to customize ```PrincipalNameResolver```. If this PR is accepted I can tackle this some time later (if Gemfire integration author cannot do it).

rwinch:
@vpavic Thanks for the PR! 

While I can see that this looks like a good idea, I don't think we want to do this just yet. The reason we have duplicated the logic is because we want to keep the implementation details private (so they can change later).

You may notice that right now the API is specific to resolving a principal name. However, `FindByIndexNameSessionRepository` supports finding by a generic index name. This means we would likely want to update the API to support generic index names. However, we will not have enough time to do this before things go GA.

Another consideration is that we would like to see what shapes up from the [Redis support](https://github.com/spring-projects/spring-data-redis/pull/156/files#diff-d2c04ffe360655561105d34b8ba1d122R31). We may find we can simply use some Spring Data Commons support rather than providing our own (I'm currently in talks with the Spring Data Team about this idea).

For these reasons, I'm going to close this PR and in favor of #376 for us to address this once we know what is happening within Spring Data. Thank you for pointing out this was undocumented issue!

Thanks again for the PR (sorry the issue wasn't documented)! I'm looking forward to seeing support for #364

vpavic:
@rwinch thanks for the insight!

I understand your reasoning, the changes provided with this PR would've been a more natural fit with ```FindByIndexNameSessionRepository```'s predecessor - ```FindByPrincipalNameSessionRepository```. Since a move to more generic index name base repository is made, the same should apply to the proposed Resolver.

I'd be happy to make another contribution when #376 shapes up with more concrete details.

PR to address #364 is coming soon, I'll just duplicate ```PrincipalNameResolver``` logic.

rwinch:
@vpavic Thanks! I'll try and remember to bump you when we know what path to take.

Looking forward to that PR! With this and the Mongo support I think that a 1.2 is going to be coming before we know it!

Issue #314, opened 2015-11-26
A new thread is created on every request when @EnableRedisHttpSession is used.


tsachev:
`RedisHttpSessionConfiguration` creates `RedisMessageListenerContainer` which by
default uses `SimpleAsyncTaskExecutor` to dispatch messages, which means a
new thread is created for every single message received by Redis.
`RedisOperationsSessionRepository` (previously `SessionMessageListener`) is
subscribed to `__keyevent@*:del` pattern.
`RedisSessionExpirationPolicy` sends a `del` event for expirations key,
every time the session is saved (those can be reduced sometimes, but I will discuss this in another issue).
And though those events are ignored by the listener, because it does not care about
expirations keys (`spring:session:expirations:XXXX`) a thread is still created.

So the effect is that a new thread is created for every request.
If the subscription pattern cannot be more specific (i.e. somehow specify that we do not want events for expirations),
then more sensible executors should be configured on the listener container.

`RedisMessageListenerContainer` has `subscriptionExecutor` and `taskExecutor`.
`subscriptionExecutor` defaults to `taskExecutor` if unset, and `taskExecutor` defaults to the bad `SimpleAsyncTaskExecutor`.
It is worth nothing that `subscriptionExecutor` cannot be bounded
thread pool with size < 2, because it starts a long running thread.

Here is one possible work-araound

```
@Configuration
public class Config extends RedisHttpSessionConfiguration {
  @Bean
  public TaskExecutor sessionRedisListenerContainerTaskExecutor() {
    ThreadPoolTaskExecutor taskExecutor = new ThreadPoolTaskExecutor();
    taskExecutor.setThreadNamePrefix("SessionRedisListenerContainer-");
    taskExecutor.setMaxPoolSize(4);
    taskExecutor.setKeepAliveSeconds(serverProperties.getSession().getTimeout());
    return taskExecutor;
  }

  @Bean
  @Override
  public RedisMessageListenerContainer redisMessageListenerContainer(
                  RedisConnectionFactory connectionFactory, RedisOperationsSessionRepository messageListener) {
    RedisMessageListenerContainer container = super.redisMessageListenerContainer(connectionFactory, messageListener);
    container.setTaskExecutor(sessionRedisListenerContainerTaskExecutor());
    return container;
  }
}
```

But probably creating a `SimpleAsyncTaskExecutor` for `subscriptionExecutor` and
bounded thead pool for `taskExecutor` with configurable `maxThreads` is better.

[...]

PR #374, opened 2016-02-22
Allow overriding task executor(s) used by `RedisMessageListenerContainer`.

tsachev:
Fix for gh-314


rwinch:
Thanks for the PR! This is now merged via 5bf12fdd93

PR #379, opened 2016-02-22
Implement support for relational databases

vpavic:
This commit provides implementation of ```SessionRepository``` based on Spring's ```JdbcOperations``` interface.

I believe I've covered all the action points listed in #364, and will also update the docs after the implementation has been reviewed and polished.

Looking forward to comments and improvement suggestions.


spring-session/src/main/java/org/springframework/session/jdbc/JdbcOperationsSessionRepository.java
```diff
@@ -0,0 +1,439 @@
...
+	@Scheduled(cron = "0 * * * * *")
+	public void cleanupExpiredSessions() {
+		long now = System.currentTimeMillis();
+		long roundedNow = roundDownMinute(now);
+
+		if (logger.isDebugEnabled()) {
+			logger.debug("Cleaning up sessions expiring at " + new Date(roundedNow));
+		}
+
+		List<String> sessionIds = this.jdbcOperations.queryForList(
```
rwinch: Rather than first querying and then deleting this could just be a delete which would prevent loading all the session ids in memory and reduce overhead on the database.

spring-session/src/main/java/org/springframework/session/jdbc/JdbcOperationsSessionRepository.java
```diff
@@ -0,0 +1,439 @@
...
+
+	private static final Log logger = LogFactory.getLog(JdbcOperationsSessionRepository.class);
+
+	public static final String DEFAULT_TABLE_PREFIX = "SPR_";
+
+	private static final String SPRING_SECURITY_CONTEXT = "SPRING_SECURITY_CONTEXT";
+
+	private static PrincipalNameResolver PRINCIPAL_NAME_RESOLVER = new PrincipalNameResolver();
+
+	private enum Query {
```
rwinch: I think I would prefer to use private static final Strings here

spring-session/src/main/java/org/springframework/session/jdbc/JdbcOperationsSessionRepository.java
```diff
@@ -0,0 +1,439 @@
...
+
+		return query;
+	}
+
+	public void delete(String id, SessionDestroyedEvent event) {
+		this.jdbcOperations.update(getQuery(Query.DELETE_SESSION), id);
+		publishEvent(event);
+	}
+
+	private void publishEvent(ApplicationEvent event) {
```
rwinch: Unfortunately, this will not publish the ApplicationEvent for every node in the cluster. For this reason, I think we should just not publish the event and establish this as a limitation of the datastore.

This ties in with why we can simply perform a delete rather than looking up the session id first.

spring-session/src/main/java/org/springframework/session/jdbc/JdbcOperationsSessionRepository.java
rwinch: NOTE: See https://github.com/spring-projects/spring-session/pull/379/files#r53860075

vpavic: I initially considered implementing it that way for the same reasons, but that would mean ```SessionExpiredEvent```s wouldn't be fired.

rwinch:
Thanks for the PR! I have provided a few comments, but will need to revisit this once we start work on 1.2. This should be rather shortly as we hope to have 1.1 out this week (depending on one last issue getting resolved)

spring-session/src/main/java/org/springframework/session/jdbc/JdbcOperationsSessionRepository.java
vpavic: Agreed, not sure how and why this one slipped through as ```public```.

spring-session/src/main/java/org/springframework/session/jdbc/JdbcOperationsSessionRepository.java
rwinch: @vpavic Thanks. See https://github.com/spring-projects/spring-session/pull/379/files#r53860177

spring-session/src/main/java/org/springframework/session/jdbc/JdbcOperationsSessionRepository.java
rwinch: I should be more clear. I'm not sure I like the enum. I would only use Strings as this aligns with Spring coding conventions.

spring-session/src/main/java/org/springframework/session/jdbc/JdbcOperationsSessionRepository.java
vpavic: OK, this approach also covers the cleanup strategy then.

spring-session/src/main/java/org/springframework/session/jdbc/JdbcOperationsSessionRepository.java
vpavic: Ahh, I see... you'd like to have something similar to way Security's ```JdbcDaoImpl``` provides SQL customization options.

I also considered this initially, but opted to go the route seen in Integration's ```JdbcMessageStore```.

My reasoning was that in practice the ```Session``` object is closer to Integration's ```Message``` than it is to Security's ```UserDetails``` due to ```Session``` being entity which is internal to the framework unlike ```UserDetails``` which is often derived and constructed from some application domain entity provided by users. That's why the ```enum``` approach seemed like a good approach to me.

artembilan: @rwinch, looks like the solution is stolen from the Spring Integration: https://github.com/spring-projects/spring-integration/blob/master/spring-integration-jdbc/src/main/java/org/springframework/integration/jdbc/JdbcMessageStore.java#L94.

Not sure if we really have some gain from the `enum`, but that might be some `hash` performance producing for the `HashMap` key...

vpavic:
Thanks for the review @rwinch. I'll update the PR accordingly over the next few days.

spring-session/src/main/java/org/springframework/session/jdbc/JdbcOperationsSessionRepository.java
vpavic: @artembilan stolen is such a harsh word :)

Previous comment clearly stated the origin of the approach taken here, as well as [this comment](https://github.com/spring-projects/spring-session/issues/364#issuecomment-183479733) did in the original issue. If something is a valid and proven approach to a given problem why not reuse it?

artembilan: @vpavic , sorry if that really sounded so hurting.
Really wasn't going to say anything bad nor to you, nor to the approach. 

Just was going to explain to @rwinch that there might no reason to nit-pick to this code.

From other side `steal` isn't applied between Spring projects. You use code from one Spring project in the another, therefore this word should be treated as enough soft.

Anyway I got your point and will use the more formal language in the future. :smile:

vpavic: No harm done @artembilan, that's why I used :) in the previous comment.
We're on the same page.

vpavic:
@rwinch I've updated the PR with following changes:

- cleanup of expired sessions is now done using a single SQL statement (delete)
- support for publishing ```SessionDestroyedEvent``` and its subclasses has been removed
- ```JdbcOperationsSessionRepository``` constructor which takes a ```DataSource``` and constructs default ```JdbcTemplate``` has been added
- tests have been updated according to aforementioned changes
- javadoc updates
- fixed integration tests for new sample projects

SQL statements are still in ```enum```, I'd like to see your feedback on my latest comments on this subject.

copa2:
@vpavic thanks for this PR.

* Use ConcurrentHashMap for the queryCache. Access and fill can be concurrent. 
* For what is the %PREFIX%? When then I would allow to change the whole table name, not just the prefix.

Personally I would not use the enum approach, just simple String which can be overwritten.

vpavic:
@copa2 thanks for your feedback.

- Since ```tablePrefix``` is not intended to be changed in runtime, concurrency is not an issue. However, synchronizing ```queryCache``` operations inside ```getQuery``` method might be a good idea.
- Prefix allows for a limited customization of table name. This is a common approach used in many frameworks, for example Spring Integration, Spring Batch, Quartz Scheduler, etc. to avoid possible name clashes with other entities in your app and make it possible to have multiple repo's backed by the same database schema.

vpavic:
I've updated the PR again, with the following changes:

- moved SQL scripts to ```org/springframework/session/jdbc/``` and updated references accordingly
- polished modifiers a bit in ```JdbcOperationsSessionRepository```

settings.gradle
```diff
@@ -21,3 +22,7 @@ include 'samples:websocket'
 include 'spring-session'
 include 'spring-session-data-gemfire'
 include 'spring-session-data-redis'
+include 'spring-session-jdbc'
+include 'samples:httpsession-jdbc-xml'
+findProject(':samples:httpsession-jdbc-xml')?.name = 'httpsession-jdbc-xml'
```
rwinch: Is this necessary?

vpavic: Certainly not, it seems like IntelliJ's Gradle plugin's messing with Gradle conf.
I remember fixing something similar in ```settings.gradle``` prior to pushing the initial commit but it slipped in the updated commit.

rwinch:
@vpavic Thanks again for sending this PR! I'm starting a full review of this PR now

rwinch:
@vpavic 

> SQL statements are still in enum, I'd like to see your feedback on my latest comments on this subject.

[You said](https://github.com/spring-projects/spring-session/pull/379#discussion_r53863096)

>> Ahh, I see... you'd like to have something similar to way Security's JdbcDaoImpl provides SQL customization options.
>> 
>> I also considered this initially, but opted to go the route seen in Integration's JdbcMessageStore.
>> 
>> My reasoning was that in practice the Session object is closer to Integration's Message than it is to Security's UserDetails due to Session being entity which is internal to the framework unlike UserDetails which is often derived and constructed from some application domain entity provided by users. That's why the enum approach seemed like a good approach to me.

Thanks for the response. I would still like these to be static final Strings rather than enums.

I see where you got the idea from. I know this is a coding style change (especially since this is all private). However, making this change will better align with Spring Session's code style and provide consistency in the project. Additionally, it will make it easier to allow custom queries w/ very little effort if we every go down that path.

spring-session/src/main/java/org/springframework/session/jdbc/JdbcOperationsSessionRepository.java
```diff
@@ -0,0 +1,483 @@
...
+
+	private final JdbcOperations jdbcOperations;
+
+	private final RowMapper<MapSession> mapper = new SessionMapper();
+
+	private final Map<Query, String> queryCache = new HashMap<Query, String>();
+
+	private ApplicationEventPublisher applicationEventPublisher = new ApplicationEventPublisher() {
+
+		@Override
```
rwinch: `@Override' on interface methods won't compile in Eclipse. Please remove this.

spring-session/src/main/java/org/springframework/session/jdbc/JdbcOperationsSessionRepository.java
```diff
@@ -0,0 +1,483 @@
...
+
+	}
+
+	private final JdbcOperations jdbcOperations;
+
+	private final RowMapper<MapSession> mapper = new SessionMapper();
+
+	private final Map<Query, String> queryCache = new HashMap<Query, String>();
+
+	private ApplicationEventPublisher applicationEventPublisher = new ApplicationEventPublisher() {
```
rwinch: If we keep this, need to add `publishEvent` to work with Spring 4.2.x (which we now compile with).

spring-session/src/main/java/org/springframework/session/jdbc/JdbcOperationsSessionRepository.java
```diff
@@ -0,0 +1,483 @@
...
+ *
+ * @author Vedran Pavic
+ * @since 1.2.0
+ */
+public class JdbcOperationsSessionRepository
+		implements FindByIndexNameSessionRepository<JdbcOperationsSessionRepository.JdbcSession>, ApplicationEventPublisherAware {
+
+	private static final Log logger = LogFactory.getLog(JdbcOperationsSessionRepository.class);
+
+	private static final String DEFAULT_TABLE_PREFIX = "SPR_";
```
rwinch: I agree with @copa2 about [the table prefix](https://github.com/spring-projects/spring-session/pull/379#issuecomment-189323488).

If you need to allow customizing the queries, then allow the entire query to be injected.

spring-session/src/main/java/org/springframework/session/jdbc/JdbcOperationsSessionRepository.java
```diff
@@ -0,0 +1,483 @@
...
+					ps.setString(1, session.getId());
+					ps.setLong(2, session.getLastAccessedTime());
+					ps.setString(3, session.getPrincipalName());
+					JdbcOperationsSessionRepository.this.lobHandler.getLobCreator()
+							.setBlobAsBytes(ps, 4, sessionBytes);
+				}
+
+			});
+			session.setNew(false);
+			publishEvent(new SessionCreatedEvent(this, session));
```
rwinch: This won't be available to all JVM instances which is inconsistent with the way Spring Session works. We should probably remove this.

spring-session/src/main/java/org/springframework/session/jdbc/JdbcOperationsSessionRepository.java
```diff
@@ -0,0 +1,483 @@
...
+
+	/**
+	 * Replace patterns in the input to produce a valid SQL query. This implementation lazily initializes a
+	 * simple map-based cache, only replacing the table prefix on the first access to a named query. Further
+	 * accesses will be resolved from the cache.
+	 *
+	 * @param base the SQL query to be transformed
+	 * @return a transformed query with replacements
+	 */
+	protected String getQuery(Query base) {
```
rwinch: If we change to using Strings for the queries, then this can be simplified (i.e. we won't need queryCache). We won't need to [worry about concurrent access to queryCache](https://github.com/spring-projects/spring-session/pull/379#issuecomment-189323488), etc.

spring-session/src/main/java/org/springframework/session/jdbc/JdbcOperationsSessionRepository.java
```diff
@@ -0,0 +1,483 @@
...
+	 */
+	private String tablePrefix = DEFAULT_TABLE_PREFIX;
+
+	/**
+	 * If non-null, this value is used to override the default value for
+	 * {@link JdbcSession#setMaxInactiveIntervalInSeconds(int)}.
+	 */
+	private Integer defaultMaxInactiveInterval;
+
+	private DeserializingConverter deserializer = new DeserializingConverter();
```
rwinch: We should probably use interface type declaration for deserializer and serializer

spring-session/src/main/java/org/springframework/session/jdbc/JdbcOperationsSessionRepository.java
```diff
@@ -0,0 +1,483 @@
...
+ * @since 1.2.0
+ */
+public class JdbcOperationsSessionRepository
+		implements FindByIndexNameSessionRepository<JdbcOperationsSessionRepository.JdbcSession>, ApplicationEventPublisherAware {
+
+	private static final Log logger = LogFactory.getLog(JdbcOperationsSessionRepository.class);
+
+	private static final String DEFAULT_TABLE_PREFIX = "SPR_";
+
+	private static final String SPRING_SECURITY_CONTEXT = "SPRING_SECURITY_CONTEXT";
```
rwinch: Move this to PrincipalNameResolver since that is the only place it is used

spring-session/src/main/java/org/springframework/session/jdbc/JdbcOperationsSessionRepository.java
rwinch: I'd also default the table to SPRING_SESSION to be more explicit

spring-session/src/main/java/org/springframework/session/jdbc/JdbcOperationsSessionRepository.java
```diff
@@ -0,0 +1,483 @@
...
+	}
+
+	@SuppressWarnings("unchecked")
+	public void setSerializer(Serializer<MapSession> serializer) {
+		Assert.notNull(serializer, "Serializer must not be null");
+		this.serializer = new SerializingConverter((Serializer) serializer);
+	}
+
+	@SuppressWarnings("unchecked")
+	public void setDeserializer(Deserializer<MapSession> deserializer) {
```
rwinch: Change to Converter<byte[], Object>

spring-session/src/main/java/org/springframework/session/jdbc/JdbcOperationsSessionRepository.java
```diff
@@ -0,0 +1,483 @@
...
+		this.defaultMaxInactiveInterval = defaultMaxInactiveInterval;
+	}
+
+	public void setLobHandler(LobHandler lobHandler) {
+		Assert.notNull(lobHandler, "LobHandler must not be null");
+		this.lobHandler = lobHandler;
+	}
+
+	@SuppressWarnings("unchecked")
+	public void setSerializer(Serializer<MapSession> serializer) {
```
rwinch: Change argument type to `Converter<Object, byte[]>`

spring-session/src/main/java/org/springframework/session/jdbc/JdbcOperationsSessionRepository.java
```diff
@@ -0,0 +1,483 @@
...
+import org.springframework.scheduling.annotation.Scheduled;
+import org.springframework.session.ExpiringSession;
+import org.springframework.session.FindByIndexNameSessionRepository;
+import org.springframework.session.MapSession;
+import org.springframework.session.Session;
+import org.springframework.session.events.SessionCreatedEvent;
+import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
+
+/**
```
rwinch: At some point in the javadoc we should probably include the default schema that should be used with this.

spring-session/src/main/java/org/springframework/session/jdbc/JdbcOperationsSessionRepository.java
```diff
@@ -0,0 +1,483 @@
...
+		JdbcSession session = new JdbcSession();
+		if (this.defaultMaxInactiveInterval != null) {
+			session.setMaxInactiveIntervalInSeconds(this.defaultMaxInactiveInterval);
+		}
+		return session;
+	}
+
+	@Override
+	public void save(final JdbcSession session) {
+		final byte[] sessionBytes = this.serializer.convert(session.delegate);
```
rwinch: It would probably be good to have some sort of optimization to allow only updating the lastUpdateTime if that is all that was changed. This would be beneficial so that the entire Session does not need to be serialized every time.

spring-session/src/main/java/org/springframework/session/jdbc/JdbcOperationsSessionRepository.java
```diff
@@ -0,0 +1,483 @@
...
+
+		public String getSql() {
+			return this.sql;
+		}
+
+	}
+
+	private final JdbcOperations jdbcOperations;
+
+	private final RowMapper<MapSession> mapper = new SessionMapper();
```
rwinch: Change to `RowMapper<ExpiringSession>` (this will involve changing all the references to `MapSession` too).

spring-session/src/main/java/org/springframework/session/jdbc/config/annotation/web/http/JdbcHttpSessionConfiguration.java
```diff
@@ -0,0 +1,126 @@
...
+
+	private Deserializer<MapSession> deserializer;
+
+	@Bean
+	public JdbcTemplate sessionJdbcTemplate(DataSource dataSource) {
+		return new JdbcTemplate(dataSource);
+	}
+
+	@Bean
+	public JdbcOperationsSessionRepository sessionRepository(@Qualifier("sessionJdbcTemplate") JdbcOperations jdbcOperations) {
```
rwinch: Let's use `springSessionJdbcOperations` as the qualifier.

spring-session/src/main/java/org/springframework/session/jdbc/JdbcOperationsSessionRepository.java
vpavic: Yeah, I figured that out myself in the meantime.

spring-session/src/main/java/org/springframework/session/jdbc/JdbcOperationsSessionRepository.java
vpavic: Ok, so assuming we take the Strings for queries route, you'd like to also remove the possibility of changing just table prefix?

In that case, users would need to provide custom queries just to the change table prefix. Consider the case where there are multiple apps sharing the same database schema - in order to avoid conflicts only the table prefix needs to be changed. Forcing users to provide custom queries in that case is a configuration overkill IMO.

If we keep table prefix option, we need some sort of ```getQuery``` method, and a cache would certainly be beneficial to prevent ```String``` manipulation on each ```SessionRepository``` operation.

vpavic:
@rwinch thanks for the full review, I'll update the PR with your suggestions within a day or two.

spring-session/src/main/java/org/springframework/session/jdbc/JdbcOperationsSessionRepository.java
rwinch: > In that case, users would need to provide custom queries just to the change table prefix. Consider the case where there are multiple apps sharing the same database schema - in order to avoid conflicts only the table prefix needs to be changed. Forcing users to provide custom queries in that case is a configuration overkill IMO.

We probably wouldn't advice placing something like session in a shared database (although this would still be possible with a little bit of work). Typically different applications would use different username / password for their database connection. The different users could have different default schema which would isolate the sessions.

We could certainly create a separate ticket for adding a table prefix and see if it gets enough votes, but don't think we'd add something like this in an initial release (once we add something it cannot be removed).

vpavic: @rwinch we're very often in situation where customer owns the database and all we get is a schema with limited privileges, meaning we have multiple services that share the same schema. I agree it's a not great situation but its completely out our control.

Therefore I think allowing users to customize the table prefix is not some exotic requirement but rather a failry common thing we heavily rely on with other frameworks we use regulary:
- [Spring Integration](http://docs.spring.io/spring-integration/docs/4.2.5.RELEASE/reference/html/jdbc.html#jdbc-message-store-generic)
- [Spring Batch](http://docs.spring.io/spring-batch/trunk/reference/html/configureJob.html#repositoryTablePrefix)
- [Quartz Scheduler](http://quartz-scheduler.org/generated/2.2.1/html/qs-all/#page/Quartz_Scheduler_Documentation_Set%2Fre-jtx_jdbc_jobstoretx_configuration.html%23)

Not to drag this any further, I'll update the PR soon according to your preferences but expect another PR to provide the table prefix customization as soon as you merge this one :)

rwinch: @vpavic Thank you for your response. Would a table name work (vs just a prefix)? We could have a setter that contains modifies the table name of every query (only if they are using the default queries).

vpavic: @rwinch yes, of course, that would be great.

vpavic:
@rwinch I've updated the PR with following changes:
- move queries from ```enum``` to ```String``` constants
- allowed default table name to be customized, instead of just table prefix
- completely removed event publishing
- use ```Converter``` interface for session (de)serialization customization
- replace ```MapSession``` references with ```ExpiringSession``` interface
- other misc improvements from your comments 
- rebased to the current master

I still owe you the persistence optimization in cases when only ```lastAccessedTime``` is updated. I need to think a bit about that one, in the meantime you can review other changes.

rwinch:
@vpavic Thanks for the updates. 

> I still owe you the persistence optimization in cases when only lastAccessedTime is updated. I need to think a bit about that one, in the meantime you can review other changes.

You can have a boolean on the `JdbcSession` that is changed when any property (other than lastAccesedTime) is set. If the boolean changes then you would persist everything, else persist only the lastAccessedTime. In fact, you might keep track of if lastAccessedTime is written too and only persist if that changes.

vpavic:
> You can have a boolean on the JdbcSession that is changed when any property (other than lastAccesedTime) is set. If the boolean changes then you would persist everything, else persist only the lastAccessedTime. In fact, you might keep track of if lastAccessedTime is written too and only persist if that changes.

Something along those lines (```boolean``` flags) was my initial idea also. Thanks.

vpavic:
@rwinch, I've updated the PR with improved/optimized persistence strategy. Test cases to cover all the scenarios on ```save``` operation were also added.

rwinch:
Have you [signed the CLA](https://github.com/spring-projects/spring-session/blob/master/CONTRIBUTING.adoc#sign-the-contributor-license-agreement)?

vpavic:
Whoops, sorry forgot to mention it in the opening comment :)
Yes, I've signed the CLA.

I also did some work on the docs for this new feature but didn't finish it yet.

rwinch:
@vpavic Thanks! I applied some very minor polish, but other than that I think things are ready to merge. You mention:

> I also did some work on the docs for this new feature but didn't finish it yet.

Can you elaborate? Is this something that we need to wait on or can I merge now?

vpavic:
@rwinch I was referring to documentation (docs module).

No need to wait for merging, I can open a new PR for the docs if that's OK with you.

rwinch:
@vpavic Thanks for the clarification. I just wanted to make sure we were both ready for the merge. I'm putting some final polish on things and will merge once that is done.

rwinch:
@vpavic I created #364 for the documentation of the samples. Ideally, we will have the documentation before we perform a release.

rwinch:
Thanks again for the PR! This is now merged via cd38e307e00 I applied some polish via 8f8cfe5d79e12 and 7b28b214ffa4925497dac3c77e9d54b874bbe89f

Issue #373, opened 2016-02-22
ClassCastException in AbstractGemFireOperationsSessionRepository.afterCreate


sshcherbakov:
I'm trying to setup a simple Spring security session sharing using spring-session and Gemfire.
I've added following dependencies to my webapp project:
````
		compile('org.springframework.session:spring-session-data-gemfire:1.1.0.RC1')
		compile('org.springframework.session:spring-session:1.1.0.RC1')
		compile('org.springframework.security:spring-security-web')
		compile('org.springframework.security:spring-security-config')
````
and almost succeeded to setup session sharing. 
I had to add several Spring dependencies to the Gemfire classpath because the spring-session-data-gemfire doesn't use PDX and thus some classes need to be present on the server side as well:
`spring-session-1.1.0.RC1.jar`,`spring-expression-4.2.4.RELEASE.jar`, `spring-security-web-4.0.3.RELEASE.jar`, `spring-security-core-4.0.3.RELEASE.jar`
The session gets stored in my Gemfire cluster and I don't get server side ClassNotFoundExceptions any more.
I also get a simple authentication prompt asking to login to my application:
````
@EnableWebSecurity
public class SecurityConfig extends WebSecurityConfigurerAdapter {

	@Autowired
	public void configureGlobal(AuthenticationManagerBuilder auth) throws Exception {
		auth
			.inMemoryAuthentication()
				.withUser("user").password("password").roles("USER");
	}
}
````
On the application side after I try to login with the those credentials I get the following exception:
````
[error 2016/02/22 02:29:59.823 CET <http-nio-8080-exec-2> tid=0x31] Exception occurred in CacheListener
java.lang.ClassCastException: com.gemstone.gemfire.internal.cache.Token$Tombstone cannot be cast to org.springframework.session.ExpiringSession
	at org.springframework.session.data.gemfire.AbstractGemFireOperationsSessionRepository.afterCreate(AbstractGemFireOperationsSessionRepository.java:212)
	at com.gemstone.gemfire.internal.cache.EnumListenerEvent$AFTER_CREATE.dispatchEvent(EnumListenerEvent.java:97)
	at com.gemstone.gemfire.internal.cache.LocalRegion.dispatchEvent(LocalRegion.java:8820)
	at com.gemstone.gemfire.internal.cache.LocalRegion.dispatchListenerEvent(LocalRegion.java:7311)
[...]

PR #383, opened 2016-02-24
Fix ClassCastException Token$Tombstone 

jxblum:
Fixes for gh-373

### DIFF
--- a/samples/security/build.gradle
+++ b/samples/security/build.gradle
@@ -8,6 +8,7 @@ sonarRunner {
 
 dependencies {
 	compile project(':spring-session-data-redis'),
+			project(':spring-session-data-gemfire'),
 			"org.springframework:spring-web:$springVersion",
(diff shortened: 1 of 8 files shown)


rwinch:
@jxblum Thanks for the PR! I'm not sure we want to modify the samples to use GemFire. We can always add more samples but we do not want to change the existing samples.

jxblum:
@rwinch you can discard the commits for the changes to the Security/Users samples.  Or, it maybe nice to include different configurations for _Spring Session's_ supported stores and provide instructions on how to change stores when running the samples.  Of course, the default could be to use Redis and preserve the work you have already done.  Either way; no worries.

rwinch:
@jxblum Thanks. I just wanted to make sure I wasn't missing something. I will just use ee38337

rwinch:
Thanks for the PR! This is now merged into master via 53d7c84f73be1df4d

Issue #396, opened 2016-02-28
GemFireHttpSessionJavaConfigurationTests.assertCacheAndRegion fails


vpavic:
Gradle task ```:spring-session:integrationTest``` is failing for me due to 3 failed test in ```GemFireHttpSessionJavaConfigurationTests```, with ```assertCacheAndRegion``` appearing to be the cause.

Here are more details:

```bash
$ ./gradlew :spring-session:integrationTest
Java HotSpot(TM) 64-Bit Server VM warning: ignoring option MaxPermSize=512M; support was removed in 8.0
The 'sonar-runner' plugin has been deprecated and is scheduled to be removed in Gradle 3.0. please use the official plugin from SonarQube (http://docs.sonarqube.org/display/SONAR/Analyzing+with+Gradle).
:spring-session:compileJava UP-TO-DATE
:spring-session:compileGroovy UP-TO-DATE
:spring-session:processResources UP-TO-DATE
:spring-session:classes UP-TO-DATE
:spring-session:compileTestJava UP-TO-DATE
:spring-session:compileTestGroovy UP-TO-DATE
:spring-session:processTestResources UP-TO-DATE
:spring-session:testClasses UP-TO-DATE
:spring-session:compileIntegrationTestJava
warning: [options] bootstrap class path not set in conjunction with -source 1.5
warning: [options] source value 1.5 is obsolete and will be removed in a future release
warning: [options] target value 1.5 is obsolete and will be removed in a future release
warning: [options] To suppress warnings about obsolete options, use -Xlint:-options.
:spring-session:compileIntegrationTestGroovy UP-TO-DATE
:spring-session:processIntegrationTestResources UP-TO-DATE
:spring-session:integrationTestClasses
:spring-session:jar UP-TO-DATE
:spring-session:integrationTest

org.springframework.session.data.gemfire.config.annotation.web.http.GemFireHttpSessionJavaConfigurationTests > gemfireCacheConfigurationIsValid FAILED
    org.junit.ComparisonFailure at GemFireHttpSessionJavaConfigurationTests.java:64

org.springframework.session.data.gemfire.config.annotation.web.http.GemFireHttpSessionJavaConfigurationTests > verifyGemFireExampleCacheRegionSessionAttributesIndexWasNotCreated FAILED
    org.junit.ComparisonFailure at GemFireHttpSessionJavaConfigurationTests.java:64

[...]

PR #402, opened 2016-03-01
Fixes gh-396

jxblum:
Applies @DirtiesContext to all Spring Session Data GemFire integration tests to ensure that no lingering GemFire (peer or client) cache instances cause conflicts on downstream tests.


rwinch:
Thanks for the PR! this is merged in via aeb0add

PR #404, opened 2016-03-01
Fixes Gh 403

jxblum:
Upgrades Spring Session to use Spring Data GemFire 1.7.4.RELEASE and addresses all TODOs.

Depends on the fix for gh-396 (PR #402); merge after PR #402


rwinch:
Thanks for the PR! This is merged into master via 3300b58af

PR #409, opened 2016-03-03
Extends classes

oassuncao:
- Extract SessionRepositoryRequestWrapper and SessionRepositoryResponseWrapper to a public class
- Created methods to override wrapRequest and wrapResponse
- Changed private fields to protected to override configuration
- Changed RedisSession to override RedisOperationsSessionRepository


rwinch:
I now understand this is on the 1.1.x branch. We are only accepting patches on 1.1.x, so I am going to close this

PR #414, opened 2016-03-06
Fix bad reference in IndexDocTests

vpavic:
Noticed this one while working on #413.

It seems ```repositoryDemo``` demo test was intended to work with ```RepositoryDemo``` instead of ```ExpiringRepositoryDemo``` (which has its own test). Additionally, needless ```@SuppressWarnings("unused")``` was removed from ```ExpiringRepositoryDemo```.

I've signed the CLA.


rwinch:

PR #333, opened 2016-01-09
implemented crawler manager similar to tomcat crawler valve

Thilak-T:
As documented in the issue https://github.com/spring-projects/spring-session/issues/147, we were using tomcat's crawler valve. So implemented for spring-session based on tomcat's valve implementation.

PR #421, opened 2016-03-11
Fix JdbcOperationsSessionRepository session cleanup

erindru:
Right now, when JdbcOperationsSessionRepository cleans up sessions, it deletes all sessions with a LAST_ACCESSED_TIME < the current time.

In practice, this is pretty much every session, and the session inactive interval is not respected. This results in every session getting deleted every minute when the `@Scheduled` task is run.

I have modified the `cleanUpExpiredSessions` function to *only* clean up sessions that have been inactive for longer than the threshold time, and added an integration test to verify this.

### DIFF
--- a/spring-session/src/integration-test/java/org/springframework/session/jdbc/JdbcOperationsSessionRepositoryITests.java
+++ b/spring-session/src/integration-test/java/org/springframework/session/jdbc/JdbcOperationsSessionRepositoryITests.java
@@ -473,6 +473,48 @@ public class JdbcOperationsSessionRepositoryITests {
 		assertThat(findByPrincipalName.keySet()).containsOnly(toSave.getId());
 	}
 
+	@Test
+	public void cleanupCleansUpInactiveSessions() {
+		this.repository.setDefaultMaxInactiveInterval(1800); // 30 minutes
+
+		JdbcOperationsSessionRepository.JdbcSession session = this.repository
+				.createSession();
+
+		this.repository.save(session);
+
+		assertThat(this.repository.getSession(session.getId())).isNotNull();
+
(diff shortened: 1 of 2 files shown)


vpavic:
@erindru thanks for testing the JDBC repository and pointing this out!
Not sure how I managed to disregard the inactivity interval during initial implementation.

rwinch:
Thanks for the PR @erindru! This is now merged into master.

PS: I like to recognize contributors for their work. Do you prefer I link to your GitHub profile, Twitter, Linked In, etc?

rwinch:
@vpavic Thanks for reviewing this!

erindru:
@rwinch happy to help, its good to see an official JDBC implementation finally. We were using our own implementation until I noticed this.

You can link to my github profile if you want to credit me, although this is the tiniest of bugfixes

rwinch:
@erindru Every fix is important! Thanks again :)

I take it the provided approach is going to work for you then?

erindru:
Yep its perfect, thanks! The only problem I have  (unrelated to this) is that the session filter creates a session after each request and I can't see how to disable this for certain URLs (we have monitoring tools that poll a status page every 10 seconds and at a session per hit this adds up to a lot of database rows if the session timeout is 30 minutes) but that's a separate issue obviously

rwinch:
@erindru Thanks for the response! With 1.1 code you can enable debugging for `org.springframework.session` and it will provide a stacktrace as to you where a session is created. It is likely something in your code that creates the session (i.e. a JSP).

rwinch:
@erindru This might be of interest to you. I need to take a look and figure out how to polish this up soon.

https://github.com/spring-projects/spring-session/pull/333

Issue #423, opened 2016-03-12
Null-result from request.getSession(false) due to invalid session id should be cached for the current request


oyvindhorneland:
If the client presents a session id that can't be found by the SessionRepository, the result is that every call to request.getSession(false) will hit the repository.getSession(id). 

Perhaps the result could be stored in a request attribute used by SessionRepositoryFilter to avoid asking the repository multiple times during the same request. It could also invalidate the cookie, since it is no longer valid.

This situation will probably only occur if an admin has cleared the session from the repository or the client for some reason presents a cookie with an old/invalid identifier, so probably rather low severity on this one. I encountered this issue while trying out the new JDBC repository support (yay!) from the latest snapshot and observed that a single request generated multiple getSession queries to the repository - one for every request.getSession(false).

oyvindhorneland:
The issue will most likely also occur when the client sends a request after the session has expired and is removed from the repository.

rwinch:
Thanks for the report! This is a nice improvement.

Any chance you would be interested in sending a PR?

oyvindhorneland:
Sure. Will do.

rwinch:
Awesome! I look forward to the PR :)

oyvindhorneland:
Added PR. 

Not invalidating the client's session cookie, yet. Do you see any scenario where invalidating the cookie would not be the correct thing to do? If we don't then the next request will trigger a new attempt to find the session by the same id.

rwinch:
@oyvindhorneland Thanks for submitting a PR! 

We don't want to invalidate the cookie as this can cause issues when working with multiple cookies of the same name but with different subdomains or multiple paths. For example, if we are in example.com we might receive a session cookie for app1.example.com which would be valid for app1, but not the parent domain. We would not want to invalidate that cookie accidentally.

oyvindhorneland:
Regarding the ICA: Can't find Spring Session at https://support.springsource.com/spring_committer_signup. Just register it on Spring Security project (If you need the ICA for this, that is)?

rwinch:
@oyvindhorneland Yes please use Spring Security. Also...if you have signed it for any Spring Project you do not need to sign again.

Please let me know once you have signed the agreement and I can merge

oyvindhorneland:
ICA signed.

rwinch:
This is now merged into master! Thanks again for the PR :)

oyvindhorneland:
Thanks! :-)

PR #426, opened 2016-03-14
Caching session not found result for the duration of the request

oyvindhorneland:
For issue #423: If a session cannot be found by id, we will cache that result for any subsequent calls to getSession for the duration of this request.


rwinch:
Thanks for the PR! This is fixed via 831d2f4152a532c1f68bdcbc08352d86efbe8e40 I applied a little polish to fix the checkstyle in 195228ff5ae52a995d4fe121cac879e4eb033d0f

PR #433, opened 2016-03-16
JdkMongoSessionConverter supports custom ClassLoader

### LINKED ISSUE #431: JdkMongoSessionConverter should allow custom ClassLoader
This will allow swapping out the `ClassLoader` used for deserialization. This will ensure that Spring Session Mongo support works with Spring Boot's devtools (which uses a custom `ClassLoader`).

I haven't looked into this, but this might imply we can combine `JacksonMongoSessionConverter` and `JdkMongoSessionConverter` by simply using custom `Serializer<Object>` and `Deserializer<Object>`. Alternatively, we probably want to use the `Converter` APIs which would allow indirect use of `Serializer<Object>` and `Deserializer<Object>`

Relates to: spring-projects/spring-boot#3805

cc @jkubrynski

jkubrynski:


rwinch:
Thanks for the PR! This was merged via 9014ac906026094131f42c61e1dc40f7f5ac515f I also applied a bit of polish for your review b58ea03a3b712f764eac1371fe7496a42cf91715

I went ahead with the release so that we could stay on schedule. If you have feedback for changes, we can include that in the next release.

Issue #406, opened 2016-03-02
Provide Redis Sample with GenericJackson2JsonRedisSerializer


rwinch:
Lots of people have been looking for how to persist with JSON. It would be good to have a sample that uses Spring Security, Spring Session Redis, and `GenericJackson2JsonRedisSerializer`.

For anyone that might be interested in this, please say comment on the issue.

Related #387

ezraroi:
Would be very useful.

I am trying to do it and I am getting the following exception on the serivce:
```

org.springframework.expression.spel.SpelEvaluationException: EL1008E:(pos 0): Property or field 'authentication' cannot be found on object of type 'java.util.LinkedHashMap' - maybe not public?
	at org.springframework.expression.spel.ast.PropertyOrFieldReference.readProperty(PropertyOrFieldReference.java:224) ~[spring-expression-4.2.5.RELEASE.jar:4.2.5.RELEASE]
	at org.springframework.expression.spel.ast.PropertyOrFieldReference.getValueInternal(PropertyOrFieldReference.java:94) ~[spring-expression-4.2.5.RELEASE.jar:4.2.5.RELEASE]
	at org.springframework.expression.spel.ast.PropertyOrFieldReference.getValueInternal(PropertyOrFieldReference.java:81) ~[spring-expression-4.2.5.RELEASE.jar:4.2.5.RELEASE]
	at org.springframework.expression.spel.ast.CompoundExpression.getValueRef(CompoundExpression.java:51) ~[spring-expression-4.2.5.RELEASE.jar:4.2.5.RELEASE]
	at org.springframework.expression.spel.ast.CompoundExpression.getValueInternal(CompoundExpression.java:87) ~[spring-expression-4.2.5.RELEASE.jar:4.2.5.RELEASE]
	at org.springframework.expression.spel.ast.SpelNodeImpl.getTypedValue(SpelNodeImpl.java:131) ~[spring-expression-4.2.5.RELEASE.jar:4.2.5.RELEASE]
	at org.springframework.expression.spel.standard.SpelExpression.getValue(SpelExpression.java:216) ~[spring-expression-4.2.5.RELEASE.jar:4.2.5.RELEASE]
	at org.springframework.session.data.redis.RedisOperationsSessionRepository$PrincipalNameResolver.resolvePrincipal(RedisOperationsSessionRepository.java:782) ~[spring-session-1.1.1.RELEASE.jar:na]
[...]

PR #434, opened 2016-03-17
Redis Sample with GenericJackson2JsonRedisSerializer

jeetmp3:
@rwinch here is sample web-app which includes Spring Security, Spring session with GenericJackson2JsonRedisSerializer (#406). Kindly review the code and let me know if any changes are required.


wgorder:
@jeetmp3  Nice I have not had a chance to dig into this too much yet but I wonder how much of the default auto-configure this negates?

https://github.com/spring-projects/spring-boot/blob/b260c20df0105842218ce9788130634f9b1f5dfa/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jackson/JacksonAutoConfiguration.java

I wonder if it could be packaged up to auto-configures after the default configuration and enhance it if a conditional property is set or some such thing.

I am sure it was probably not your intention to worry about this for a sample, but I am thinking about making it reusable as well.

Thanks,

wgorder:
Hmm nvm I see there is no  `@Bean` on your ObjectMapper and its probably better to use a separate one anyway considering the need for default typing etc.  I will have to try it out.  Someone on my team just asked me for this so maybe it will save me some time, thanks!

rwinch:
@jeetmp3 Wow this is awesome! I need to spend a little more time reviewing this on Monday. Thank you for submitting this.

We may want to get some of the Jackson Support moved to Spring Security as this would be beneficial in many other contexts. I'll review on Monday and we can go from there. Thanks for this!

rwinch:
@jeetmp3 PS If contributing the Jackson Support is something that interests you, please comment on https://github.com/spring-projects/spring-security/issues/3736 If we can get some tests added for the support, that will be a good start getting this into Spring Security proper.

Thanks again!

jeetmp3:
@wgorder this sample doesn't use Jackson default auto-configure. I haven't focused on JacksonAutoConfiguration to configure ObjectMapper. 
I think we can use Jackson2ObjectMapperBuilder to build ObjectMapper and register all the Mix-ins. By doing this we can take advantage of default auto-configure.

jeetmp3:
@rwinch sounds good to me I'll comment on that. Just wanted to confirm about test cases, you want tests for mix-in classes right?

rwinch:
@jeetmp3 Thanks for the fast reply. Yes I would like to ensure we have some tests for the mix-in classes. At that point we can add the mixins to Spring Security.

rwinch:
@jeetmp3 Also...in terms of JacksonAutoConfiguration this is something we can do later. If we have the support there at least it can be done even if it is a little more work.

We will be releasing Spring Security 4.1 RC1 on Wednesday, so it would be ideal for us to get it in before  that. However, that is probably not realistic so we may need to wait until another release.

jeetmp3:
@rwinch  I've added tests for mix-in classes.

jeetmp3:
ObjectMapper configuration changed. Now all mixins will be used from pull request  [3812](https://github.com/spring-projects/spring-security/pull/3812)

Issue #246, opened 2015-07-24
Create Grails 3 Sample and Guide


rwinch:
Related to #245

https://github.com/rwinch/spring-session/tree/gh-245

rwinch:
@bobbywarner or @erichelgeson Given your involvement in the original issue... Any chance you would be interested in creating a minimal Grails 3.1.1 / Spring Session sample, tests, and a guide? It would be similar to [Spring Session - HttpSession (Quick Start)](http://docs.spring.io/spring-session/docs/1.0.2.RELEASE/reference/html5/guides/httpsession.html) but a grails version. I'm thinking if one of you contributed the sample it would be a lot more "groovy" than if I did it. Thanks for your consideration!

rwinch:
@bobbywarner or @erichelgeson - Even if you were only able to contribute the Grails 3 / Spring Session sample code that would be extremely valuable. I could provide all the additional pieces.

bobbywarner:
@rwinch Sure, we can definitely create a sample.  Not sure if we'll get to it this week or not, but we will soon! (cc @erichelgeson)

rwinch:
@bobbywarner That sounds great! Thank you very much :) 

As I alluded to...I'm perfectly fine with this issue being broken up into multiple issues (i.e. first create a sample, then test, then a guide, etc) if that can help make progress

cc @erichelgeson

erichelgeson:
@rwinch Finally found a few min to create a minimal grails 3 app. With Grails 3.1.4, Spring-session 1.1.1 and Spring Security Core plugin 3.0.4 it all is very straight forward now https://github.com/erichelgeson/grails3-redis-session (see the 3 individual commits)

The guidance is now pretty much "add it to your build.gradle and it works!" which is great! - Do we still want a quickstart guide? (we can it's just it'll be pretty light)

I'll open a PR for this code to `samples/` shortly.

rwinch:
@erichelgeson Thanks for the response! Yes we would still like to add this even though it is very simple (it is just as simple with Boot and we demo that too). Part of the instructions will be demonstrating updating the dependencies / version that should be used.

PR #442, opened 2016-03-25
Add Grails 3 Sample

erichelgeson:
First 1/2 of #246

Shows example of session integration along with spring security core plugin integration.


rwinch:
@erichelgeson Thanks for submitting this!

I'll wait for the second half before merging/reviewing in detail. One thing we will want to ensure we have is some tests to ensure this is working.

erichelgeson:
No problem, took a first stab at the docs too (could use a second set of eyes on that as well).

erichelgeson:
@rwinch tests adapted from the spring boot example. Can be run by `./gradlew integrationTest`

rwinch:
@erichelgeson Thanks for the updates!

Some feedback.

* We should add this project to settings.gradle

```
include 'samples:grails3'
```

* The build is not currently passing. I get the following error:

```
:integrationTest

sample.HomeSpec > Anonymous page not redirected to login FAILED
    java.lang.IllegalStateException
        Caused by: org.springframework.boot.context.embedded.EmbeddedServletContainerException
            Caused by: java.lang.IllegalStateException
```

* Perhaps we can change test/index to be the index at the context root so users don't have an additional step.

erichelgeson:
The tests are passing, just I didn't have a random port assigned (the boot example didnt work in grails, but i found one that does). the root `/` is not protected, then `/test/` is a protected resource in the tests. I think it's valuable to leave both tests in (anon ok and anon redirect)

Pushed up the changes, let me know!

rwinch:
Thanks for the PR! This is now merged into master. I have applied a bit of polish via 3825a46418358a9b807f5ef9a8d8d2ba095fadd1

Issue #443, opened 2016-03-25
Documentation for web.xml filter-mapping is incorrect


sc-moonlight:
The documented web.xml's don't include the dispatcher type setting that is included in the Java configuration via AbstractHttpSessionApplicationInitializer.

This caused me some pain.  I have a doc fix, and am creating this issue so I can create a PR for it.

PR #444, opened 2016-03-25
gh-443 add Dispatcher types to filter-mapping to match the existing J…

sc-moonlight:
add Dispatcher types to filter-mapping to match the existing Java configuration


rwinch:
Thanks for the PR @sc-moonlight! This is now merged into master via bd2d84691794b8c23f081cec998289b67c99b037

Issue #445, opened 2016-03-25
JdbcOperationsSessionRepository ignores lastAccessTime


kherrala:
Session always expires after maxInactiveInterval since creation, because lastAccessTime is not loaded from database column.

Session inactivity is checked using de-serialized object, which is **not** updated using UPDATE_SESSION_LAST_ACCESS_TIME_QUERY executed when there are no changes to session attributes.

Session in deleted by the check session.isExpired() in JdbcOperationsSessionRepository.getSession(String id)

rwinch:
@vpavic Can I get your eyes on this?

vpavic:
@rwinch I'll get a look and try to reproduce later today.

vpavic:
Hopefully I'll get to fix this on Monday.

vpavic:
@rwinch I've created #450 to fix this.

@kherrala thanks for testing the ```JdbcOperationsSessionRepository``` and spotting this issue!

PR #450, opened 2016-03-28
Fix loading of JdbcSession's lastAccessedTime attribute

vpavic:
Fixes gh-445.

### DIFF
--- a/spring-session/src/main/java/org/springframework/session/jdbc/JdbcOperationsSessionRepository.java
+++ b/spring-session/src/main/java/org/springframework/session/jdbc/JdbcOperationsSessionRepository.java
@@ -102,7 +102,7 @@ public class JdbcOperationsSessionRepository implements
 
 	private static final String CREATE_SESSION_QUERY = "INSERT INTO %TABLE_NAME%(SESSION_ID, LAST_ACCESS_TIME, PRINCIPAL_NAME, SESSION_BYTES) VALUES (?, ?, ?, ?)";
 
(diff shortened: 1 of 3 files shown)


spring-session/src/main/java/org/springframework/session/jdbc/JdbcOperationsSessionRepository.java
```diff
@@ -355,6 +360,8 @@ private static GenericConversionService createDefaultConversionService() {
 
 		private final ExpiringSession delegate;
 
+		private long lastAccessedTime;
```
rwinch: Is there a reason that we cannot just set and get the lastAccessedTime on the delegate?

rwinch:
Thanks for the PR @vpavic! I had one question inline

spring-session/src/main/java/org/springframework/session/jdbc/JdbcOperationsSessionRepository.java
vpavic: Ahh yes, we can. I'll change the ```RowMapper``` to update the deserialized delegate with ```lastAccessedTime``` from column. It will also minimize the impact in other places.

rwinch: Thanks!

vpavic:
@rwinch I've updated the PR.

rwinch:
@vpavic Thanks for the PR! This is now merged into master

PR #451, opened 2016-03-28
Polish JdbcHttpSessionConfigurationTests

vpavic:
This was probably missed in 8f8cfe5.


rwinch:
@vpavic Thanks for catching this! This is now merged into master

Issue #459, opened 2016-04-01
Spring Session JDBC does not seem to commit transaction after session creation


flomorgen:
I have a Spring Boot application using an Oracle database and I'm trying to integrate Spring Session JDBC to enable several application instances to share user sessions. For user authentication Spring Security and for database access Spring Data JPA is used. The database connection pool configuration is done via _spring.datasource.*_ properties in the application.properties.
The application is configured with having _spring.datasource.default-auto-commit=false_, declaring all transactions via the @_Transactional_ annotation.

After logging in the client receives a session cookie as expected, but the database session table seems to be still empty, although the JdbcTemplate logs execution of an insert statement. Further requests using the session cookie fail because the session cannot be found in the database.(the database isolation level is READ_COMMITTED) When shutting the application down the session becomes visible in the database table. (seems like transactions are committed on shutdown)
Having _spring.datasource.default-auto-commit=true_ the session handling works fine, but this is not an option for the application.
I took a look into the Spring Session JDBC code and to me it looks like the JdbcTemplate actually does not work in context of transactions yet. Is this correct? If yes, is it possible to integrate transaction handling to Spring Session JDBC?
I could reproduce the behaviour with a slightly adapted variant of the Spring Boot example too.

vpavic:
@flomorgen thanks for testing the JDBC support and reporting this issue.

Take a look at #446, it is a different problem but the root cause is the same and its resolution is in progress.

PR #464, opened 2016-04-03
Enable transaction management for JdbcOperationsSessionRepository operations

### LINKED ISSUE #446: JdbcSession update fails during read-only transaction
Flushing or closing HttpServletResponse outputStream while still inside readOnly transaction fails because JdbcOperationsRepository tries to execute SQL update query.  The expection is as follows:

`org.springframework.jdbc.UncategorizedSQLException: PreparedStatementCallback; uncategorized SQLException for SQL [UPDATE SPRING_SESSION SET LAST_ACCESS_TIME = ? WHERE SESSION_ID = ?]; SQL state [25006]; error code [0]; ERROR: cannot execute UPDATE in a read-only transaction; nested exception is org.postgresql.util.PSQLException: ERROR: cannot execute UPDATE in a read-only transaction`

Behaviour can be replicated by calling following fragment inside spring annotated transaction:
`try (InputStream is = connection.getInputStream()) { IOUtils.copy(is, response.getOutputStream()); }`

Default configuration should avoid sharing user defined transaction in JdbcTemplate or defer updating database session until filter chain has been processed in SessionRepositoryFilter.

vpavic:
This PR adds option to enable transactional execution of write operations in ```JdbcOperationsSessionRepository```.

Transactional execution is enabled by using ```@EnableJdbcHttpSession(transactional = true)``` and providing ```PlatformTransactionManager``` or , in case of manual configuration, by setting ```PlatformTransactionManager``` directly on ```JdbcOperationsSessionRepository``` using appropriate setter method. Transactions will be executed using propagation ```REQUIRES_NEW```.

This change fixes #446 and #459.

@rwinch please review. If you're OK with this approach I'll proceed to update the PR documentation updates.

I've signed the CLA.

### DIFF
--- a/docs/src/docs/asciidoc/guides/httpsession-jdbc-xml.adoc
+++ b/docs/src/docs/asciidoc/guides/httpsession-jdbc-xml.adoc
@@ -85,6 +85,7 @@ The filter is what is in charge of replacing the `HttpSession` implementation to
 In this instance Spring Session is backed by a relational database.
 <2> We create a `dataSource` that connects Spring Session to an embedded instance of H2 database.
 We configure the H2 database to create database tables using the SQL script which is included in Spring Session.
+<3> We create a `transactionManager` that manages transactions for previously configured `dataSource`.
 
 == XML Servlet Container Initialization
 
--- a/docs/src/docs/asciidoc/guides/httpsession-jdbc.adoc
+++ b/docs/src/docs/asciidoc/guides/httpsession-jdbc.adoc
@@ -83,6 +83,7 @@ The filter is what is in charge of replacing the `HttpSession` implementation to
 In this instance Spring Session is backed by a relational database.
 <2> We create a `dataSource` that connects Spring Session to an embedded instance of H2 database.
 We configure the H2 database to create database tables using the SQL script which is included in Spring Session.
+<3> We create a `transactionManager` that manages transactions for previously configured `dataSource`.
 
 == Java Servlet Container Initialization
 
--- a/docs/src/docs/asciidoc/index.adoc
+++ b/docs/src/docs/asciidoc/index.adoc
@@ -1012,7 +1012,7 @@ A typical example of how to create a new instance can be seen below:
 include::{indexdoc-tests}[tags=new-jdbcoperationssessionrepository]
 ----
 
-For additional information on how to create and configure a `JdbcTemplate`, refer to the Spring Framework Reference Documentation.
+For additional information on how to create and configure `JdbcTemplate` and `PlatformTransactionManager`, refer to the Spring Framework Reference Documentation.
 
 [[api-jdbcoperationssessionrepository-config]]
 ==== EnableJdbcHttpSession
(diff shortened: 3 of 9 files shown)


rwinch:
@vpavic Thanks for the PR! 

I think I'd prefer to allow the user to create the transactional boundary by exposing a bean of a particular name. 
* This aligns with Spring Session's other points of customization. For example, this is how we users can provide custom serializers. 
* Additionally, using a bean provides more flexibility for the user than the boolean flag on the annotation (which can only be on or off).

What are your thoughts?

vpavic:
@rwinch I see your point here - you'd prefer to take a less opinionated route on how to approach the transaction management within the ```JdbcOperationsSessionRepository```, and instead of focusing on providing concrete solutions for detected problems with transactions we'd equip the users with flexiblity (by setting their own ```TransactionOperations``` implementation) so they can come up with their own prefered solution.

I guess this makes sense, and it actually makes the job quite easier for us in terms of implementation as well as testing. Documentation should provide some concrete examples for this approach.

Do you prefer to keep offer this only for write operations, or read ones as well?

rwinch:
@vpavic Thanks for the fast reply. I'd say we should do transactions or reads also.

vpavic:
@rwinch I've updated the PR.

Please share your thoughts. Documentation is still pending.

kherrala:
I would still prefer using REQUIRES_NEW propagation behaviour as a sane default, because I would personally expect this to "just work" without extra configuration. My previous suggestion was to avoid introducing unnecessary configuration for the time being, but the behaviour on rollback is not quite the same with session persist after succesful commit.

vpavic:
@kherrala FWIW I agree with you on ```REQUIRES_NEW``` being the sane default.

However @rwinch has so far insisted on opt-in approach for transactional behavior, with argument that transactions as an expensive operation might not be needed/wanted by all users.

rwinch:
@kherrala @vpavic Thanks for your feedback.

I've been doing some more thinking about this and I think you two have convinced me that my approach was not ideal. We should work out of the box, so I think you two are right that we should create the transactions by default.

What's more is we can always make it optional later on and still remain passive. If we start without requiring it and it causes problems, we are stuck with this default to remain passive.

@vpavic I'm sorry on flip-flopping on this, but hopefully getting to the "right answer" is some consultation. Do you mind updating the PR to take these changes into account.

vpavic:
> I'm sorry on flip-flopping on this, but hopefully getting to the "right answer" is some consultation.

@rwinch I'm always in favor of thinking things through a bit more event at the expense of delaying things so :+1: on that from my side.

No problems on updating the PR, I'll take care of that soon.

Now that we've decided what the default behavior is, the question is what means do we provide for changing the defaults?

> What's more is we can always make it optional later on and still remain passive. If we start without requiring it and it causes problems, we are stuck with this default to remain passive.

If I'm reading this correctly you wouldn't provide an option for users to disable transactions right now, but rather at some later stage if there's need for this?

rwinch:
> so :+1: on that from my side.

Glad to hear that :)

> No problems on updating the PR, I'll take care of that soon.

Awesome! Thanks :)

> If I'm reading this correctly you wouldn't provide an option for users to disable transactions right now, but rather at some later stage if there's need for this?

Correct

vpavic:
@rwinch I've updated the PR.

rwinch:

PR #468, opened 2016-04-06
Couchbase support - with namespaces

mkopylec:


.travis.yml
```diff
@@ -3,6 +3,15 @@ language: java
 services:
   - redis-server
 
+before_install:
```
rwinch: Can we just use services as shown https://docs.travis-ci.com/user/database-setup/#CouchDB

docs/src/docs/asciidoc/guides/couchbase.adoc
```diff
@@ -0,0 +1,133 @@
...
+:toc:
+
+This guide describes how to use Spring Session backed by Couchbase.
+
+== Updating Dependencies
+Before you use Spring Session, you must ensure to update your dependencies.
+We assume you are working with a working Spring Boot web application.
+If you are using Gradle, ensure to add the following dependencies:
+
+.gradle.build
```
rwinch: Can you demonstrate with Maven like the other examples?

docs/src/docs/asciidoc/guides/couchbase.adoc
```diff
@@ -0,0 +1,133 @@
...
+public class SessionController {
+
+    @RequestMapping("uri")
+    public void doSomething(HttpSession session) {
+        ...
+    }
+}
+----
+
+**Namespaces support:**
```
rwinch: Is there a reason for having a global namespace? In general, I'm not sure I like the additional complexity of this (not to mention none of the other implementations support this).

spring-session/src/main/java/org/springframework/session/data/couchbase/CouchbaseSessionRepository.java
```diff
@@ -0,0 +1,206 @@
...
+package org.springframework.session.data.couchbase;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+import com.couchbase.client.java.document.json.JsonObject;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import org.slf4j.Logger;
```
rwinch: We use commons-logging (like Spring Framework)

spring-session/src/main/java/org/springframework/session/data/couchbase/DelegatingSessionStrategy.java
```diff
@@ -0,0 +1,72 @@
+/*
```
rwinch: I don't think this should be needed. I think this is probably due to the global support which I think we should remove. If we do add global session support, we would move it to a different PR.

spring-session/src/main/java/org/springframework/session/data/couchbase/DelegatingSessionStrategy.java
```diff
@@ -0,0 +1,72 @@
...
+ * limitations under the License.
+ */
+package org.springframework.session.data.couchbase;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+import org.springframework.session.Session;
+import org.springframework.session.web.http.CookieHttpSessionStrategy;
+import org.springframework.session.web.http.MultiHttpSessionStrategy;
```
rwinch: This package should not depend on servlet dependencies.

spring-session/src/main/java/org/springframework/session/data/couchbase/Serializer.java
```diff
@@ -0,0 +1,87 @@
+/*
```
rwinch: I don't think we should have a couchbase specific API here. We could use something like Spring's converter APIs to convert the values.

spring-session/src/main/java/org/springframework/session/web/http/OnCommittedResponseWrapper.java
```diff
@@ -529,6 +530,14 @@ public void println(double d) throws IOException {
 			this.delegate.println(d);
 		}
 
+		public boolean isReady() {
```
rwinch: These changes should be in a distinct PR

spring-session/src/main/java/org/springframework/session/web/http/SessionRepositoryFilter.java
```diff
@@ -94,6 +94,12 @@
 			+ ".invalidSessionId";
```
rwinch: These changes should be in its own PR

spring-session/src/main/java/org/springframework/session/data/couchbase/CouchbaseDao.java
```diff
@@ -0,0 +1,118 @@
...
+import org.springframework.util.Assert;
+
+/**
+ * Data access object that communicates with Couchbase database in order to manage HTTP
+ * session data persistence.
+ *
+ * @author Mariusz Kopylec
+ * @since 1.2.0
+ */
+public class CouchbaseDao {
```
rwinch: Is there a reason we need a distinct `CouchbaseDao`? Perhaps this would be better in the `CouchbaseSessionRepository`

docs/src/docs/asciidoc/guides/couchbase.adoc
mkopylec: Sure, I thought that if the whole project is build on gradle I will stick to one build technology.

.travis.yml
mkopylec: No those two are different databases: http://www.couchbase.com/couchbase-vs-couchdb

docs/src/docs/asciidoc/guides/couchbase.adoc
mkopylec: Generally this is for microservices architecture systems. There couple of reasons I decided to go with namespace:
1. Session data in a large distributed system (aspecially in microservices architecture) can be pretty big. Having app and global namespaces reduces the amount of data to read every `HttpSession.getSession()` invocation. You can gain a significant performance boost using session namespaces.
2. There is alomst none additional complexity for a developer. If you don't want to use them you can set the same namespace in all your apps in the system or use only global namespace. Maybe by default namespaces support could be turned off.
3. With namespaces multiple different microservices in teh system can set or get a session attribute of the same without conflicting each other.

spring-session/src/main/java/org/springframework/session/data/couchbase/CouchbaseSessionRepository.java
mkopylec: Ok, I'll change that

spring-session/src/main/java/org/springframework/session/data/couchbase/DelegatingSessionStrategy.java
mkopylec: Yes, this because of namespaces support.

spring-session/src/main/java/org/springframework/session/data/couchbase/Serializer.java
mkopylec: Sure, good idea.

spring-session/src/main/java/org/springframework/session/web/http/OnCommittedResponseWrapper.java
mkopylec: This is because `DelegatingSessionStrategy` depends on servlet api 3.1.0

spring-session/src/main/java/org/springframework/session/data/couchbase/CouchbaseDao.java
mkopylec: I'm not quite sure it's a good idea. From the DDD side the `CouchbaseSessionRepository` playes a domian repository role for session entity and the `CouchbaseDao` is just a data access object that communicates directly with couchbase server.
Beside that joining those two guys will create a quite big, hard to read class.

spring-session/src/main/java/org/springframework/session/web/http/OnCommittedResponseWrapper.java
rwinch: Right but you will notice that none of the other repository implementations have web dependencies in them. Doing this will cause [package tangles](http://structure101.com/blog/2008/12/software-erosion-and-package-tangles/).

I really think we need to consider global namespace as its own thing as a PR to add new repository should not contain web dependencies nor should it change anything out of its package structure.

docs/src/docs/asciidoc/guides/couchbase.adoc
rwinch: Thanks. I think eventually we should do Gradle too, but most people that know Gradle syntax know Maven. However, people that know Maven don't know Gradle.

.travis.yml
rwinch: Thanks for pointing that out. I did not read carefully.

docs/src/docs/asciidoc/guides/couchbase.adoc
rwinch: If we are going to consider global namespace, it will need to be a distinct ticket and PR. We don't want to change too much in one ticket / commit as this makes troubleshooting difficult.

mkopylec: Ok, I understand. I can create such a ticket and PR.One important question here: should the PR consider namespaces support only for data couchbase or maybe the changes should be global and made in the core Spring Session API (`ExpiringSession`, `FindByIndexNameSessionRepository`, etc.)?

rwinch: Thanks! Lets first create a ticket and discuss if this is something we want to add to Spring Session. It is quite likely a useful feature, but I'm not entirely convinced Spring Session is the right place for it.

rwinch:
Global namespace does not fit in with Spring Session. Closing in favor of #477

PR #472, opened 2016-04-08
Ehcache session repository implementation

Leaghte:


Leaghte:
I noticed one show stopper that i need to fix, then i will reopen pull request :/

Issue #65, opened 2014-11-18
Spring Security Concurrent Session Integration


rwinch:
Spring Security allows for ensuring only one user with a given username is authenticated at a time through the [concurrency control](http://docs.spring.io/spring-security/site/docs/current/reference/html/session-mgmt.html#concurrent-sessions). This is nice for something that is a subscription model (i.e. pay per user) to ensure that people are not sharing usernames.

If two sessions exist with the same username, Spring Security will invalidate one of the sessions. Currently it does this by marking a session as invalid and the next time the user visits the page the session is actually invalidated. This is necessary due to Java EE limitations. 

Spring Security should provide an implementation of concurrency control that will use Spring Session's APIs to immediately invalidate a session.

yacota:
Is there any plan to include this issue in a release?
From one of your comments at https://spring.io/blog/2014/11/18/spring-session-1-0-0-rc1-released#comment-1699673845 ... we'll have to wait until 1.1.0 at least.

rwinch:
@yacota Thanks for your interest. At this time there are no concrete plans of when this integration would be made. Of course a good PR with proper testing would almost guarantee inclusion in the next feature release :)

One other question is...does integration between Spring Session & Security belong in Spring Security or in Spring Session? This will be something that we can easily change later, but it is something I may want feedback from @wilkinsona on.

yacota:
Thanks @rwinch for your rapid answer.
I see your point about where does this integration belong to. 
Another point I am afraid of is that implementation details could vary "a lot" depending on which session repository is used (Redis, Hazelcast, ...), maybe a core API interface that deals with main concepts like session invalidation, .... with other jars that really implement it (-redis.jar, -hazelcast.jar) could work.
I think that this is a similar approach took in Spring Social design

Thanks

dnang:
Hello, I am new here. Can anyone explain what this issue about? I was confused when it links to a section in Spring Security official doc that I am reading.

rwinch:
@dnang See if the updated description answers your question.

dnang:
Excellent @rwinch, it's clear now.

rwinch:
@dnang As an FYI this is now supported in the latest snapshot (indirectly) through resolving #4 See the updated documentation http://docs.spring.io/spring-session/docs/current-SNAPSHOT/reference/html5/#httpsession-httpsessionlistener

We still plan on adding an improved implementation as described in this issue, but I wanted to communicate that you can somewhat work around the limitation now.

[...]

PR #473, opened 2016-04-09
Spring security session registry

jkuipers:
This provides support to use a Spring Session-backed SessionRepository for Spring Security, which allows concurrent session control in a clustered environment. It relates to https://github.com/spring-projects/spring-session/issues/65.

I've created this PR per Rob's request: I've tried to clean up my sample code that I wrote earlier a bit and have added some documentation on configuration and limitiations of the provided solution. I hope this helps!

I have already signed the Contributor License Agreement for Spring in the past.

### DIFF
--- a/docs/build.gradle
+++ b/docs/build.gradle
@@ -35,6 +35,7 @@ dependencies {
 				"org.springframework:spring-websocket:${springVersion}",
 				"org.springframework:spring-messaging:${springVersion}",
 				"org.springframework:spring-jdbc:${springVersion}",
+				"org.springframework.security:spring-security-config:${springSecurityVersion}",
 				"org.springframework.security:spring-security-web:${springSecurityVersion}",
 				"org.springframework.security:spring-security-test:${springSecurityVersion}",
 				'junit:junit:4.11',
--- a/docs/src/docs/asciidoc/index.adoc
+++ b/docs/src/docs/asciidoc/index.adoc
@@ -501,6 +501,53 @@ Before using WebSocket integration, you should be sure that you have <<httpsessi
 
 include::guides/websocket.adoc[tags=config,leveloffset=+2]
 
+[[spring-security-concurrent-sessions]]
+== Spring Security Integration
+
+Spring Session provides integration with Spring Security to support its concurrent session control.
+This allows limiting the number of active sessions that a single user can have concurrently, but unlike the default
+Spring Security support this will also work in a clustered environment. This is done by providing a custom
+implementation of Spring Security's `SessionRegistry` interface.
+
+[[spring-security-concurrent-sessions-how]]
+=== Configuring Spring Security's concurrent session management
+
+When using Spring Security's Java config DSL, you can configure the custom `SessionRegistry` through the
+`SessionManagementConfigurer` like this:
+[source,java,indent=0]
+----
+include::{docs-test-dir}docs/security/SecurityConfiguration.java[tags=class]
+----
+
+This assumes that you've also configured Spring Session to provide a `FindByIndexNameSessionRepository` that
+returns `ExpiringSession` instances.
+
+When using XML configuration, it would look something like this:
+[source,xml,indent=0]
+----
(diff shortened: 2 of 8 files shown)


rwinch:
@jkuipers Thanks for the PR!

I'm wondering if we would be able to support the concept of marking the session as expired by setting a session attribute rather than deleting the session. Then when Spring Session obtains the `SessionInformation` it would read this attribute to determine if the session is expired.

Honestly, I do like the fact that the current implementation deletes the session, but this might be a nice feature for users who need the ability to display a more useful error message.

Thoughts?

rwinch:
Note I'm assigning this to 1.3.0 M1 because we plan to go GA with Spring Session 1.2 next and we need have at least one RC of a new feature before going GA.

jkuipers:
A custom session attribute to support "Spring Security expiration" makes a lot of sense: we should even be able to support true deletion after marking a session as expired because the SessionRegistry is already called to perform that cleanup. I'll update the implementation and docs to ensure that this in place. This will also cause the filtering on expired session to actually do something useful, since it won't conflate Spring Security's notion of an expired session with that of Spring Session: they're different things, 
I could also support repositories that return plain Sessions rather than ExpiringSessions this way: the SessionRegistry already gets a callback to update the last accessed timestamp of a session, so for plain Session (that don't provide this information themselves like ExpiringSessions do) we could store the info in another custom session attribute. The code could determine and cache whether plain or expiring sessions are being returned by the repository, so the additional overhead of updating the Session with a last accessed attribute would only be incurred for plain Sessions. Do you think that this is worth the trouble, or will all relevant implementations implement ExpiringSession anyway, in your opinion?

rwinch:
Thanks for the feedback. I think all relevant implementations will implement ExpiringSession. Honestly, the split between the two APIs was probably a mistake. However, something that cannot really be changed now.

jkuipers:
OK, I've pushed an updated implementation that tracks expiry in the Spring Security sense through a custom Session attribute. When Spring Security's `ConcurrentSessionFilter` finds an expired SessionInformation it will call all registered `LogoutHandler` beans, which typically would lead to the Session being deleted. Even if this wouldn't happen, then the Session will simply expire (in the Spring Session sense) eventually.

edethoor:
@rwinch have an idea of the release date of 1.3.0 ?
I have to develop this feature but if it is supported out of the box in the next release, I prefer to wait.

Thanks.

jkuipers:
Note that you can take the code from this pull request and use it as-is with the current version of spring-session: that should provide you with a smooth upgrade path if the request is indeed merged for 1.3.

rwinch:
@edethoor Unfortunately we don't have an ETA on 1.3.0 at this time. As @jkubrynski suggests, you should be able to take the code from this PR as is and then remove it when 1.3.0 comes around

rwinch:
Thanks for the PR @jkuipers This is now merged into master

codecracker2014:
@jkuipers i'm using infinispan for cache in cluster which provides Map based cache implementation. i need to use SpringSessionBackedSessionRegistry with MapSessionRepository but MapSessionRepository doesn't implement FindByIndexNameSessionRepository. How can i achieve this.

jkuipers:
@codecracker2014 It wouldn't be hard to add support for the `FindByIndexNameSessionRepository` to the `MapSessionRepository`: a naive implementation could simply loop through the stored `ExpiringSession`s and find the ones with the given principal name, a more optimized impl could simply have a map from principal names to a collection of corresponding sessions to make lookups efficient. Not sure what would be best for Infinispan, since I'm not familiar with that.

AmitMathur:
@jkuipers 
this is great... thanks for the blog it was helpful.
I have a situation, where I want to show user a custom page if he reaches the max limit for concurrent session. On that page, when the user clicks on Continue, that's when we invalidate the oldest one and create a new one. 
I wanted to see if that is something supported now?

Appreciate your help!!!

Issue #407, opened 2016-03-02
Actuator Endpoint to Delete Sessions


rwinch:
It would be nice to add an actuator [Endpoint](https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-endpoints.html#production-ready-customizing-endpoints-programmatically) that:
- Allows finding sessions by username
- Displays the resulting sessions
- Then allows deleting the session

I think an HTML version of this would be awesome for administrators, but we would likely want to support all the various Actuator endpoint styles (i.e. JMX, JSON, etc)

Related #408

vpavic:
This is scheduled to be implemented in Boot 2.0 via spring-projects/spring-boot#8342.

PR #480, opened 2016-04-12
Add actuator endpoint to find and delete sessions

eddumelendez:
- [X] I have signed the CLA

See gh-407


spring-session-boot/src/main/java/org/springframework/session/endpoint/SessionEndpoint.java
```diff
@@ -0,0 +1,65 @@
...
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+
+/**
+ * {@link MvcEndpoint} to expose actuator session.
+ *
+ * @author Eddú Meléndez
+ * @since 1.2.0
+ */
```
tsachev: Why not use `@ConfigurationProperties` here and make `sensible` and `path` configurable?

spring-session-boot/src/main/java/org/springframework/session/endpoint/EndpointConfiguration.java
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * {@link EnableAutoConfiguration Auto-configuration} for session management
+ * {@link Endpoint}s.
+ *
+ * @author Eddú Meléndez
+ * @since 1.2.0
+ */
+@Configuration
+@ConditionalOnBean(FindByIndexNameSessionRepository.class)
```
tsachev: This cloud be spring boot autoconfiguration registered in `spring.factories`

vpavic:
@eddumelendez This is a very useful addition however IMO it should be targeted at Spring Boot since first-class Spring Session integration already exists there.

WDYT @rwinch?

rwinch:
@vpavic @eddumelendez I do think it may be better fit for Spring Boot as this is where most actuator endpoints live.

vpavic:
@rwinch Also I don't think it's a wise idea to have a dependency on Boot (other than in sample apps that is).

Since Boot's POM actually provides the dependency management for Session that would effectively create a circular dependency between the projects.

cemo:
+1 @vpavic

eddumelendez:
Thanks guys for the feedback I will move my changes to spring-boot project... Hope to see all new improvements there soon.

eddumelendez:
Close in favor of https://github.com/spring-projects/spring-boot/pull/5730

Issue #488, opened 2016-04-18
JdbcHttpSessionConfiguration should expose setters for annotation attributes


snicoll:
I am trying to integration spring session based on a PR. I am moving away from the current proposal to extend from each configuration classes (which allows me to customize the store specific settings exposed by each individual annotation).

`HazelcastHttpSessionConfiguration` has a `setMaxInactiveIntervalInSeconds` but `JdbcHttpSessionConfiguration` doesn't. There is no setter to customize the `tableName`

snicoll:
`MongoHttpSessionConfiguration` is also affected. I can submit a PR if needed.

rwinch:
This is fixed via #490

PR #490, opened 2016-04-18
Add tableName setter for JdbcHttpSessionConfiguration

eddumelendez:
- [X] I have signed the CLA

See gh-488

### DIFF
--- a/spring-session/src/test/java/org/springframework/session/jdbc/config/annotation/web/http/JdbcHttpSessionConfigurationTests.java
+++ b/spring-session/src/test/java/org/springframework/session/jdbc/config/annotation/web/http/JdbcHttpSessionConfigurationTests.java
@@ -40,6 +40,7 @@ import static org.mockito.Mockito.mock;
  * Tests for {@link JdbcHttpSessionConfiguration}.
  *
  * @author Vedran Pavic
+ * @author Eddú Meléndez
  * @since 1.2.0
  */
 public class JdbcHttpSessionConfigurationTests {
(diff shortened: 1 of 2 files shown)


vpavic:
@eddumelendez this does not address the #488 since you've only added the setter for ```tableName```.

Also the test you added doesn't actually test anything relevant here. You're just using ```tableName``` setter directly on ```JdbcOperationsSessionRepository``` instead of the one you added in ```JdbcHttpSessionConfiguration```.

snicoll:
@eddumelendez a good way to test this would be to reuse what I've done in Spring Boot (i.e. [extending the regular configuration](https://github.com/snicoll/spring-boot/commit/f46d83938b5668de3d1eec8b0665eebab0edada1#diff-09c1399b224b0c1e813969745a9b661dR39))

eddumelendez:
@vpavic I noticed that when I was looking for Mongo implementation.

@snicoll Thanks for the hint. I will polish this PR.

spring-session/src/test/java/org/springframework/session/jdbc/config/annotation/web/http/JdbcHttpSessionConfigurationTests.java
```diff
@@ -181,6 +196,17 @@ public PlatformTransactionManager transactionManager() {
 	}
 
 	@Configuration
+	static class CustomTableNameSetConfiguration extends JdbcHttpSessionConfiguration {
+
+		@Autowired
+		public void customize() {
+			DirectFieldAccessor fieldAccessor = new DirectFieldAccessor(this);
+			fieldAccessor.setPropertyValue("tableName", "custom_session");
```
snicoll: uh? Shouldn't you just be using `setTableName`?

snicoll:
It's also missing a `setMaxInactiveIntervalInSeconds`

spring-session/src/test/java/org/springframework/session/jdbc/config/annotation/web/http/JdbcHttpSessionConfigurationTests.java
eddumelendez: my mistake. now it is fixed. thanks!

spring-session/src/test/java/org/springframework/session/jdbc/config/annotation/web/http/JdbcHttpSessionConfigurationTests.java
```diff
@@ -181,6 +207,17 @@ public PlatformTransactionManager transactionManager() {
 	}
 
 	@Configuration
+	static class CustomTableNameSetConfiguration extends JdbcHttpSessionConfiguration {
+
+		@Autowired
+		public void customize() {
```
vpavic: Why a random ```@Autowired``` method?
Simply overriding the default constructor and using the desired setter inside it is much cleaner and more understandable.

Also separate config for each test are preferred.

snicoll: I guess it's taken from what I've done in Spring Boot. I agree an `@Autowired` method with no parameter is not really a good idea.

vpavic:
@eddumelendez Looks good now, thanks for updating the PR.

eddumelendez:
@vpavic thanks to you for the feedback.

rwinch:

PR #495, opened 2016-04-20
Implement individual attribute persistence in JdbcOperationsSessionRepository

### LINKED ISSUE #486: Independent update of attributes for JDBC Implemetation
from #398 

> I'm wondering if in practice we are going to need to change the JDBC implementation to write attributes > independently so we can only write attributes that have changed. Can you create a new ticket?

vpavic:
This resolves #486.

@rwinch Please review but don't merge yet as there are still a couple of TODOs:

- [x] Check and update the javadocs
- [x] Check and update the documentation
- [x] Decide whether to provide an option on persistence mode (single BLOB vs individual attributes)

I've ran the integration test against PostgreSQL, Oracle, MySQL/MariaDB, H2, HSQL and Derby.

- [X] I have signed the CLA


vpavic:
PR updated with javadoc/documentation changes.

cemo:
@vpavic This is awesome. Do you hit any performance penalty?

vpavic:
@cemo Thanks for the reminder, I did some benchmarks.

The test uses 20 threads to create 10000 sessions, then updates them using different scenarios, reloads them from database, and finally deletes them. Sessions are created with 5 randomly generated ```String``` attributes, one static ```String``` attribute and ```SecurityContext``` attribute

I ran that on PostgreSQL. Here are the results:

Current master:
> Created 10000 sessions in 6718 millis
> Updated (add new attribute) 10000 sessions in 6050 millis
> Updated (modify existing attribute) 10000 sessions in 4684 millis
> Updated (modify last accessed time) 10000 sessions in 4735 millis
> Updated (modify security context) 10000 sessions in 7008 millis
> Loaded 10000 sessions in 2925 millis
> Deleted 10000 sessions in 3723 millis

This PR:
> Created 10000 sessions in 13794 millis
> Updated (add new attribute) 10000 sessions in 5348 millis
> Updated (modify existing attribute) 10000 sessions in 3171 millis
> Updated (modify last accessed time) 10000 sessions in 3370 millis
> Updated (modify security context) 10000 sessions in 5209 millis
> Loaded 10000 sessions in 3359 millis
> Deleted 10000 sessions in 3625 millis

cemo:
> Decide whether to provide an option on persistence mode (single BLOB vs individual attributes)

There might be one another choice for this. Individual attributes are required in scenarios where they have to be updated outside from the application such as cron jobs. In these cases actually we are just interested in some attributes not all of them. Simple session attributes which are updated only by user are not necessary to be stored in individual attributes table. I do not know what can we do for this but I just wanted to share this as well.   

Thank you so much for everything. <3

vpavic:
Thanks for the feedback @cemo!

Interesting approach on attributes persistence. It seems optimal from the perspective of number of BLOBs involved but I see two downsides - it makes the configuration more complex, and it seems to require some extra logic in session handling.

The more I look at this the more it seems to me that individual attributes is the way to go. The only downside is the performance on initial save but I've been able to optimize that a bit by using ```JdbcTemplate#batchUpdate```.

rwinch:
@vpavic Thanks for the updated PR! I have been swamped with other things and haven't had a chance to look at this in any detail. I will review first thing Monday morning.

rwinch:
Thanks for the PR! This is merged via 00e7110594e39bd81da79934719ad491275ddb80

Issue #275, opened 2015-08-21
CookieHttpSessionStrategy should look at all cookies not just the first


rwinch:
Right now if multiple cookies are submitted to Spring Session, it only considers the first cookie with the correct name. We should inspect additional cookies if the session id in the first cookie was not valid.

We need to select the first session id that is valid for this application. So for example, if a request to `/other/` is made

``` http
GET /other/ HTTP/1.1
Host: example.com

HTTP/1.1 200 OK
Set-Cookie: SESSION=df540dc9-6919-4fe2-a410-5b04cdf9917c; Path=/other/
```

Then a request to `/` is made

``` http
GET / HTTP/1.1
Host: example.com

HTTP/1.1 200 OK
Set-Cookie: SESSION=e4616883-678c-4e1b-a154-d3106fe60bf6; Path=/
```

Finally another request to `/other/` is made in which both SESSION cookies are submitted with the request:

``` http
GET /other/ HTTP/1.1
Host: example.com
Cookie: SESSION=eba15ac9-3fa6-40cb-8666-20bcc4b6bf22; SESSION=ddaa23ff-698b-4af4-9a89-285185bf1ef9
```

If these two sessions are for different applications, then we must select the correct session id. Note that the client will not include the path information in the request (it is only in the response).

We must also consider a similar scenario for the domain.

Part of the struggle with this is how to make the changes passive (or as passive as possible) without looking up the session multiple times.

Relates to gh-362

PR #505, opened 2016-04-27
Look at all cookies instead just the first

eddumelendez:
- [X] I have signed the CLA

Previous to this commit, just the first cookie is considered. Now, all
cookies are considered to look for SESSION cookie.

See gh-275


rwinch:
@eddumelendez This is a step in the right direction. However, it isn't quite what #275 had in mind. The problem is that a cookie with the same name can be presented multiple times in the request. This happens if the cookie is associated to different paths or different domains. I updated the original issue to better illustrate the problem.

eddumelendez:
@rwinch thanks for the details. I will take a look and see if can update the PR.

Issue #514, opened 2016-05-05
Session @Scope bean and destruction callback


denov:
I have a user perf bean that's session scoped.  

```
@Component
@Scope(value=WebApplicationContext.SCOPE_SESSION, proxyMode=ScopedProxyMode.INTERFACES)
public class UserPreferencesImpl implements UserPreferences {
...
}
```

I'm seeing a _org.springframework.web.context.request.ServletRequestAttributes.DESTRUCTION_CALLBACK.scopedTarget.userPreferencesImpl_ saved along with this bean and SPRING_SECURITY_CONTEXT

Is this normal, a bug or a mis-config? Can I prevent this from being persisted?  The name is too long for the default schema.

vpavic:
@denov We need to allow bigger attribute keys with default schema. As a workaround you can manually update your schema to increase the column size.

@rwinch Any thoughts on the column size we should go with?

rwinch:
It's tricky because attribute names can be arbitrary length. I think the increase you did is fine. We can always bump it up later (or users can customize it themselves).

rwinch:
@denov Yes this is expected. It has to do with how the bean is destroyed.

denov:
@rwinch are there any docs about this?  is the actually callback code being saved?

i also went to 200 has the column size as i don't expect to have a class name of 80+ chars.

rwinch:
@denov I'm honestly not sure if/where this is documented as this doesn't really have anything to do with Spring Session (it is a Spring Framework feature). This would be something within http://docs.spring.io/spring/docs/current/spring-framework-reference/htmlsingle/#beans-factory-scopes-session

The reasoning behind it is to handle lifecycle methods on the Bean at the time the session expires or is deleted.

I'm guessing the details are buried within the Javadoc somewhere.

PR #515, opened 2016-05-06
Increase default size for attribute_name column

vpavic:
This addresses the #514.
I've settled at 200 for the default value for the column length.

- [X] I have signed the CLA


rwinch:
Thanks merged!

Issue #525, opened 2016-05-20
MapSessionRepository should only save if something has changed


rwinch:
We should improve MapSessionRepository to only perform a save if something has changed. This will improve performance when multiple saves are invoked per request (i.e. once on response committed and once on the SessionRepositoryFilter completion).

vpavic:
We have decided not to add more complexity to `MapSessionRepository` as we are not promoting it as production grade `SessionRepository` implementation.

PR #528, opened 2016-05-21
Perform save operation in MapSessionRepository when something has

eddumelendez:
- [X] I have signed the CLA

changed

Nowadays, MapSessionRepository perform a save at anytime. This commit
ensure to perform a save when session is created and when something has
changed.

Fixes gh-525

### DIFF
--- a/spring-session/src/main/java/org/springframework/session/MapSession.java
+++ b/spring-session/src/main/java/org/springframework/session/MapSession.java
@@ -41,6 +41,7 @@ import java.util.concurrent.TimeUnit;
  * </p>
  *
  * @author Rob Winch
+ * @author Eddú Meléndez
  * @since 1.0
  */
 public final class MapSession implements ExpiringSession, Serializable {
@@ -53,6 +54,7 @@ public final class MapSession implements ExpiringSession, Serializable {
 	private Map<String, Object> sessionAttrs = new HashMap<String, Object>();
 	private long creationTime = System.currentTimeMillis();
 	private long lastAccessedTime = this.creationTime;
+	private boolean changed;
 
 	/**
 	 * Defaults to 30 minutes.
@@ -75,6 +77,7 @@ public final class MapSession implements ExpiringSession, Serializable {
 	 */
 	public MapSession(String id) {
 		this.id = id;
+		this.changed = true;
 	}
 
 	/**
@@ -101,6 +104,7 @@ public final class MapSession implements ExpiringSession, Serializable {
 
 	public void setLastAccessedTime(long lastAccessedTime) {
 		this.lastAccessedTime = lastAccessedTime;
+		this.changed = true;
 	}
 
 	public long getCreationTime() {
@@ -117,6 +121,7 @@ public final class MapSession implements ExpiringSession, Serializable {
 
 	public void setMaxInactiveIntervalInSeconds(int interval) {
 		this.maxInactiveInterval = interval;
+		this.changed = true;
 	}
 
 	public int getMaxInactiveIntervalInSeconds() {
@@ -150,11 +155,17 @@ public final class MapSession implements ExpiringSession, Serializable {
 		}
 		else {
 			this.sessionAttrs.put(attributeName, attributeValue);
+			this.changed = true;
 		}
 	}
 
 	public void removeAttribute(String attributeName) {
 		this.sessionAttrs.remove(attributeName);
+		this.changed = true;
+	}
+
+	boolean isChanged() {
+		return this.changed;
 	}
 
 	/**
@@ -187,4 +198,8 @@ public final class MapSession implements ExpiringSession, Serializable {
 	}
 
 	private static final long serialVersionUID = 7160779239673823561L;
+
+	void markUnchanged() {
+		this.changed = false;
+	}
 }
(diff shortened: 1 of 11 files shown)


spring-session/src/main/java/org/springframework/session/MapSession.java
```diff
@@ -53,6 +54,8 @@
 	private Map<String, Object> sessionAttrs = new HashMap<String, Object>();
 	private long creationTime = System.currentTimeMillis();
 	private long lastAccessedTime = this.creationTime;
+	private boolean isNew;
```
rwinch: I don't think we need isNew. When the MapSession is new, the id has changed.

spring-session/src/main/java/org/springframework/session/MapSession.java
```diff
@@ -187,4 +203,9 @@ public int hashCode() {
 	}
 
 	private static final long serialVersionUID = 7160779239673823561L;
+
+	public void changeFlags() {
```
rwinch: Can we change this to method to `markUnchanged`

spring-session/src/main/java/org/springframework/session/MapSession.java
```diff
@@ -150,11 +156,21 @@ public void setAttribute(String attributeName, Object attributeValue) {
...
 	public void removeAttribute(String attributeName) {
 		this.sessionAttrs.remove(attributeName);
+		this.changed = true;
+	}
+
+	public boolean isNew() {
+		return this.isNew;
+	}
+
+	public boolean isChanged() {
```
rwinch: I think we should change this to package scope.

spring-session/src/main/java/org/springframework/session/MapSession.java
rwinch: We should also change this to package scope.

rwinch:
Thanks for the PR @eddumelendez! You are a machine :)

I gave some feedback in the code.

stojsavljevic:
Hi,

will we have an option to choose between saving immediately upon change and saving once per request?

Thanks!

spring-session/src/main/java/org/springframework/session/MapSession.java
```diff
@@ -150,11 +155,17 @@ public void setAttribute(String attributeName, Object attributeValue) {
 		}
 		else {
 			this.sessionAttrs.put(attributeName, attributeValue);
+			this.changed = true;
 		}
 	}
 
 	public void removeAttribute(String attributeName) {
 		this.sessionAttrs.remove(attributeName);
+		this.changed = true;
```
tsachev: I wonder if we should mark this as `changed` if `this.sessionAttrs.remove(attributeName)` returns `null`?

spring-session/src/main/java/org/springframework/session/MapSessionRepository.java
```diff
@@ -34,6 +34,7 @@
  * </p>
  *
  * @author Rob Winch
+ * @author Eddú Meléndez
  * @since 1.0
  */
 public class MapSessionRepository implements SessionRepository<ExpiringSession> {
```
tsachev: Maybe we should use `MapSession` in the generic type like in other repositories?
@rwinch Is this breaking change?

rwinch: @tsachev I think  you are right. While this would be a breaking change, it is quite minor and can be acceptable for a minor release (most users leveraging it will likely be using `SessionRepository<? extends ExpiringSession>`.

@eddumelendez Can you update this to use MapSession generic type?

rwinch:
@stojsavljevic Thanks for the question. This PR is not about writing immediately vs writing in a batch. Instead this is about ensuring that if nothing on the session has been set, then it will not be saved again.

@eddumelendez We should probably provide a flag to ensure that the default behavior is preserved. This is necessary for someone who might make changes like this:

```java
Session session = ...;
User user = session.getAttribute("user");
user.setFirstName("New");
```

Currently this change will be visible. However, with the changes, setAttribute will not have been invoked and thus the change will not persist.

For this reason, we should probably add a flag named something like `persistOnSessionChange` which defaults to false.

spring-session/src/main/java/org/springframework/session/MapSessionRepository.java
eddumelendez: @rwinch Sure, I wil.

pivotal-issuemaster:
@eddumelendez Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-session&pullRequestId=528)!

spring-session/src/main/java/org/springframework/session/hazelcast/config/annotation/web/http/EnableHazelcastHttpSession.java
```diff
@@ -71,4 +72,13 @@
...
 	String sessionMapName() default "spring:session:sessions";
 
+	/**
+	 * This is the flag that will be used to persist sessions after
+	 * something has changed or immediately. By default, session will be persisted
+	 * immediately.
+	 *
+	 * @return the flag to persist sessions after changes.
+	 */
+	boolean persistOnSessionChange() default false;
```
stojsavljevic: I find this javadoc misleading. In my mind, this flag is not about when session will be saved (after changes or immediately). It's about avoiding save when changes are not detected. But we must state that changes might not be detected (as @rwinch pointed out). Also, we should state that flag can help to reduce the possibility of race conditions.

vpavic:
Thanks for the PR @eddumelendez, however we have decided not to add more complexity to `MapSessionRepository` as we are not promoting it as production grade `SessionRepository` implementation.

Issue #519, opened 2016-05-12
SessionRepositoryFilterTests doFilterLastAccessedTime fails intermittently


wilkinsona:
The latest [IO Platform compatibility build](https://build.spring.io/browse/PLATFORM-COM-JOB1-101/test/case/185755773) failed with:
```
org.junit.ComparisonFailure: expected:<146305517200[8]L> but was:<146305517200[9]L>
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:57)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at org.springframework.session.web.http.SessionRepositoryFilterTests$4.doFilter(SessionRepositoryFilterTests.java:153)
	at org.springframework.session.web.http.SessionRepositoryFilterTests$DoInFilter.doFilter(SessionRepositoryFilterTests.java:1464)
	at org.springframework.session.web.http.SessionRepositoryFilterTests$89.doFilterInternal(SessionRepositoryFilterTests.java:1454)
	at org.springframework.session.web.http.OncePerRequestFilter.doFilter(OncePerRequestFilter.java:80)
	at org.springframework.mock.web.MockFilterChain.doFilter(MockFilterChain.java:134)
	at org.springframework.session.web.http.SessionRepositoryFilter.doFilterInternal(SessionRepositoryFilter.java:164)
	at org.springframework.session.web.http.OncePerRequestFilter.doFilter(OncePerRequestFilter.java:80)
	at org.springframework.session.web.http.SessionRepositoryFilterTests.doFilter(SessionRepositoryFilterTests.java:1457)
	at org.springframework.session.web.http.SessionRepositoryFilterTests.doFilterLastAccessedTime(SessionRepositoryFilterTests.java:149)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:606)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
[...]

PR #541, opened 2016-06-06
Increase sleep interval in SessionRepositoryFilterTests.doFilterLastAccessedTime

vpavic:
The test failure reported in #519 seems to indicate the test proceeded to next request before assertion for the first part was evaluated, hence the reported 1 millisecond difference.

I've updated the test to sleep for 50 millis, making it consistent with interval used in ```SessionRepositoryFilterTests#doFilterCreateDate``` test.

- [X] I have signed the CLA


rwinch:
Thanks for the PR! This is merged into master

Issue #516, opened 2016-05-10
Session persistance optimization (Hazelcast as storage)


stojsavljevic:
Hi,

I'm using version 1.1.1.RELEASE of Spring Session and Hazelcast 3.6.2 as session storage (_MapSessionRepository_).
I would like to have session persistence highly optimized.

So, it would be great if a session gets persisted only if it has been changed.
Currently, session is always committed.
In my mind, session implementation would have to keep track if something has been changed.

Further, it would be great to have configuration option to choose if we want a repo to be updated immediately upon change (after _session.setAttribute()_ for example) or once per request.
This way we will be able to fine tune performance depending on application's needs.
Also, with immediate write we can avoid some potential concurrent issues.

Hazelcast's _SpringAwareWebFilter_ has both features. Later one is called deferred-write - http://docs.hazelcast.org/docs/3.6.1/manual/html-single/index.html#caching-locally-with-deferred-write-

What do you think about this?
Do you maybe have plans for that?

I would like to have these features in Spring Session because it is a great project!
Thanks for the great work!

rwinch:
@stojsavljevic We'd love to get a pull request for such enhancements. Is this something you would be interested in doing?

vpavic:
:+1:

Similar optimizations for Hazelcast support came across my mind too, especially considering that session implementations for other stores keep track of the changes to optimize their write operations.

If @stojsavljevic can't I'd consider fitting this into my schedule.

stojsavljevic:
Hi guys,

I got green light from my boss to start working on this.

I will contact @vpavic for the way forward.

Thanks!

PR #544, opened 2016-06-09
Add first-class Hazelcast support

vpavic:
Hazelcast support is currently based on generic ```MapSessionRepository``` which implements only ```SessionRepository``` interface. The goal of this PR is to provide first-class support based around dedicated session repository making Hazelcast up to par with other supported backends. This includes:

- ```HazelcastSessionRepository``` which implements ```FindByIndexNameSessionRepository```
- ```spring-session-hazelcast``` module to improve user experience

Implementing ```HazelcastSessionRepository``` required upgrading Hazelcast to ```3.6``` since [custom attributes extraction](http://docs.hazelcast.org/docs/3.6/manual/html-single/index.html#custom-attributes) feature seemed the best way to implement ```FindByIndexNameSessionRepository#findByIndexNameAndIndexValue```. ~~Using this feature presents one challenge though - it requires configuring Hazelcast map (see ```HazelcastSessionRepository#configureSessionMap```) which is not possible when using Hazelcast in client mode.~~

Still a TODO:

- [x] update documentation/javadocs
- [x] ```HazelcastSessionRepository``` unit tests
- [x] update existing integration tests

@rwinch Can you briefly review it and share your initial thoughts before I go further with this? @shakuzen I'd like you to get a quick look too.

/cc @stojsavljevic Having in mind #516, this might be of your interest also.

- [X] I have signed the CLA


pivotal-issuemaster:
@vpavic Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-session&pullRequestId=544)!

rwinch:
@vpavic Thanks for the PR! It appears the build is failing. Any chance you could take a look?

vpavic:
Hey @rwinch, yes, it's failing as I didn't finish the configuration part that sets up the principal name extracting.

I just opened the PR to get your initial thoughts before finishing the config part and other remaining action points.

rwinch:
@vpavic Thanks for the fast response. I should have been more diligent about reading the message.

Is it possible to only write the session attributes that have changed vs saving the entire map every time? This has turned out to be a pretty important approach to reduce the possibility of race conditions.

vpavic:
@rwinch I've updated the PR.

The configuration part is done and the build now passes. I've also added Javadoc on new classes, other TODOs from the original comments are still pending but hopefully I'll get to finish those by the end of this week.

Regarding saving the session only when changes are detected, the new ```HazelcastSessionRepository``` uses plain ```MapSession``` so I'll just make use of the changes done there in #528.

Additionally, I've been in contact with @stojsavljevic over the past few weeks and he's started to do some work on #516 which is somewhat related to this.

rwinch:
@vpavic - These are great changes!

I don't know if it is possible, but I'd like to see an option similar to REDIS and JDBC repositories which only writes the session attribute that changed. For example, if I invoke:

```
session.setAttribute("foo","bar");
repo.save(session);
session.setAttribute("new","test");
repo.save(session);
```

The second save should not need to write the session attribute "foo". This helps reduce the likely hood of a race condition. At minimum it would be good to try to ensure if the last update time is changed, that only that needs to be written vs the entire Session.

This might not be possible with Hazelcast, but I'd like to make sure we investigate this possibility.

vpavic:
@rwinch Thanks for the feedback!

> This might not be possible with Hazelcast, but I'd like to make sure we investigate this possibility.

I'm afraid it's not but I'll look into it over the next few days.

vpavic:
@rwinch I've updated PR with tests and have polished the general implementation a bit.

Regarding the individual attribute writes, both @stojstavljevic and I don't see how we could achieve that without dropping the traditional, single ```Map``` semantics on session persistence and employing a more complicated implementation that would dissect session persistence into multiple ```Map```s.

manderson23:
Do we really want to drop support for Hazelcast in client mode? This was raised as an issue (#339) back when the original implementation was added so could be viewed as a regression.

vpavic:
@manderson23 The client mode support is not dropped with this change.

I assume my original comment confused you - sorry for that. It was a problem in the initial implementation that has been taken care of since, in the subsequent updates of this PR. I'll update to original comment to address this.

Also, you can see that the integration tests for client mode are passing now.

manderson23:
OK thanks. 

FWIW I agree that individual attribute writes would be difficult with Hazelcast. I wouldn't like to see multiple Maps introduced.

vpavic:
@rwinch I've updated the PR with documentation and samples changes.

vpavic:
@rwinch Is it safe to assume this will make the ```1.3.0.M1```?

I've completed all the initially planned TODOs, regarding other changes, once this hits the master @stojsavljevic will follow up with a PR of his own to address the #516. There will also be some changes depending on the outcome of #557.

vpavic:
@rwinch I've updated the PR with some final polish, namely to use the current Hazelcast 3.6.5 release and to apply the changes equivalent to #608.

spring-session/src/main/java/org/springframework/session/hazelcast/HazelcastSessionRepository.java
```diff
@@ -0,0 +1,245 @@
...
+ *
+ * config.getMapConfig("spring:session:sessions")
+ *         .addMapAttributeConfig(attributeConfig)
+ *         .addMapIndexConfig(new MapIndexConfig(
+ *                 HazelcastSessionRepository.PRINCIPAL_NAME_ATTRIBUTE, false));
+ *
+ * Hazelcast.newHazelcastInstance(config);
+ * </pre>
+ *
+ * This implementation listen for events on the Hazelcast-backed SessionRepository and
```
shakuzen: listen → listens

spring-session/src/main/java/org/springframework/session/hazelcast/HazelcastSessionRepository.java
```diff
@@ -0,0 +1,245 @@
...
+ * translate those events into the corresponding Spring Session events. Publish the Spring
```
shakuzen: translate → translates

spring-session/src/main/java/org/springframework/session/hazelcast/HazelcastSessionRepository.java
```diff
@@ -0,0 +1,245 @@
...
+	}
+
+	@PreDestroy
+	private void close() {
+		this.sessions.removeEntryListener(this.sessionListenerId);
+	}
+
+	/**
+	 * Sets the {@link ApplicationEventPublisher} that is used to publish
+	 * {@link SessionDestroyedEvent}. The default is to not publish a
```
shakuzen: This is the publisher for all of the Session events, but the JavaDoc here only mentions the `SessionDestroyedEvent`, which I think might be a little confusing for users.

spring-session/src/main/java/org/springframework/session/hazelcast/HazelcastSessionRepository.java
```diff
@@ -0,0 +1,245 @@
...
+	 */
+	public void setApplicationEventPublisher(
+			ApplicationEventPublisher applicationEventPublisher) {
+		Assert.notNull(applicationEventPublisher,
+				"ApplicationEventPublisher cannot be null");
+		this.eventPublisher = applicationEventPublisher;
+	}
+
+	/**
+	 * Set the maximum inactive interval in seconds between requests before newly created
```
shakuzen: Nit-pick: "newly created" seems to be adding emphasis but I'm not sure why. Any session that has been inactive for the set amount of time will be invalidated, right?

shakuzen:
I left a few comments on JavaDocs. The actual code looks good to me. Great work, @vpavic

vpavic:
@shakuzen Thanks for the review, I've addressed most of you comments with the latest PR update.

spring-session/src/main/java/org/springframework/session/hazelcast/HazelcastSessionRepository.java
vpavic: This description is actually taken from the `RedisOperationsSessionRepository`. IMO such wording is used to emphasis that `defaultMaxInactiveInterval` is actually taken into consideration during creation of events in `SessionRepository#createSession`.

rwinch:
Thanks for the PR! This was merged via 3efb4c9c9e06dcd064ebe4a015f1e5492dd103fd

spring-session/src/main/java/org/springframework/session/hazelcast/HazelcastSessionRepository.java
shakuzen: To me that sounds like leaking implementation details that are not what I would expect a person using this method to need to know about or care about.

PR #557, opened 2016-06-20
Introduce IndexResolver

### LINKED ISSUE #376: Extract IndexResolver
As pointed out in #372 there is currently quite a bit of duplication of logic for `PrincipalNameResolver`. It would be good to extract this logic out in a way that works for any index name and value. However, we should wait and see what shapes up in Spring Data as it may provide a convenient mechanism for this via [Redis Index](https://github.com/spring-projects/spring-data-redis/pull/156/files#diff-d2c04ffe360655561105d34b8ba1d122R31) (I'm in discussion to see if this makes sense to make indexing available in Spring Data Commons

vpavic:
This PR introduces `IndexResolver` as a strategy interface for resolving index values in `FindByIndexNameSessionRepository` implementations.

This PR assumes `IndexResolver` implementations are single purpose, meaning implementations support single index name. Each `FindByIndexNameSessionRepository` implementation then in turn defines its set of supported `IndexResolver`s. For this purpose, a convenience `CompositeIndexResolver` is provided.

Out of the box the `PrincipalNameIndexResolver` is provided as the driver for this improvement, however each `FindByIndexNameSessionRepository`` implementation can easily provide own resolvers, depending on capabilities of backing technology.

This resolves #376.

### DIFF
--- a/spring-session-core/src/main/java/org/springframework/session/DelegatingIndexResolver.java
+++ b/spring-session-core/src/main/java/org/springframework/session/DelegatingIndexResolver.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2014-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.session;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * An {@link IndexResolver} that resolves indexes using multiple @{link IndexResolver}
+ * delegates.
+ *
+ * @param <S> the type of Session being handled
+ * @author Vedran Pavic
+ * @since 2.2.0
+ */
+public class DelegatingIndexResolver<S extends Session> implements IndexResolver<S> {
+
+	private Set<IndexResolver<S>> delegates;
+
+	public DelegatingIndexResolver(Set<IndexResolver<S>> delegates) {
+		this.delegates = Collections.unmodifiableSet(delegates);
+	}
+
+	public Map<String, String> resolveIndexesFor(S session) {
+		Map<String, String> indexes = new HashMap<>();
+		for (IndexResolver<S> delegate : this.delegates) {
+			indexes.putAll(delegate.resolveIndexesFor(session));
+		}
+		return indexes;
+	}
+
+}
--- a/spring-session-core/src/main/java/org/springframework/session/IndexResolver.java
+++ b/spring-session-core/src/main/java/org/springframework/session/IndexResolver.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2014-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.session;
+
+import java.util.Map;
+
+/**
+ * Strategy interface for resolving the {@link Session}'s indexes.
+ *
+ * @param <S> the type of Session being handled
+ * @author Rob Winch
+ * @author Vedran Pavic
+ * @since 2.2.0
+ * @see FindByIndexNameSessionRepository
+ */
+public interface IndexResolver<S extends Session> {
+
+	/**
+	 * Resolve indexes for the session.
+	 * @param session the session
+	 * @return a map of resolved indexes, never {@code null}
+	 */
+	Map<String, String> resolveIndexesFor(S session);
+
+}
--- a/spring-session-core/src/main/java/org/springframework/session/PrincipalNameIndexResolver.java
+++ b/spring-session-core/src/main/java/org/springframework/session/PrincipalNameIndexResolver.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2014-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.session;
+
+import org.springframework.expression.Expression;
+import org.springframework.expression.spel.standard.SpelExpressionParser;
+
+/**
+ * {@link IndexResolver} to resolve the principal name from session attribute named
+ * {@link FindByIndexNameSessionRepository#PRINCIPAL_NAME_INDEX_NAME} or Spring Security
+ * context stored in the session under {@code SPRING_SECURITY_CONTEXT} attribute.
+ *
+ * @param <S> the type of Session being handled
+ * @author Vedran Pavic
+ * @since 2.2.0
+ */
+public class PrincipalNameIndexResolver<S extends Session> extends SingleIndexResolver<S> {
+
+	private static final String SPRING_SECURITY_CONTEXT = "SPRING_SECURITY_CONTEXT";
+
+	private SpelExpressionParser parser = new SpelExpressionParser();
+
+	public PrincipalNameIndexResolver() {
+		super(FindByIndexNameSessionRepository.PRINCIPAL_NAME_INDEX_NAME);
+	}
+
+	public String resolveIndexValueFor(S session) {
+		String principalName = session.getAttribute(getIndexName());
+		if (principalName != null) {
+			return principalName;
+		}
+		Object authentication = session.getAttribute(SPRING_SECURITY_CONTEXT);
+		if (authentication != null) {
+			Expression expression = this.parser.parseExpression("authentication?.name");
+			return expression.getValue(authentication, String.class);
+		}
+		return null;
+	}
+
+}
--- a/spring-session-core/src/main/java/org/springframework/session/SingleIndexResolver.java
+++ b/spring-session-core/src/main/java/org/springframework/session/SingleIndexResolver.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2014-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.session;
+
+import java.util.Collections;
+import java.util.Map;
+
+import org.springframework.util.Assert;
+
+/**
+ * Base class for {@link IndexResolver}s that resolve a single index.
+ *
+ * @param <S> the type of Session being handled
+ * @author Rob Winch
+ * @author Vedran Pavic
+ * @since 2.2.0
+ */
+public abstract class SingleIndexResolver<S extends Session> implements IndexResolver<S> {
+
+	private final String indexName;
+
+	protected SingleIndexResolver(String indexName) {
+		Assert.notNull(indexName, "Index name must not be null");
+		this.indexName = indexName;
+	}
+
+	protected String getIndexName() {
+		return this.indexName;
+	}
+
+	public abstract String resolveIndexValueFor(S session);
+
+	public final Map<String, String> resolveIndexesFor(S session) {
+		String indexValue = resolveIndexValueFor(session);
+		return (indexValue != null) ? Collections.singletonMap(this.indexName, indexValue) : Collections.emptyMap();
+	}
+
+}
--- a/spring-session-core/src/test/java/org/springframework/session/DelegatingIndexResolverTests.java
+++ b/spring-session-core/src/test/java/org/springframework/session/DelegatingIndexResolverTests.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2014-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.session;
+
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Tests for {@link DelegatingIndexResolver}.
+ *
+ * @author Vedran Pavic
+ */
+class DelegatingIndexResolverTests {
+
+	private DelegatingIndexResolver<MapSession> indexResolver;
+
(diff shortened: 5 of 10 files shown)


pivotal-issuemaster:
@vpavic Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-session&pullRequestId=557)!

spring-session/src/main/java/org/springframework/session/AbstractIndexResolver.java
```diff
@@ -0,0 +1,48 @@
...
+			return null;
+		}
+		return doResolve(session);
+	}
+
+	/**
+	 * Indicate supported index name.
+	 * @return the index name
+	 */
+	protected abstract String supportedIndexName();
```
rwinch: What do we do if it supports multiple indexes?

spring-session/src/main/java/org/springframework/session/IndexResolver.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+public interface IndexResolver<S extends Session> {
+
+	/**
+	 * Resolve the value of the given index from the session.
+	 * @param session the session
+	 * @param indexName the index name to resolve
+	 * @return the resolved index value, {@code null} if none found
+	 */
+	String resolve(S session, String indexName);
```
rwinch: Perhaps the input should just be the session and the result that has a mapping of index name to index value. Is there any advantage in only supporting a single session attribute?

spring-session/src/main/java/org/springframework/session/AbstractIndexResolver.java
vpavic: Well, this is basically just an abstract base class that simplifies implementing single-purpose ```IndexResolver```. Nothing in the ```IndexResolver``` contract itself prevents the implementations from supporting multiple indexes, as demonstrated by ```CompositeIndexResolver```.

spring-session/src/main/java/org/springframework/session/IndexResolver.java
vpavic: IMO that wouldn't be optimal solution since it implies execution of logic for all supported indexes. This way you extract only what you need in the given situation.

rwinch: How would implementations that support multiple indexes (i.e. Gemfire) use this API?

vpavic: Each ```FindByIndexNameSessionRepository``` implementation would internally define its supported ```IndexResolver```s, either using a ```CompositeIndexResolver``` or some other way. These would then get consumed according to repository needs, i.e. the same way duplicated principal name resolving logic is consumed now.

This approach allows having some resolvers that are reused across all repository implementations, such as ```PrincipalNameIndexResolver```, as well as each implementation having its specific resolvers. If you would go the _input should just be the session and the result that has a mapping of index name to index value_ route then you wouldn't be able to reuse this logic.

rwinch: > either using a CompositeIndexResolver or some other way.

This requires another API because the current interface must have knowledge of the index name.

> If you would go the input should just be the session and the result that has a mapping of index name to index value route then you wouldn't be able to reuse this logic.

Why not? Seems you could do this:

```java

public interface IndexResolver<S extends Session> {
	Map<String,String> resolveIndexes(S session);
}

public class PrincipalNameIndexResolver<S extends Session> implements IndexResolver<S> {
	private SpelExpressionParser parser = new SpelExpressionParser();
	private String indexName = FindByIndexNameSessionRepository.PRINCIPAL_NAME_INDEX_NAME;

	public String resolvePrincipal(Session session) {
		String principalName = session.getAttribute(FindByIndexNameSessionRepository.PRINCIPAL_NAME_INDEX_NAME);
		if(principalName != null) {
			return principalName;
		}
		Object authentication = session.getAttribute("SPRING_SECURITY_CONTEXT");
		if(authentication != null) {
			Expression expression = parser.parseExpression("authentication?.name");
			return expression.getValue(authentication, String.class);
		}
		return null;
	}

	@Override
	public Map<String, String> resolveIndexes(S session) {
		String index = resolvePrincipal(session);
		if(index == null) {
			return Collections.emptyMap();
		}
		return Collections.singletonMap(indexName, index);
	}
}

public class GemfireIndexResolver<S extends Session> implements IndexResolver<S> {
	private PrincipalNameIndexResolver<S> principalResolver = new PrincipalNameIndexResolver<>();

	@Override
	public Map<String, String> resolveIndexes(S session) {
		Map<String, String> gemfireIndexes = resolveThem(session);
		gemfireIndexes.putAll(principalResolver.resolveIndexes(session));
		return gemfireIndexes;
	}

	private Map<String, String> resolveThem(S session) {
		// TODO
		return null;
	}
}
```

vpavic: > This requires another API because the current interface must have knowledge of the index name.

Yes, that's by design since I originally intended ```IndexResolver``` implementations to be single purpose. Each ```FindByIndexNameSessionRepository``` implementation holds the knowledge of its supported indexes anyway, doesn't it? Or did you mean _another API_ as in API targeted at users? Perhaps ```FindByIndexNameSessionRepository``` interface itself should communicate the index names it is capable of resolving?

> Why not? Seems you could do this:

That kind of implementation is exactly what I wanted to avoid :) Your ```GemfireIndexResolver``` will get messy once other ```IndexResolver``` resolvers get in there too since you're relying on the concrete ```IndexResolver``` implementation.

Regarding your proposed ```IndexResolver``` API and Gemfire implementation, consider things from perspective of current usage of ```PrincipalNameResolver```s and equivalent logic - you'd really like to run the resolving for all the indexes and then extract the single value you're interested in instead of current ```PrincipalNameResolver#resolvePrincipal```?

rwinch: > Each FindByIndexNameSessionRepository implementation holds the knowledge of its supported indexes anyway, doesn't it? 

It does. Introducing this new API we have yet another API that needs to be aware of the index names and be kept in sync. The proposed change allows all the index logic to be moved into the new resolver api.

> Your GemfireIndexResolver will get messy once other IndexResolver resolvers get in there too since you're relying on the concrete IndexResolver implementation.

How would this be cleaner with the original design? Either way you would need to have that specific functionality.

When are we interested in only a single index?

vpavic: I've updated the PR with changes to ```FindByIndexNameSessionRepository``` implementations to better demonstrate the intended usage of ```IndexResolver```s

> When are we interested in only a single index?

Well, at any given time you're attempting to resolve an index you're interested in a single value, right? Unless I'm missing something obvious.

rwinch: Most the time I want to write all the index's (that have changed) to the data store. This could easily be done with the current API.

If I'm looking up an index, then the user provides the index name and the index value so I do not need to resolve anything.

vpavic: But this is dependent on the technology used to implement the repository, isn't it? For example, consider the [PrincipalNameExtractor](https://github.com/spring-projects/spring-session/pull/544/files#diff-961fafd60c6c9cf34e7772ea7993d626) from #544. It's basically a wrapper around an ```IndexResolver```.

OK, it certainly makes sense to provide API to communicate all the supported indexes. What are you thoughts on making it a part of ```FindByIndexNameSessionRepository```?

rwinch: > It's basically a wrapper around an IndexResolver

Right, but the logic to resolve the index would be able to be reused. This is the goal. If we need adapters to translate the results of the IndexResolver into implementation specific code this is much easier than rewriting the code for resolving the indexes.

> What are you thoughts on making it a part of FindByIndexNameSessionRepository?

* This interface is already GA, so we cannot add a new method on it
* I don't think this is the right place for it. User's do not care what the indexes are. It is the internal implementation that cares about the indexes. So the `FindByIndexNameSessionRepository` implementation would have an `IndexResolver` injected into it.

vpavic:
@rwinch I've updated the PR to address the concerns from our previous discussion.

```IndexResolver``` now provides operations to indicate the supported indexes, and resolve all indexes. It also includes ```Delegate``` contract for delegates such as ```PrincipalNameIndexResolver```.

```FindByIndexNameSessionRepository``` implementations currently do not adopt these changes. I'll do that part after your review.

**Comment by rwinch** [comment]
@vpavic Thanks for the update. This doesn't appear to be what I had in mind. Is there a reason we need anything other than:

```java
public interface IndexResolver<S extends Session> {
    Map<String,String> resolveIndexes(S session);
}
```

as outlined in my comments?

vpavic:
Having the possibility to check whether given index name is supported by the resolver vastly simplifies the checks found in ```FindByIndexNameSessionRepository#findByIndexNameAndIndexValue``` implementations. This can now be handled via single call of ```IndexResolver#supportsIndex```. I thought you referred to that by _Introducing this new API we have yet another API that needs to be aware of the index names and be kept in sync_.

```Delegate``` spec simplifies the ```IndexResolver``` implementation, with the provided ```DefaultIndexResolver``` IMO covers all the use cases and avoids having ugly ```if-then-else``` logic in ```IndexResolver``` implementations.

Resolving a single index is also pretty common case IMO, there are two examples in ```RedisOperationsSessionRepository``` alone:

- [```cleanupPrincipalIndex```](https://github.com/spring-projects/spring-session/blob/master/spring-session/src/main/java/org/springframework/session/data/redis/RedisOperationsSessionRepository.java#L543)
- [```RedisSession``` constructor](https://github.com/spring-projects/spring-session/blob/master/spring-session/src/main/java/org/springframework/session/data/redis/RedisOperationsSessionRepository.java#L699)

Also with this design we could easily support users adding their own indexes with some ```FindByIndexNameSessionRepository``` implementations - they would simply implement ```Delegate``` contract and register beans which we would then get automatically register with ```IndexResolver```.

**Comment by rwinch** [comment]
@vpavic Thanks for your response.

> Having the possibility to check whether given index name is supported by the resolver vastly simplifies the checks found in FindByIndexNameSessionRepository#findByIndexNameAndIndexValue implementations. 

Ok. I see your point. Sorry for the confusion. It might be good to continue doing this sort of optimization.

I was planning on having the implementation just attempt to perform the lookup regardless of if the index was supported. If the value wasn't indexed, it would be a null session.

Do you see reason for having both `supportedIndexes` and `supportsIndex` method?

I'd really like to keep this API as simple as possible. I think it may be valuable to remove both the `supportedIndexes` and `supportsIndex` methods. It seems that we will have all the information we need with a single method.

> Delegate spec simplifies the IndexResolver implementation, with the provided DefaultIndexResolver IMO covers all the use cases and avoids having ugly if-then-else logic in IndexResolver implementations.
> ...
> Resolving a single index is also pretty common case IMO

Agreed. This is a common use case.

Could this be an abstract class rather than an interface?

> Also with this design we could easily support users adding their own indexes with some FindByIndexNameSessionRepository implementations - they would simply implement Delegate contract and register beans which we would then get automatically register with IndexResolver.

We can do the same by having them provide multiple `FindByIndexNameSessionRepository` implementations. Right?

**Comment by vpavic** [comment]
> I was planning on having the implementation just attempt to perform the lookup regardless of if the index was supported. If the value wasn't indexed, it would be a null session.

I see, however I was driven by optimization considerations from the start - why run the resolver logic if we have enough knowledge to know it won't yield any result?

> Do you see reason for having both ```supportedIndexes``` and ```supportsIndex``` method?

I agree we probably don't need both, ```supportsIndex``` is by far more relevant for a typical usage of ```IndexResolver```.

> I'd really like to keep this API as simple as possible. I think it may be valuable to remove both the ```supportedIndexes``` and ```supportsIndex``` methods. It seems that we will have all the information we need with a single method.

Could you please clarify this a bit more? What single method do you refer to if we remove the both support-methods? Or did you mean remove one of the methods rather than remove both?

> Could this be an abstract class rather than an interface?

Yes, I've also considered such approach.

> We can do the same by having them provide multiple ```FindByIndexNameSessionRepository``` implementations. Right?

Right. But I feel there should be a single ```SessionRepository``` bean in the app context, anything else requires some sort of compromise when it comes to consuming the repository.

So to sum things up before I make an update, we'd trim the ```IndexResolver``` contract to something like this:

```java
public interface IndexResolver<S extends Session> {

    boolean supportsIndex(String indexName);

    String resolve(S session, String indexName);

    Map<String, String> resolveAll(S session);

}
```

The ```Delegate``` spec would be moved to the ```AbstractIndexResolver```.

rwinch:
@vpavic 

> I see, however I was driven by optimization considerations from the start - why run the resolver logic if we have enough knowledge to know it won't yield any result?

We don't know that we are the only repository writing to the data store. There might be something else providing the indexes. Perhaps a background job or another application. 

There is a cost to maintaining this additional code with (in my opinion) little gain. In short, I still think this needs to be removed.

I really think our API should look closer to:

```java
public interface IndexResolver<S extends Session> {

    Map<String, String> resolveIndexesFor(S session);
}
```

Then we can provide an abstract class that supports resolving a single index. For example:

```java
public abstract class AbstractSingleIndexResolver<S extends Session>  implements IndexResolver<S> {
    private final String sessionIndexName;

    protected AbstractSingleIndexResolver(String sessionIndexName) {
        this.sessionIndexName = sessionIndexName;
    }

    protected abstract String resolveIndexValueFor(S session);

    public final Map<String,String> resolveIndexesFor(S session) {
        String sessionIndexValue = resolveIndexValueFor(session);
        return sessionIndexValue == null ? Collections.<String,String>emtpyMap() : Collections.singletonMap(sessionIndexName, sessionIndexValue);
    }
}
```

vpavic:
> We don't know that we are the only repository writing to the data store. There might be something else providing the indexes. Perhaps a background job or another application.

@rwinch Thanks for the insight, all clear now, I wasn't aware that this was an option to consider.

I can update the PR with your suggested API within few days and try to wire it altogether into repository implementations.

rwinch:
@vpavic Thanks!

vpavic:
@rwinch I've updated the PR. Please verify that the implementation is now OK with you, before I continue forward.

rwinch:
@vpavic Sorry for the delay getting back to you. Yes this is what I had in mind. Thanks again!

rwinch:
@vpavic I should have been more explicit. I can merge this once you are able to add testing and documentation.

Thanks again for everything you do for Spring Session and Spring in general!

vpavic:
@rwinch Thanks for the feedback!

I'll try to get back to this next week and integrate the `IndexResolver` into `FindByIndexNameSessionRepository` implementations, as well as write the tests.

spring-session/src/main/java/org/springframework/session/data/gemfire/AbstractGemFireOperationsSessionRepository.java
```diff
@@ -495,18 +503,8 @@ public synchronized void setPrincipalName(String principalName) {
...
-				Object authentication = getAttribute(SPRING_SECURITY_CONTEXT);
-
-				if (authentication != null) {
-					Expression expression = this.parser.parseExpression("authentication?.name");
-					principalName = expression.getValue(authentication, String.class);
-				}
-			}
-
-			return principalName;
+			Map<String, String> indexes = indexResolver.resolveIndexesFor(this);
```
vpavic: This would benefit from `IndexResolver` having a method to resolve single index. However the GemFire repository implementation is the one I'm least familiar with, so this altogether could perhaps be refactored in a more optimal way.

spring-session/src/main/java/org/springframework/session/data/redis/RedisOperationsSessionRepository.java
```diff
@@ -540,7 +543,9 @@ private void cleanupPrincipalIndex(RedisSession session) {
 			return;
 		}
 		String sessionId = session.getId();
-		String principal = PRINCIPAL_NAME_RESOLVER.resolvePrincipal(session);
+		Map<String, String> indexes = RedisOperationsSessionRepository.this
+				.indexResolver.resolveIndexesFor(session);
```
vpavic: This would benefit from `IndexResolver` having a method to resolve single index.

spring-session/src/main/java/org/springframework/session/data/redis/RedisOperationsSessionRepository.java
```diff
@@ -696,7 +708,9 @@ static String getSessionAttrNameKey(String attributeName) {
 		RedisSession(MapSession cached) {
 			Assert.notNull("MapSession cannot be null");
 			this.cached = cached;
-			this.originalPrincipalName = PRINCIPAL_NAME_RESOLVER.resolvePrincipal(this);
+			Map<String, String> indexes = RedisOperationsSessionRepository.this
+					.indexResolver.resolveIndexesFor(this);
```
vpavic: This would benefit from `IndexResolver` having a method to resolve single index.

spring-session/src/main/java/org/springframework/session/hazelcast/PrincipalNameExtractor.java
```diff
@@ -34,42 +32,16 @@
...
-	private static final PrincipalNameResolver PRINCIPAL_NAME_RESOLVER =
-			new PrincipalNameResolver();
+	private final SingleIndexResolver<MapSession> indexResolver =
+			new PrincipalNameIndexResolver<MapSession>();
 
 	@SuppressWarnings("unchecked")
 	public void extract(MapSession target, String argument,
 			ValueCollector collector) {
-		String principalName = PRINCIPAL_NAME_RESOLVER.resolvePrincipal(target);
+		String principalName = this.indexResolver.resolveIndexValueFor(target);
```
vpavic: This would benefit from `IndexResolver` having a method to resolve single index. ATM I'm using `SingleIndexResolver` directly here, which I had to change a bit (make `SingleIndexResolver#resolveIndexValueFor` `public` instead of `protected` as originally outlined). `ValueExtractor`s are instantiated by Hazelcast, so perhaps the best would be to have Hazelcast specific `IndexResolver` implementation.

vpavic:
@rwinch I've updated the PR with tests and refactored `FindByIndexNameSessionRepository` implementations to use the new API.

I've also left some comments on the code concerning the previously discussed availability of method that resolves a single index.

vpavic:
@rwinch Any feedback on the latest updates? Since there are two PRs with new repository implementations scheduled for `1.3.0.M3` as well it might be good idea to get this one in earlier because it will have impact on them.

vpavic:
Rebased the PR against `1.4.x`.

vpavic:
With #1145 resolved, I've finally revisited this and rebased on `master`. I believe this is now inline with your feedback @rwinch - it's been a while, but maybe we can finally get this merged for `2.2.0`.

PR #562, opened 2016-06-28
Fix issue when using SpringMVC to issue redirect in controller

edeandrea:
Fix to https://support.pivotal.io/tickets/30469 & issue #563 

I do notice 2 bugs in the spring-session code, both inside the
CookieHttpSessionStrategy class. I have a small sample spring-boot application which shows this issue that I can provide if needed.

The first, using their own sample application, after getting a new
session alias from the HttpSessionManager, it calls
HttpSessionManager.encodeURL and then stores the output into a request
attribute which is later rendered. However, in a Spring MVC
application, a Controller may want to issue a redirect to a
newly-encoded URL, like this:

HttpSessionManager sessionManager = (HttpSessionManager)
request.getAttribute(HttpSessionManager.class.getName());
String newSessionAlias = sessionManager.getNewSessionAlias(request);
String currentSessionAlias =
sessionManager.getCurrentSessionAlias(request);

return String.format("redirect:%s",
sessionManager.encodeURL("/test", newSessionAlias));

The problem here is that Spring MVC will funnel the redirect back
through
CookieHttpSessionStrategy.MultiSessionHttpServletResponse.encodeRedirect
URL, which then looks up the current alias & re-encodes with that
alias, effectively replacing the new alias which the controller added.
The fix would be for the
CookieHttpSessionStrategy.MultiSessionHttpServletResponse to examine
the inputted url for the appropriate parameter & don't re-encode if it
finds it there.

The second issue is that inside CookieHttpSessionStrategy line 346
(return path + "?" + query;) is that if there is no query string (on
the default session alias) then the resulting URL will end in a ? with
nothing after it. In that situation the ? should be removed as well
because the query string is empty.


pivotal-issuemaster:
@edeandrea Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-session&pullRequestId=562)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-session&pullRequestId=562) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/faq) for frequently asked questions.

pivotal-issuemaster:
@edeandrea Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-session&pullRequestId=562)!

rwinch:
Thanks for the PR!

* Can you please add tests?
* Spring Session's web module cannot rely on Spring framework. Any chance you can update the PR to no longer use UriComponents?

edeandrea:
I can surely do that. As for the module not relying on Spring Framework - I can certainly do that. My initial reaction that that though is that if the dependency isn't allowed - shouldn't the build fail with a compilation error? The fact that it didn't means that Spring Framework is on the compile classpath (I checked before I made the change as I did not want to introduce a dependency that wasn't already there). If indeed the Spring Framework shouldn't be there, then the build itself should be cleaned up as well so that the dependency isn't there to begin with.

edeandrea:
I think I noticed another bug as well in CookieHttpSessionStrategy.encodeUrl.

Run this test - it fails...

```java
@Test
public void encodeURLWithSameAlias() {
  String url = String.format("/url?%s=1", CookieHttpSessionStrategy.DEFAULT_SESSION_ALIAS_PARAM_NAME);
  assertThat(this.strategy.encodeURL(url, "1")).isEqualTo(url);
}
```

It will give you this output:
org.junit.ComparisonFailure: expected:<"/url?_s=1[]"> but was:<"/url?_s=1[&_s=1]">

I will fix this as well in my next PR.

rwinch:
Thanks for the updates! This is merged into master via bff0f8f8452bb4b87a6576fe66168b89f50602b3

PR #582, opened 2016-08-02
Optimize `save` operation in `JdbcOperationsSessionRepository`

vpavic:
This PR improves saving of new sessions to only execute batch update operation if there really are any attributes to save.

Also I've found the current implementation to be problematic with HSQLDB, and results in _statement is not in batch mode_ error when attempting to save a new session that does not contain any attributes.


rwinch:
@vpavic Thanks for the PR! Any chance we could get a test added?

vpavic:
@rwinch Done - I've updated the PR.

rwinch:
Thanks for the fast turnaround @vpavic! This is now merged into master

PR #587, opened 2016-08-05
Use webjars-locator in samples

### LINKED ISSUE #566: Samples should use webjars-locator
We should update the samples to use webjars-locator so that the samples do not need to reference the specific version of the assets.

Relates to #458

vpavic:
This resolves #566.


rwinch:
Thanks for the PR @vpavic! This is now merged into master

PR #598, opened 2016-08-17
Add support for custom Prefix

### LINKED ISSUE #597: Allow custom Redis namespace prefix
Allow customization of Default Redis Namespace.

e.g.: `spring:session:` to `foo:bar:`.

Currently `spring:session:` is hardcoded

The scenario is that i currently have multiple services that will consult the Session Data within Redis and i don't want the Namespace to begin with `spring:session:`

Even if i customize it through `@EnableRedisHttpSession(redisNamespace = "foo")` i still get `spring:session:foo` instead of `foo`.

I suggest creating an extra property on the Annotation like `defaultPrefix` to set this value, that defaults to `spring:session:`

driverpt:
This closes #597

### DIFF
--- a/spring-session/src/main/java/org/springframework/session/data/redis/RedisOperationsSessionRepository.java
+++ b/spring-session/src/main/java/org/springframework/session/data/redis/RedisOperationsSessionRepository.java
@@ -288,7 +288,7 @@ public class RedisOperationsSessionRepository implements
 	static final String SESSION_ATTR_PREFIX = "sessionAttr:";
 
 	/**
-	 * Default Attribute separator used
+	 * Default Attribute separator used.
 	 */
 	static final String DEFAULT_ATTR_SEPARATOR = ":";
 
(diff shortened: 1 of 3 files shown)


driverpt:
Any feedback on this ?

vpavic:
@driverpt Thanks for the PR. We'd prefer a simpler approach where `RedisOperationsSessionRepository#setRedisKeyNamespace` would configure the entire namespace, not just the part of it. Are you willing to update the PR in that direction?

driverpt:
What do you mean with "entire namespace"?

vpavic:
Sorry for not being clear enough.

With the current state, invoking `RedisOperationsSessionRepository#setRedisKeyNamespace` with `test` will result in `spring:session:test:` namespace. It would be more straightforward if this resulted in `test:` namespace. This would be much simpler and I believe it would address your concern from #597.

driverpt:
@vpavic , the only reason i haven't changed ```RedisOperationsSessionRepository#setRedisKeyNamespace``` behaviour was to maintain backwards compatibility.
It's highly possible that you have systems relying on this behaviour, and changing it might cause an "earthquake".
I can change the code, but it may cause backwards-compatibility problems.

vpavic:
@driverpt 2.0 is our chance to make such change, together with other non-passive one we have already prepared for the new major release.

driverpt:
@vpavic , ok, i'll change this PR. But i need to rebase it against 2.0 branch

vpavic:
Thanks! Note that the target branch is still `master`.

driverpt:
@vpavic , created a new PR. #919

PR #608, opened 2016-08-31
Polish JDBC configuration

vpavic:
This PR improves JDBC configuration by reusing the existing constants to remove duplication.


rwinch:
Thanks for the PR @vpavic! This is now merged into master

PR #610, opened 2016-09-01
Use bean factory class loader for deserializing JDBC sessions

vpavic:
This commit addresses the issue with deserializing JDBC sessions in Spring Boot applications that use DevTools. Previously, such configuration would cause `ClassCastException` when deserializing JDBC sessions due to app class loader being used instead of restart class loader.

See spring-projects/spring-boot#3805 for more background.

### DIFF
--- a/spring-session/src/main/java/org/springframework/session/jdbc/config/annotation/web/http/JdbcHttpSessionConfiguration.java
+++ b/spring-session/src/main/java/org/springframework/session/jdbc/config/annotation/web/http/JdbcHttpSessionConfiguration.java
@@ -19,6 +19,7 @@ import java.util.Map;
 
 import javax.sql.DataSource;
 
+import org.springframework.beans.factory.BeanClassLoaderAware;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.context.annotation.Bean;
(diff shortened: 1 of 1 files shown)


spring-session/src/main/java/org/springframework/session/jdbc/JdbcOperationsSessionRepository.java
```diff
@@ -261,6 +261,22 @@ public void setConversionService(ConversionService conversionService) {
 		this.conversionService = conversionService;
 	}
 
+	/**
+	 * Set the {@link ClassLoader} to use for deserialization with the default
+	 * {@link ConversionService}. Applicable only with Spring 4.2.1 or later.
+	 * @param classLoader the bean class loader
+	 */
+	public void setClassLoader(ClassLoader classLoader) {
```
rwinch: User's might be a bit surprised to find out if the manually invoke `setConversionService` and then `setClassLoader` that the `ConversionService` is overridden.

Perhaps a better approach could be to remove the `setClassLoader` method and have the configuration class create a default `ConversionService`  with the appropriate `ClassLoader` if the user has not explicitly provided one.

vpavic: Agreed @rwinch, that scenario came across my mind today as well.

vpavic:
@rwinch I've updated the PR as per your suggestions.

Issue #609, opened 2016-08-31
Provide support for customizing SQL queries in `JdbcOperationsSessionRepository`


vpavic:
In order to address the #580 properly, we need to be able to customize SQL queries used by `JdbcOperationsSessionRepository`.

I'll provide the PR to address both this and #580 shortly.

PR #613, opened 2016-09-04
Fix `cleanUpExpiredSessions` operation in `JdbcOperationsSessionRepository`

### LINKED ISSUE #580: JdbcOperationsSessionRepository does not respect per Session expiration times
I am using Spring Session v1.2.1 with Postgres. The sessions are persisted into my Postgres database using the `JdbcOperationsSessionRepository`.

The default inactive time that I am using for my sessions is 30 days. Once a user logs in, I do `request.getSession().setMaxInactiveInterval()` and change it to 180 days.

However for some reason the 180 days is not being respected and sessions keep getting deleted every hour. For example:

This session should have lasted 180 days but it got deleted as soon as the next hour started.

Does anyone know how I can stop these sessions from getting deleted and have them stick around for 180 days?

I am guessing [this function](https://github.com/spring-projects/spring-session/blob/master/spring-session/src/main/java/org/springframework/session/jdbc/JdbcOperationsSessionRepository.java#L464) has something to do with this.

Here's the part of my application that extends the user's max inactive interval:

```
private void setupSession(HttpServletRequest request, User user) {
   HttpSession session = request.getSession();
   session.setMaxInactiveInterval(15552000);
   session.setAttribute("user-id", user.id);
}
```

vpavic:
Currently, `JdbcOperationsSessionRepository#cleanUpExpiredSessions` only considers the repository defined max inactive interval which causes incorrect cleanup of sessions that define custom inactive interval. This commit fixes the problem by delegating calculation of deletion interval to the underlying SQL DELETE statement.

Additionally, due to shortcomings in H2 and Derby databases which make them incompatible with newly introduced DELETE statement, the setters for customizing SQL queries have been also added.

This resolves #580 and #609.

### DIFF
--- a/docs/src/docs/asciidoc/index.adoc
+++ b/docs/src/docs/asciidoc/index.adoc
@@ -1080,6 +1080,7 @@ However, you can override the default `ConversionService` by providing a Bean na
 
 By default, this implementation uses `SPRING_SESSION` and `SPRING_SESSION_ATTRIBUTES` tables to store sessions.
 Note that the table name can be easily customized as already described. In that case the table used to store attributes will be named using the provided table name, suffixed with `_ATTRIBUTES`.
+If further customizations are needed, SQL queries used by the repository can be customized using `set*Query` setter methods. In this case you need to manually configure the `sessionRepository` bean.
 
 Due to the differences between the various database vendors, especially when it comes to storing binary data, make sure to use SQL script specific to your database.
 Scripts for most major database vendors are packaged as `org/springframework/session/jdbc/schema-\*.sql`, where `*` is the target database type.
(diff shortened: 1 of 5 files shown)


rwinch:
Thanks for the PR @vpavic! 

You can use a single PR, but can you please split the commits into two distinct commits for the specific issues?

I'm wondering if we can improve this slightly so that the `JdbcHttpSessionConfiguration` detects the Database vendor and updates the query for H2 and Derby using something like:

```

@Bean
public JdbcOperationsSessionRepository sessionRepository(
		@Qualifier("springSessionJdbcOperations") JdbcTemplate jdbc,
		PlatformTransactionManager transactionManager) throws Exception {
	String vendor = jdbc.getDataSource().getConnection().getMetaData().getDatabaseProductName();
	boolean isH2 = "h2".equalsIgnoreCase(vendor);
        ...

```

Thoughts?

vpavic:
> You can use a single PR, but can you please split the commits into two distinct commits for the specific issues?

Done.

> I'm wondering if we can improve this slightly so that the JdbcHttpSessionConfiguration detects the Database vendor and updates the query for H2 and Derby

I like the idea. Done as well with the latest PR update.

rwinch:

PR #614, opened 2016-09-05
Add Base64 encoding option to `DefaultCookieSerializer`

### LINKED ISSUE #611: Add Base64 encoding option to DefaultCookieSerializer
In order to support [RFC 6265](https://tools.ietf.org/html/rfc6265) we should consider adding Base64 encoding option to `DefaultCookieSerializer` since the spec states:

> To maximize compatibility with user agents, servers that wish to
> store arbitrary data in a cookie-value SHOULD encode that data, for
> example, using Base64

**NOTE**: 

* We must remain JDK 5 compliant, so we will likely need to copy code from Spring Security into a package private class. See [Base64](https://github.com/spring-projects/spring-security/blob/37c6605062531f17e10a8e9442df2cbd61636322/crypto/src/main/java/org/springframework/security/crypto/codec/Base64.java) and [Base64Tests](https://github.com/spring-projects/spring-security/blob/37c6605062531f17e10a8e9442df2cbd61636322/crypto/src/test/java/org/springframework/security/crypto/codec/Base64Tests.java)
* We should also be safe when decoding the value. If we get an error decoding the cookie value, we should return `null`. This prevents internal server errors if a user has a cookie with a non-base64 encoded value (i.e. there was a migration). Additionally, it prevents information leakage in the event a malicious user specifies an invalid value

vpavic:
This resolves #611.

### DIFF
--- a/spring-session/src/main/java/org/springframework/session/web/http/Base64.java
+++ b/spring-session/src/main/java/org/springframework/session/web/http/Base64.java
@@ -1,0 +1,640 @@
+/*
+ * Copyright 2014-2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


rwinch:
Fixed in 2792d2a0e919e26a37e4599c56eda31943d52035 and polished in 97fef0f9bd1119cdf3dc83e56b4f17802af9abe6

Issue #616, opened 2016-09-07
Allow custom cron expression for cleanup task


ajthom90:
I have a Spring Boot application using Spring Session 1.2.1.RELEASE.  I have two servers running the application behind a load balancer, and the session information is stored in MySQL databases that use master-master replication.  When the Scheduled task runs to delete sessions (JdbcOperationsSessionRepository.cleanUpExpiredSessions()), it runs on both web servers, which then tries to run on both databases and the replication runs into issues.  

It would be nice to be able to set a system property so the scheduled cleanUpExpiredSessions() method will only run on one database, and MySQL replication will take care of the rest behind the scenes.

rwinch:
Thanks for the report!

I'm not sure it is quite that simple since you need fail-over. What happens when one of the server instances crashes? Nothing is cleaning up expired sessions then.

ajthom90:
Ahh, very good point!  Hadn't thought about what would happen if the server in charge of cleaning up sessions goes down.

vpavic:
@rwinch We could expose the cron expression used by `JdbcOperationsSessionRepository#cleanUpExpiredSessions` as a configurable parameter. Users could then adjust the expression to ensure the executions on multiple nodes do not overlap.

ajthom90:
I created Pull Request #621 with @vpavic's suggestion about exposing the cron expression.  Let me know what you think

(sorry it took me so many commits to get a passing build... apparently I decided to ignore Checkstyle today)

vpavic:
@ajthom90 Having in mind this situation can be viewed as an edge case, perhaps it's not worth the complexity it would add to the configuration support. IMO it falls in the same category as custom SQL queries which will be introduced as a part of #613. Setting custom SQL will require users to manually declare `JdbcOperationsSessionRepository` bean.

That being said, you could easily work around your problem with a custom `JdbcOperationsSessionRepository` like this example:

```java
public class CustomJdbcOperationsSessionRepository
		extends JdbcOperationsSessionRepository {

	public CustomJdbcOperationsSessionRepository(JdbcOperations jdbcOperations,
			PlatformTransactionManager transactionManager) {
		super(jdbcOperations, transactionManager);
	}

	@Override
	@Scheduled(cron = "desired cron expression")
	public void cleanUpExpiredSessions() {
		super.cleanUpExpiredSessions();
	}

}
```

[...]

PR #621, opened 2016-09-08
Allow using a system property to set cron task of cleaning up expired JDBC sessions

ajthom90:
Issue #616 

Allow using a system property to set CRON expression for cleaning up expired JDBC sessions.  It looks like the `@Scheduled` annotation wouldn't allow a placeholder to be resolved with a system property, so adding the task to the task registrar is now handled in the `JdbcHttpSessionConfiguration`.  Also, there is now a property on `@EnableJdbcHttpSession` to set the CRON expression.

### DIFF
--- a/spring-session/src/main/java/org/springframework/session/jdbc/JdbcOperationsSessionRepository.java
+++ b/spring-session/src/main/java/org/springframework/session/jdbc/JdbcOperationsSessionRepository.java
@@ -134,6 +134,9 @@ public class JdbcOperationsSessionRepository implements
 	 */
 	public static final String DEFAULT_TABLE_NAME = "SPRING_SESSION";
 
+	/**
+	 * The default CRON expression used by Spring Session to clean up expired sessions.
+	 */
 	public static final String CLEAN_UP_EXPIRED_SESSIONS_DEFAULT_CRON = "0 * * * * *";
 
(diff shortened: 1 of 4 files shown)


rwinch:
Thanks for the PR! I'm interested in why you find this change necessary? Generally speaking I don't think this would really need to change.

ajthom90:
Hi @rwinch, thanks for the feedback!!  As I described in Issue #616, I'm running into issues in environments with multiple application servers connected to multiple databases.  At the top of the minute, both servers are kicking off the `JdbcOperationsSessionRepository#cleanUpExpiredSessions()` method to clean up sessions on their databases.  When the databases try to delete the data on the other servers through replication, it runs into errors because the data it is expecting to be there is actually not there.  This change would allow a system property to be set on each server for the Cron expression, so, for example, one server could run the `cleanUpExpiredSessions()` method at the top of every minute, and the other server could easily be set to run `cleanUpExpiredSessions()` at the 30 second point of every minute.  Sorry, I'm not the best at explaining things sometimes, so I hope this makes sense!  Let me know if you have any other questions.

In my app, I've essentially set up the solution that @vpavic suggested in his comment on Issue #616 yesterday, but I was just adding a pull request in case others thought it may be useful!

Issue #362, opened 2016-02-11
HttpSessionStrategy getRequestedSessionId->getRequestedSessionIds


rwinch:
Previously `HttpSessionStrategy` contained a flaw in that it only allowed resolving a single session id. However, it is common for multiple session ids to be present in the request. For example, when using cookies the same cookie name may be present for multiple domains or multiple paths.

This commit changes `HttpSessionStrategy` from:

``` java
String getRequestedSessionId();
```

to

``` java
List<String> getRequestedSessionIds();
```

This means that implementations of HttpSessionStrategy should change from
something like:

``` java
public String getRequestedSessionId() {
   ...
   return result;
}
```

to

``` java
public List<String> getRequestedSessionIds() {
    ...
    return result == null ? 
        Collections.<String>emptyList() : Collections.singleton(result);
}
```

PR #622, opened 2016-09-08
Rename getRequestedSessionId -> getRequestedSessionIds

eddumelendez:
Currently the method signature at HttpSessionStrategy is
getRequestedSessionId. Now, has been renamed to getRequestedSessionIds
in order to store all the session ids present in the request.

Fixes gh-362


edeandrea:
Wouldn't this be a breaking change to a public API? I know I have my own custom implementation of HttpSessionStrategy and upgrading to a version of spring-session which contains this change would break my code as well as break anyone else's code that is either extending or using the public API.

eddumelendez:
@edeandrea yes, this will break your code. That's the reason why this is schedule for version 2.0 because it will be a major release.

edeandrea:
Got it - thanks!

vpavic:
@eddumelendez Thanks for the PR, we are addressing this in a bit different way as a part of #906.

PR #624, opened 2016-09-10
Improve JDBC integration tests

vpavic:
It would be useful to improve JDBC integration tests to cover multiple RDBMS vendors. This will be particularly useful once #613 is merged, since it introduces multiple configuration paths which results in different SQL statement used for some vendors.

This PR parameterizes `JdbcOperationsSessionRepositoryITests` to run on all embedded database types supported by Spring's `EmbeddedDatabaseBuilder`.


rwinch:
Thanks for the PR! This is merged via 34948d64518843dbc51432f21af7c6f0a3ac3f2d

PR #638, opened 2016-09-22
Remove code repeat inside RedisSession and use this in methods call

marcoblos:
Maintain standard using `this.` prefix inside the RedisSession to methods call.

Generalizing `this.delta.put` and `this.flushImmediateIfNecessary` methods to keep an unique point of maintenance.

Thanks in advance.

### DIFF
--- a/spring-session/src/main/java/org/springframework/session/data/redis/RedisOperationsSessionRepository.java
+++ b/spring-session/src/main/java/org/springframework/session/data/redis/RedisOperationsSessionRepository.java
@@ -16,6 +16,7 @@
 
 package org.springframework.session.data.redis;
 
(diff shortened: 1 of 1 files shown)


rwinch:
Thanks for the PR @marcoblos! This is now merged into master via 4c2581d432bdbcf99f003042c62afa6868ba2884 Please note that we do not want to `RedisSession` to public because this is an implementation detail. Users can use the `ExpiringSession` interface instead. For this reason, I omitted the changes in the last commit.

marcoblos:
@rwinch Thanks for your explanation about `RedisSession`. This project is very very useful. Congratulations about your job.

Issue #508, opened 2016-04-29
Sequential consistency problem?


TyronZerafa:
SessionRepositoryFilter is committing the session of a request twice:
1) first commit when the response is flushed (i.e., written to the client) through the `onResponseCommitted()` method in the `SessionRepositoryResponseWrapper`
2) second commit after the response is flushed in the final block of the `SessionRepositoryFilter` 

We encountered a use case that breaks sequential consistency when the **system is under heavy load**.
In our use case, we are performing two sequential requests; (1) first request reads a session attribute and then (2) a second request updates this same session attribute. We noticed that, sometimes, the second commit of the first request (after flushing the response) overwrites the commits performed by the second request (which was fired right after the first request returned). 

Is there any reason why sessions are committed twice to the repository?

rwinch:
Thank you for the question! First what version of Spring Session and SessionRepository implementation are you using?

TyronZerafa:
I'm using version 1.1.1.RELEASE of Spring-session.

I managed to replicate this issue both with the MapSessionRepository (i.e., with the distributed map of the Hazelcast-based repository) and the MongoSessionRepository whose implementation can be found [here](https://github.com/tmurakam/spring-session-mongodb/blob/master/spring-session-mongodb/src/main/java/org/tmurakam/spring/session/data/mongodb/MongoSessionRepository.java).

Hazelcast version 3.6.2

rwinch:
We save at response committed because we need to ensure that as soon as the client has a reference to the session id (which can happen just after the response is committed). We save after the filter is done executing to ensure that if someone has made a change since the response was committed that the changes are persisted.

Some repository implementations are "smarter" than others. For example, the JDBC and Redis implementations both keep track of what has changed and only save the changes. This means that if only a single session attribute changes only that attribute is written. This also means if the save method is invoked more than once and nothing has changed nothing happens.

We should likely add similar enhancements for Mogno and the Map implementation. One down side is that with implementations like the Map implementation we can only write "all of the session" or nothing. This means if anything changes we have to write everything.

I created #525 and #526 to address this

rwinch:
Closing since this should be taken care of via the referenced issues above. Please reopen if you disagree

PR #641, opened 2016-09-26
Optimize save operation in HazelcastSessionRepository (#516)

### LINKED ISSUE #516: Session persistance optimization (Hazelcast as storage)

stojsavljevic:
This resolves #516:
- improves saving of sessions to only execute save operation if something has been changed.
- introduces configurable flush mode that specifies when to write to the backing Hazelcast instance. It can be 'on save' (default) or 'immediate'.

### DIFF
--- a/spring-session/src/main/java/org/springframework/session/hazelcast/HazelcastSessionRepository.java
+++ b/spring-session/src/main/java/org/springframework/session/hazelcast/HazelcastSessionRepository.java
@@ -20,6 +20,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
 import javax.annotation.PostConstruct;
@@ -36,8 +37,10 @@ import org.apache.commons.logging.LogFactory;
 
 import org.springframework.context.ApplicationEvent;
 import org.springframework.context.ApplicationEventPublisher;
+import org.springframework.session.ExpiringSession;
 import org.springframework.session.FindByIndexNameSessionRepository;
 import org.springframework.session.MapSession;
+import org.springframework.session.Session;
 import org.springframework.session.events.AbstractSessionEvent;
 import org.springframework.session.events.SessionCreatedEvent;
 import org.springframework.session.events.SessionDeletedEvent;
@@ -100,10 +103,11 @@ import org.springframework.util.Assert;
  * @author Vedran Pavic
  * @author Tommy Ludwig
  * @author Mark Anderson
+ * @author Aleksandar Stojsavljevic
  * @since 1.3.0
  */
 public class HazelcastSessionRepository implements
-		FindByIndexNameSessionRepository<MapSession>,
+		FindByIndexNameSessionRepository<HazelcastSessionRepository.HazelcastSession>,
 		EntryAddedListener<String, MapSession>,
 		EntryEvictedListener<String, MapSession>,
 		EntryRemovedListener<String, MapSession> {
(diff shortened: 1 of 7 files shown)


pivotal-issuemaster:
@stojsavljevic Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-session&pullRequestId=641)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-session&pullRequestId=641) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/faq) for frequently asked questions.

pivotal-issuemaster:
@stojsavljevic Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-session&pullRequestId=641)!

vpavic:
LGTM :+1:
Great work @stojsavljevic!

spring-session/src/main/java/org/springframework/session/hazelcast/HazelcastSessionRepository.java
```diff
@@ -242,4 +261,106 @@ public void entryRemoved(EntryEvent<String, MapSession> event) {
...
+		}
+
+		/**
+		 * Creates a new instance from the provided {@link MapSession}.
+		 *
+		 * @param cached the {@link MapSession} that represents the persisted session that was
+		 * retrieved. Cannot be null.
+		 */
+		HazelcastSession(MapSession cached) {
+			Assert.notNull("MapSession cannot be null");
```
shakuzen: Should be:
```java
Assert.notNull(cached, "MapSession cannot be null");
```

stojsavljevic: Thanks for catching this. Going to fix it ASAP..

stojsavljevic: It's fixed. Thanks again.

stojsavljevic:
@shakuzen 
This PR might slightly improve performance. But the main motivation is to avoid race conditions described in #508.

Another thing - I introduced new attribute in `EnableHazelcastHttpSession` - `hazelcastFlushMode`. I named it like that to be in line with `redisFlushMode` in `EnableRedisHttpSession`.
But I would like to call it `flushMode` to avoid noise. What do you think?

vpavic:
Thanks for the contribution @stojsavljevic! This is now merged to the `master` in 6a78101 with a minor polish in 4a06b38.

shakuzen:
@vpavic Will there be a mention in the release notes about this (particularly as it changes the default behavior)?

vpavic:
@shakuzen With this PR being a part of an effort started in #544 to bring the first-class support for Hazelcast, I'm sure things will get a proper mention in the `1.3.0` release notes.

@stojsavljevic One thing I forgot to comment on while closing the PR:

> Another thing - I introduced new attribute in EnableHazelcastHttpSession - hazelcastFlushMode. I named it like that to be in line with redisFlushMode in EnableRedisHttpSession.
But I would like to call it flushMode to avoid noise. What do you think?

@rwinch and I have discussed this and while we like the simple `flushMode` name, the `hazelcastFlushMode` was kept in order to maintain consistency with the existing naming of `redisFlushMode`.

PR #645, opened 2016-10-04
Look at all cookies instead just the first

eddumelendez:
Previous to this commit, just the first cookie is considered. Now, all
cookies are considered to look for SESSION cookie.

See gh-275

Initial draft

### DIFF
--- a/spring-session/src/main/java/org/springframework/session/web/http/CookieHttpSessionStrategy.java
+++ b/spring-session/src/main/java/org/springframework/session/web/http/CookieHttpSessionStrategy.java
@@ -136,6 +136,7 @@ import org.springframework.util.Assert;
  *
  *
  * @author Rob Winch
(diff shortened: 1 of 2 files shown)


oatesp:
I'm really interested in getting this merged as it's currently holding me back from using spring-session in my project. Is anyone able to take a look at this and consider if it is mergeable?

vpavic:
@oatesp We are currently working on a broader refactoring of `HttpSessionStrategy` that will, among other things, address  #275. See #906 to track the progress of that effort, we expect it to be a part of the upcoming `2.0.0.RC1` milestone.

Also note that this PR by itself does not address the original issue since the resolved session id's are not actually validated against the session store.

vpavic:
@eddumelendez Thanks for the PR, we are addressing this in a bit different way as explained in the comment above.

PR #651, opened 2016-10-25
Provide Redis with Lettuce sample application with guide

mp911de:
Add sample application using Redis with the Lettuce client.


mp911de:
Will change all samples to use Lettuce instead of Jedis after discussing with @rwinch

PR #652, opened 2016-10-25
Update Samples and Guides to use Lettuce

mp911de:
Favor lettuce because of multiplexing and improved scalability. Using lettuce requires a fixed number of connections hence using lettuce improves application scalability.

---

Leaving out Spring Boot samples as lettuce is not yet supported by Spring Boot with auto-configuration.


rwinch:

Issue #653, opened 2016-10-25
Exception on Initialization - Error creating bean with name 'enableRedisKeyspaceNotificationsInitializer'


ghsbhatia:
spring-session-1.0.2.RELEASE

Steps to Reproduce:
1. Include the following dependency
   
   ```
   <dependency>
       <groupId>org.springframework.session</groupId>
       <artifactId>spring-session</artifactId>
   </dependency>
   ```
2. Start the application when Redis server is not running

Exception:

Error creating bean with name 'enableRedisKeyspaceNotificationsInitializer' defined in class path resource[org/springframework/session/data/redis/config/annotation/web/http/RedisHttpSessionConfiguration.class]: Invocation of init method failed; nested exception is org.springframework.data.redis.RedisConnectionFailureException: Cannot get Jedis connection; nested exception is redis.clients.jedis.exceptions.JedisConnectionException: Could not get a resource from the pool

Root Cause:

A connection is requested from RedisConnectionFactory. If Redis is not running, exception is thrown in the follwing code in RedisHttpSessionConfiguration.java  

```
static class EnableRedisKeyspaceNotificationsInitializer implements InitializingBean {
    private final RedisConnectionFactory connectionFactory;

    private ConfigureRedisAction configure;

    EnableRedisKeyspaceNotificationsInitializer(RedisConnectionFactory connectionFactory, ConfigureRedisAction configure) {
        this.connectionFactory = connectionFactory;
        this.configure = configure;
    }

    public void afterPropertiesSet() throws Exception {
        RedisConnection connection = connectionFactory.getConnection();
        configure.configure(connection);
    }
}
```

Enahancement:
1. Configure ConfigureRedisAction instance only when RedisConnection is available.
2. Change signature of ConfigureRedisAction.configure to accept RedisConnectionFactory rather than RedisConnection. That way, when ConfigureRedisAction.NO_OP is autowired, no attempt is made to create a connection that would be unutilized.

DaqingFeng:
Please check your application.properties config, The  host or password can not allow letter space

rwinch:
I think this makes sense. If anyone is interested in submitting a PR, please mention on here to avoid duplicating efforts.

vpavic:
This was resolved in #657.

PR #657, opened 2016-10-30
Skip Redis getConnection() if ConfigureRedisAction.NO_OP

panchenko:
Temporary errors with Redis (like SocketTimeoutException) should not prevent application startup, especially if nothing needs to be done with that connection. 

Issue: gh-653


rwinch:
Thanks for the PR! This is merged via 2aec28289e2f3034b5d7e8e2b7356a23ce3aacb2

PR #666, opened 2016-11-11
Clear warnings from spring-boot gradle plugin

eddumelendez:


rwinch:
Thanks for the PR @eddumelendez! This is now merged into master

Issue #189, opened 2015-04-16
Spring Security Remember Me Integration


rwinch:
https://jira.spring.io/browse/SEC-2939

coco1979ka:
How is the current state for this issue? I think your suggestion in https://jira.spring.io/browse/SEC-2939 will not work, as this is not a real remember me - it only works as long as the browser is opened. If you close and reopen it, the session cookie (similar to JSESSION cookie) is invalid. Is it possible to send the Session Cookie with a fixed expiration time?

ptahchiev:
:+1: on this one

rwinch:
Fixed via #667

PR #667, opened 2016-11-11
Add Spring Security's `RememberMeServices` implementation backed by Spring Session

vpavic:
This PR aims to address #189.

This is still a WIP - although `SpringSessionRememberMeServices` itself is implemented, it will work only if `CookieSerializer` is configured according to `validitySeconds` from `SpringSessionRememberMeServices` which isn't ideal.

IMO it should be possible to have behavior that is similar to one of Google Account's `SID` cookie - if `Stay signed in` option is selected, session cookie is persistent meaning it has expiration specified, otherwise it's not persistent. So `SpringSessionRememberMeServices` should be linked with `CookieHttpSessionStrategy` (perhaps extend it?) to make sure cookie is serialized in a different manner depending on whether remember-me was requested or not. WDYT @rwinch?

### DIFF
--- a/spring-session/build.gradle
+++ b/spring-session/build.gradle
@@ -25,7 +25,8 @@ dependencies {
 			"org.springframework:spring-web:$springVersion",
 			"org.springframework:spring-messaging:$springVersion",
 			"org.springframework:spring-websocket:$springVersion",
-			"org.springframework.security:spring-security-core:$springSecurityVersion"
+			"org.springframework.security:spring-security-core:$springSecurityVersion",
+			"org.springframework.security:spring-security-web:$springSecurityVersion"
 	provided "javax.servlet:javax.servlet-api:$servletApiVersion"
 	integrationTestCompile "redis.clients:jedis:$jedisVersion",
 			"org.apache.commons:commons-pool2:2.2",
--- a/spring-session/src/main/java/org/springframework/session/config/annotation/web/http/SpringHttpSessionConfiguration.java
+++ b/spring-session/src/main/java/org/springframework/session/config/annotation/web/http/SpringHttpSessionConfiguration.java
@@ -19,22 +19,29 @@ package org.springframework.session.config.annotation.web.http;
 import java.util.ArrayList;
 import java.util.List;
 
+import javax.annotation.PostConstruct;
 import javax.servlet.ServletContext;
 import javax.servlet.http.HttpSessionListener;
 
+import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.ApplicationContextAware;
 import org.springframework.context.annotation.Bean;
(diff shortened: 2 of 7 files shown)


ptahchiev:
What are the chances - I just wanted to do the same and googled to see if someone has done it already :)

rwinch:
I think I would prefer to keep the two objects decoupled, but provide an easy way to configure them. One way this could be done by determining if `SpringSessionRememberMeServices` is a Bean to set `CookieSerializer` expiration property.

Another issue we should consider is how does this behave with Spring Security's session fixation protection. In older servlet environments the session is invalidated and then recreated at log in. I cannot recall off the top of my head if this happens before or after the remember me logic is invoked.

vpavic:
@rwinch Thanks for the feedback, I'll look into it.

> One way this could be done by determining if SpringSessionRememberMeServices is a Bean to set CookieSerializer expiration property.

But this doesn't (unless I'm missing something) address the situation described in my previous comment:

> IMO it should be possible to have behavior that is similar to one of Google Account's `SID` cookie - if `Stay signed in` option is selected, session cookie is persistent meaning it has expiration specified, otherwise it's not persistent.

We need to somehow serialize the cookie with different max age depending on whether remember-me was requested or not. Having two `CookieSerializer`s doesn't look appealing to me, maybe `CookieSerializer` interface could be extended with `writeCookieValue` method that allows specifying the `cookieMaxAge` parameter?

rwinch:
@vpavic 

I'm not sure why this wouldn't help. If `SpringSessionRememberMeServices` was a Bean, the configuration would automatically ensure that the `CookieSerializer` was setup properly.

vpavic:
@rwinch I've updated the PR. `SpringSessionRememberMeServices` now sets request attribute upon successful remember-me login. `DefaultCookieSerializer` can now be configured to look for the presence of remember-me request attribute and apply the appropriate cookie max age. I went with request attribute (vs request param as we originally discussed) due to the fact its more reliable way to indicate remember-me login - the request param might be present without remember-me being involved at all, which is sort of a risk.

The automatic configuration of `DefaultCookieSerializer` was somewhat tricky due to `SpringSessionRememberMeServices` relying on `spring-security-web` dependency which is optional.

rwinch:
Thanks for the PR! This is merged into master. I provided some polish in 49e3a1c7cde44ab1c6bffc85a648fd067ac5a55c

jbollacke:
Isn't the documentation missing out on the CookeSerializer configuration?

I have to configure rememberMeRequestAttribute on DefaultCookieSerializer (see below) for remember-me to work.
```kotlin
@Bean
fun cookieSerializer() = DefaultCookieSerializer().apply {
  setRememberMeRequestAttribute(SpringSessionRememberMeServices.REMEMBER_ME_LOGIN_ATTR)
}
````

vpavic:
I think you are right @jbollacke, thanks for the feedback. Can you open a new issue to address this?

jbollacke:
Sure, #1157.

PR #668, opened 2016-11-11
Upgrade Gradle to 3.1

vpavic:


artembilan:
@vpavic , 

thank you for contribution as usual! :smile: 
I wonder what do you think about Spring IO Platform compatibility https://github.com/spring-gradle-plugins/dependency-management-plugin/issues/115.

I'm also interested in the Gradle 3.1 upgrade but that issue is really stopping me, although I don't see any problems in one my small project which is a part of Spring IO Platform and really build SUCCESSFULy with the `springIoCheck`  task when I add `classpath 'io.spring.gradle:dependency-management-plugin:0.6.1.RELEASE'` as well.

WDYT?

vpavic:
Hey @artembilan, you're welcome! :smile:

I've been following the situation with Gradle 3 and Dependency management plugin as we (at my company) use Gradle exclusively so it's a topic of our interest. We have two medium to large projects ongoing that are on Gradle 3 (3.1 ATM to be exact) and we haven't faced any issues (both are Spring IO Platform and Boot based). Spring Security has also moved to Gradle 3 (spring-projects/spring-security#4052) some time ago. So IMO the best approach would be to attempt the Gradle upgrade and see where that gets you, since chances are good (from my experience at least) that you won't run into any issue.

Upgrade of Dependency management plugin to 0.6.x is required due to spring-gradle-plugins/dependency-management-plugin#87 since Spring IO plugin is still based on 0.5.x, I assume you encountered that one in your project as well.

PR #675, opened 2016-11-19
Remove deprecations

vpavic:
This PR removes `SessionMessageListener` and `CookieHttpSessionStrategy#setCookieName` which both were deprecated since `1.1.0`.


rwinch:
Thanks for the PR! We need to remove these in a non-passive release, so scheduling for 2.0 M1

vpavic:
Updated the PR to rebase to current master and remove `SessionEntryListener` which was deprecated in `1.3.0`.

PR #695, opened 2016-12-20
Allow easier customization of cookie max age logic

vpavic:
This PR refactors `DefaultCookieSerializer`'s cookie max age logic to allow users to easier apply their custom cookie max age. This change allows to customize max age either via inheritance or composition.

The default max age values are now contained in `CookieSerializer.CookieValue`.

@rwinch Please review the proposed change and let me know if this approach is OK with you.

### DIFF
--- a/spring-session-core/src/main/java/org/springframework/session/web/http/CookieSerializer.java
+++ b/spring-session-core/src/main/java/org/springframework/session/web/http/CookieSerializer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2016 the original author or authors.
+ * Copyright 2014-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


spring-session-core/src/main/java/org/springframework/session/web/http/CookieSerializer.java
```diff
@@ -105,5 +111,24 @@ public HttpServletResponse getResponse() {
...
 			return this.cookieValue;
 		}
+
+		/**
+		 * Get the cookie max age. The default is -1 which signals to delete the cookie
+		 * when the browser is closed, or 0 if cookie value is empty.
+		 * @return the cookie max age
+		 */
+		public int getCookieMaxAge() {
+			return "".equals(this.cookieValue) ? 0 : this.cookieMaxAge;
```
rwinch: I don't like this because it means we cannot have an empty String value. Could the code that sets the cookie value to "" set the max age too?

vpavic:
@rwinch I've updated the PR to address your comments.

rwinch:
Thanks for the PR! This is merged into master via 8e3371aed9e6a92b79298a993f0974096191bb1a

PR #713, opened 2017-01-15
Add support for configuring default `CookieSerializer` using `SessionCookieConfig`

### LINKED ISSUE #87: Configure Session Cookie using SessionCookieConfig
The configuration for the Session Cookie (i.e. SessionCookieConfig) should impact Spring Session's settings.

vpavic:
This resolves #87.

The proposed solution doesn't map two attributes from `SessionCookieConfig`:

- `secure`: `DefaultCookieSerializer#useSecureCookie` defaults to `null` and relies on `ServletRequest#isSecure` which means mapping a value from `SessionCookieConfig` would override our default (since `secure` is a primitive `boolean` over there and typically defaults to `false`)
- `httpOnly`: `DefaultCookieSerializer#useHttpOnlyCookie` defaults to `true` in Servlet 3.0+ environments which means our default would be overridden in `httpOnly` typically defaults to `false`

~~Other than that integration tests in Boot based samples needed to be updated with explicit `MockServletContext` due to `MockSessionCookieConfig#maxAge` defaulting to `0` instead of `-1`. I've opened [SPR-15142](https://jira.spring.io/browse/SPR-15142) to address that.~~

### DIFF
--- a/spring-session/src/main/java/org/springframework/session/config/annotation/web/http/SpringHttpSessionConfiguration.java
+++ b/spring-session/src/main/java/org/springframework/session/config/annotation/web/http/SpringHttpSessionConfiguration.java
@@ -21,7 +21,11 @@ import java.util.List;
 
 import javax.annotation.PostConstruct;
 import javax.servlet.ServletContext;
+import javax.servlet.SessionCookieConfig;
 import javax.servlet.http.HttpSessionListener;
+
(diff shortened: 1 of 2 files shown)


vpavic:
PR updated due to [SPR-15142](https://jira.spring.io/browse/SPR-15142) being fixed in the `master` via spring-projects/spring-framework#1298.

Issue #735, opened 2017-02-16
Remove Base64


rwinch:
We won't need the Base64 class once we are on Java 8

PR #737, opened 2017-02-16
Remove Base64

eddumelendez:
Since verion 2.0.0 will use java 8 as a base, use the Base64 class
provided by java instead.

Fixes gh-735


vpavic:
Thanks for the PR @eddumelendez, this is merged to `master` in 6327d36.

PR #791, opened 2017-05-25
Update minimum requirements

vpavic:
Our [minimum requirements section](http://docs.spring.io/spring-session/docs/2.0.0.M1/reference/html5/#minimum-requirements) is outdated for `2.0.x` - @rwinch please confirm the updates proposed here are correct.

### DIFF
--- a/docs/src/docs/asciidoc/index.adoc
+++ b/docs/src/docs/asciidoc/index.adoc
@@ -979,11 +979,11 @@ Spring Session is Open Source software released under the http://www.apache.org/
 
 The minimum requirements for Spring Session are:
 
-* Java 5+
-* If you are running in a Servlet Container (not required), Servlet 2.5+
-* If you are using other Spring libraries (not required), the minimum required version is Spring 3.2.14.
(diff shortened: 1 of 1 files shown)


vpavic:
Thanks for the feedback @tsachev. Actually we have [a sample app](https://github.com/spring-projects/spring-session/tree/master/samples/misc/hazelcast) that demonstrates the usage without the Spring Core. `org.springframework.util.Assert` is indeed used within the Spring Session, but to my knowledge in parts of the framework that need Spring Core anyway. Did you encounter any particular problem?

tsachev:
well I did not since I am using spring, too but just noted that the repo filter is using spring core thought this doc states that spring is not required.
or it is when using sevlwt containers?

vpavic:
@tsachev Are you referring to `SessionRepositoryFilter`? I'm not aware of it using Spring Core (previously linked [sample app](https://github.com/spring-projects/spring-session/tree/master/samples/misc/hazelcast) wouldn't work if it did). Can you point to the specific usage we're talking about here?

PR #798, opened 2017-06-07
Use TestContainers for integration tests

### LINKED ISSUE #795: Use TestContainers for integration tests
We should update our integration tests and samples to use [TestContainers](https://www.testcontainers.org/).

This would bring the following benefits:
- make Docker a single external dependency for project build (currently we require Redis)
- allow us to improve JDBC support integration tests
- provide better isolation for integration tests

vpavic:
This resolves #795.

~~Note: this is work in progress since WAR samples are still looking at `localhost:6379`.~~

Regarding integration tests, I've added JDBC tests for PostgreSQL, MySQL and MariaDB. We should still decide which exact versions we want to test against, especially with MySQL/MariaDB. The same applies to Redis.

As far as those WAR samples are concerned, @rwinch do you have some idea how to handle this having in mind current Tomcat based integration tests?


rwinch:
We can default the port to 6379 and override it using a system property with the correct port. In our configuration we can use standard PropertyPlaceholderConfigurer

vpavic:
@rwinch I've added a commit that demonstrates the approach taken for integration testing of WAR based samples.

This fails for reasons we've discussed the other day - I've opened testcontainers/testcontainers-java#369.

vpavic:
I've updated the PR to pick up recent changes to `Session` API and module structure.

The previous problem is still present though. There is however some feedback on testcontainers/testcontainers-java#369.

bsideup:
Hi folks,

@testcontainers is here :)

I think I know why application server doesn't stop, I'll debug it on TestContainers' side. 

Meanwhile could you please provide more details on why are you trying to run testcontainers' containers as WAR-ed app? :)

vpavic:
Hi @bsideup - thanks for taking a look at this!

Regarding the motivation, our build currently depends on Redis service being available, which is used for integration testing of both our Redis-backed components and sample applications that use those components. We'd like to standardize and improve our integration tests using TestContainers, which would also make Docker the only external service our build depends on.

Integration testing of Redis-backed components and sample apps that are Spring Boot based is straightforward using JUnit `@ClassRule`, however some sample apps are packaged as WARs and then integration tested on Tomcat using gretty plugins. Our idea with those sample apps was to introduce an application profile which would be used in those integration tests, and which would sort of embed Redis instance (via TestContainers) using approach demonstrated in testcontainers/testcontainers-java#369.

vpavic:
Thanks to @bsideup we have a workaround and are no longer blocked by testcontainers/testcontainers-java#369. I'll revisit the PR to address the integration testing of WAR based samples during this week.

vpavic:
This should be ready now with the latest updates to the PR.

Integration tests for `RedisOperationsSessionRepository` (as well as other Redis integration tests) are based on Redis `3.2.9`.

Integration tests for `JdbcOperationsSessionRepository` on top of existing H2, HSQLDB and Derby based now also include:

- MariaDB 5.x & 10.x
- MySQL 5.x
- PostgreSQL 9.x

TestContainers also provide Oracle XE module, but this is problematic to include due to well know situation with availability of Oracle JDBC driver in public repos. There is also SQL Server module in the works, so we might include that once the release is out.

Regarding the tricky part of this PR, the integration testing of WAR based samples using Gretty, I've opted to introduce a separate Redis configuration which is activated with `embedded-redis` Spring profile. This way the original Redis/Session configuration classes were mostly untouched which is important since the guides include code snippets from those classes.

Let me know what you think @rwinch.

samples/boot/findbyusername/src/integration-test/java/sample/FindByUsernameTests.java
```diff
@@ -72,4 +84,18 @@ public void login() {
...
 
+	static class Initializer
+			implements ApplicationContextInitializer<ConfigurableApplicationContext> {
+
+		@Override
+		public void initialize(
+				ConfigurableApplicationContext configurableApplicationContext) {
+			TestPropertyValues
+					.of("spring.redis.host=" + redisContainer.getContainerIpAddress(),
+							"spring.redis.port=" + redisContainer.getMappedPort(6379))
```
bsideup: you can also use `getFirstMappedPort()` since 1.4 :)

vpavic: Thanks for the tip!

rwinch:
Can you please update the README

vpavic:
> Can you please update the README and guides?

Done. I've added a reference to Docker as an alternative to Redis installation.

rwinch:
Thanks for the PR! This is now [merged](https://github.com/spring-projects/spring-session/commits/63b67a501d0b5f7ce71d3bddc9f6ca247ce02355)

PR #804, opened 2017-06-19
Replace put of IMap [Hazelcast]

dhawani:
Based on [article](https://blog.hazelcast.com/performance-top-5-1-map-put-vs-map-set), replace put with set to avoid serialisation and de-serialisation of entires.

### DIFF
--- a/spring-session-hazelcast/src/integration-test/java/org/springframework/session/hazelcast/config/annotation/web/http/EnableHazelcastHttpSessionEventsTests.java
+++ b/spring-session-hazelcast/src/integration-test/java/org/springframework/session/hazelcast/config/annotation/web/http/EnableHazelcastHttpSessionEventsTests.java
@@ -120,6 +120,8 @@ public class EnableHazelcastHttpSessionEventsTests<S extends Session> {
 		assertThat(this.registry.receivedEvent(sessionToSave.getId())).isTrue();
 		assertThat(this.registry.<SessionExpiredEvent>getEvent(sessionToSave.getId()))
 				.isInstanceOf(SessionExpiredEvent.class);
+		assertThat(this.registry.<SessionExpiredEvent>getEvent(sessionToSave.getId()).<Session>getSession())
+				.isNotNull();
 
 		assertThat(this.repository.<Session>findById(sessionToSave.getId())).isNull();
 	}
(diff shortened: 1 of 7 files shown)


vpavic:
Thanks for the PR @dhawani.

Personally I'd keep using the existing `IMap` operation for consistency. As you've discovered yourself while trying to also replace the use of `IMap#remove` with `IMap#delete`, these operations are breaking some of the listener contracts as stated in the javadoc:

> This method breaks the contract of EntryListener. When an entry is updated by set(), it fires an EntryEvent with a null oldValue.

Even though `IMap#set` works within `HazelcastSessionRepository` I can see some users relying on `EntryListener` contract.

Do you have any opinion on this @rwinch?

vpavic:
@dhawani We've discussed this internally a bit and we would make the use of `IMap`s `set` and `delete` operations optional for users. That is, `HazelcastSessionRepository` would have a flag that would allow user to switch to `set`/`delete` operations (with default being existing ones, `put`/`remove`). Note that when using `set`/`delete` session event publishing would be disable.

Are you willing to rework you PR to do that?

dhawani:
@vpavic I have made this change based on the flag. 

Instead of disabling the event entirely, i am passing what all is available. In case of create event there is no change at all. In case of destroy and expire, the old value isn't available. But still we are able to pass the event. Then it becomes user's choice to subscribe to it. 

In the application where i am using this, I am only interested for create.  

Take a look at it and let me know if there are any more changes required.

vpavic:
Thanks for the update @dhawani - will provide some feedback after I take a closer look.

dhawani:
@vpavic Any update on this ?

vpavic:
@dhawani This will be addressed as a part of one of the upcoming 2.0 milestone releases. Most of our focus currently is with reactive support and some breaking changes we plan to introduce as a part of 2.0.

vpavic:
I'm putting this on hold as we're looking to use Hazelcast's `EntryProcessor` for write operations - see #850. We can revisit once that's done.

vpavic:
Thanks for the PR @dhawani, however we've addressed this as a part of #850 in f5912da. Check out [this comment](https://github.com/spring-projects/spring-session/issues/850#issuecomment-346894660) for more details.

Note that we've kept `IMap#remove` for session removal due to event considerations.

PR #821, opened 2017-07-05
Simplify Hazelcast sample

vpavic:
The basic Hazelcast sample (Java EE and `MapSessionRepository` based) is perhaps too complex for a sample app since it provides a custom serialization implementation. Even the logic to set the port explicitly could be removed since Hazelcast itself will by default increment port to find available one in 5701 is already used.

This PR simplifies the mentioned sample.


rwinch:
Thanks for the PR! This is merged into master

PR #825, opened 2017-07-10
Add Spring Session BOM module

vpavic:
This PR adds Spring Session Maven Bill of Materials (BOM) module.

The approach taken in this initial version is borrowed from Spring Security's BOM module, which means BOM currently provides dependency management only for modules provided by this repo. We should extend this to support external ones, like `spring-session-data-mongodb`. I guess we can address support for that via spring-gradle-plugins/spring-build-conventions#1.

Also we should consider introducing BOM versioning theme.

### DIFF
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -1,0 +1,5 @@
+apply plugin: 'groovy'
+
+dependencies {
+	compile gradleApi()
+}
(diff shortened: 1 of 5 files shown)


vpavic:
Closing in favor of creating a separate project for BOM. Prototype is available is in [this repo](https://github.com/vpavic/spring-session-bom).

PR #832, opened 2017-07-14
Provide `SessionRepository` implementations discovery

vpavic:
This PR provides discovery of `SessionRepository` implementations using `spring.factories` to help improve Spring Boot auto-configuration for Spring Session, as suggested in spring-projects/spring-boot#9683.

Once `ReactorSessionRepository` implementations become available, we should add a separate key for that as well.

### DIFF
--- a/spring-session-data-redis/src/main/resources/META-INF/spring.factories
+++ b/spring-session-data-redis/src/main/resources/META-INF/spring.factories
@@ -1,0 +1,2 @@
+org.springframework.session.config.annotation.web.http.EnableSpringHttpSession=\
+org.springframework.session.data.redis.config.annotation.web.http.RedisHttpSessionConfiguration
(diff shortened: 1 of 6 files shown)


rwinch:
I don't think that `spring.factories` for `SpringSession` is fitting here. Typically `spring.factories` is used for constructing a new instance of a class that can be used. However, we cannot create new instances of `SessionRepository` and expect to use them using `SpringFactoriesLoader`.

vpavic:
My idea was that this could be modeled somewhat similar to Boot's `org.springframework.boot.autoconfigure.EnableAutoConfiguration` key from `spring.factories` - since that entry also isn't directly used to construct a new instance of a class.

snicoll:
>  to help improve Spring Boot auto-configuration for Spring Session

@vpavic I had a look to the PR and I don't understand how Spring Boot could benefit from this.

Also, if you use different keys in `spring.factories`, that breaks the intend of it IMO: you should have one key only (that might be poor feedback on my end as I don't understand the intend yet).

snicoll:
@vpavic I am so sorry. I just re-read my own comment :(

> If Spring Session provides a way to discover the available implementations (as Spring Data does with spring.factories ) 

Now we are on the same page, those entries in `spring.factories` should have a unique key.

vpavic:
With spring-projects/spring-boot#9683 resolved we don't need this anymore.

PR #855, opened 2017-08-17
Add WebFlux sample

gregturn:


rwinch:
Other than the obvious checkstyle problems (I don't mind cleaning these up), there were a few problems with this PR.

The PR contained a fix for two different issues and the sample based off my branch. The fix to the commit changed the sample to explicitly use the `WebSessionManager` which is not desirable.

I closed this in favor of #856 and #857 which I have pushed fixes for.

Speaking to @gregturn offline he is going to submit a separate PR for the final issue.

PR #860, opened 2017-08-22
Use Spring null assertions.

gregturn:


rwinch:
At it's core (including `MapReactorSessionRepository`) Spring Session should have no dependencies on Spring. Therefore I am closing this as wont' fix

PR #866, opened 2017-09-07
Remove `MapSessionRepository` default constructor

vpavic:
This PR removes the default `MapSessionRepository` so that the users are required to explicitly supply the `Map` used to store the sessions.

The current default does not handle purging of expired sessions which is an important aspect of the `SessionRepository` implementation.


rwinch:
Thanks this is now merged into master. 

NOTE: I removed passivity since we aren't concerned with passivity for something that has never been in a release

PR #868, opened 2017-09-07
Configure WebSessionManager's WebSessionIdResolver by bean definition

gregturn:
Allow a WebSessionIdResolver registered as a Spring bean to be wired into the WebSessionManager.

### DIFF
--- a/spring-session-core/src/main/java/org/springframework/session/SpringWebSessionConfiguration.java
+++ b/spring-session-core/src/main/java/org/springframework/session/SpringWebSessionConfiguration.java
@@ -15,11 +15,13 @@
  */
 package org.springframework.session;
 
+import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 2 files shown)


gregturn:
I coded this with:

```
@Autowired(required = false)
WebSessionIdResolver webSessionIdResolver;
```
...but I can also do essentially the same thing by injecting it into the `webSessionManager()` as `webSessionManager(@Autowired(required = false) WebSessionIdResolver webSessionIdResolver)`, but I thought that looked a bit long.

If you prefer the alternate, I can switch it. over.

rwinch:
Thanks for the PR! I pushed this to master with a bit of polish to fix checkstyle 5c294ae1d208ffba05cb27a4486b864278fc980d I also fixed the tests to ensure that the close the `ApplicationContext` 52ea98b4ceb631979a24aa10a2a911e921d1377e

gregturn:
:+1

PR #872, opened 2017-09-11
Optimize JDBC session cleanup SQL statement

### LINKED ISSUE #847: Improve cleanUpExpiredSessions query
Hello,

I was profiling our production database, which is MySQL, and found that expiration query is not optimized.  
```
DELETE FROM %TABLE_NAME% WHERE MAX_INACTIVE_INTERVAL < (? - LAST_ACCESS_TIME) / 1000;
```
This query is not using any index. However, if you change query to something like this you will get a point;
```
DELETE FROM %TABLE_NAME% WHERE (? - (MAX_INACTIVE_INTERVAL * 1000)) >  LAST_ACCESS_TIME
```
We know that MAX_INACTIVE_INTERVAL is constant in our table, so this expression can be evaluated before query execution
```
(? - (MAX_INACTIVE_INTERVAL * 1000)) 
```
Now, final query will be one of these:
```
DELETE FROM %TABLE_NAME% WHERE ? >  LAST_ACCESS_TIME

or 

DELETE FROM %TABLE_NAME% WHERE LAST_ACCESS_TIME < ?
```
Simple and index friendly. 

Is there a chance to review this?

Thank you.

vpavic:
This PR improves session cleanup handling in  `JdbcOperationsSessionRepository#cleanUpExpiredSessions` by optimizing the used SQL statement. This is done by calculating the session expiry time when persisting the session, which in turn allows the cleanup SQL statement to be more index-friendly.

This resolves #847.


rwinch:

cemo:
It seems good for me as well. Thank you guys.  <3

PR #887, opened 2017-10-03
Improve Redis configuration

### LINKED ISSUE #864: Improve Redis configuration
Our current Redis configuration i.e. `RedisHttpSessionConfiguration` causes problems in some Boot apps due to the fact it registers a `RedisTemplate<Object, Object>` bean named `sessionRedisTemplate`. This bean doesn't switch off `RedisTemplate<Object, Object>` auto-configured by Boot since that one is conditional on `redisTemplate` bean name. So the users that want to inject a `RedisTemplate<Object, Object>` into their own component are unable to do so without qualifier. See spring-projects/spring-boot#9995 for more background.

There was a similar situation with JDBC however this was resolved by adding `@Primary` on Boot's auto-configured `JdbcTemplate` bean. There's also a proposal to improve our JDBC config in #863.
There are some differences in JDBC vs Redis scenarios though:

- Boot's auto-configured `JdbcTemplate` is conditional on bean whereas `RedisTemplate<Object, Object>` is conditional on bean name
- Spring Session does some custom configuration on the `RedisTemplate` and users need to be able to obtain this instance if they are to access the session store outside what's offered by `RedisOperationsSessionRepository` whereas with `JdbcTemplate` there's no special config hence no need to obtain Spring Session specific instance as well.

So our configuration improvements should ensure that:

- it isn't possible to inject the wrong `RedisTemplate<Object, Object>` instance into their components in a Boot app (specifically Spring Session specific instance instead of general purpose one provided by Boot)
- the users can easily obtain Spring Session specific `RedisTemplate<Object, Object>` instance in order to be able to do their custom data store access
- provide a simple way to specify an alternative `RedisConnectionFactory` to be used by Spring Session, for instance in scenarios with multiple Redis instances (similar to what #863 does for JDBC)

There are a couple ideas floating around after a discussion with @rwinch:

- provide an annotation meta-annotated with `@Value` that uses SpEL to access a bean that does the lookup for Spring Session specific `RedisTemplate<Object, Object>` instance
- we could also provide a strategy interface that can be used to retrieve Spring Session specific `RedisTemplate<Object, Object>` instance, and have `RedisOperationsSessionRepository` implement it

/cc @snicoll - this might be of your interest considering the relation with spring-projects/spring-boot#9995

vpavic:
This PR improves Redis configuration by introducing `@SpringSessionRedisConnectionFactory` qualifier for explicitly declaring a `RedisConnectionFactory` to be used by Spring Session. This is in particular useful in scenarios with multiple `RedisConnectionFactory` beans present in the application context.

Redis configuration is simplified and no longer registers a Spring Session specific `RedisOperations<Object,Object>` bean with the application context.

Users are however able to obtain `RedisOperations<Object,Object>` instance used by Spring Session using newly introduced `@SpringSessionRedisOperations` annotation.

This resolves #864.

### DIFF
--- a/samples/boot/redis-json/src/integration-test/java/sample/RedisSerializerTest.java
+++ b/samples/boot/redis-json/src/integration-test/java/sample/RedisSerializerTest.java
@@ -21,13 +21,13 @@ import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.testcontainers.containers.GenericContainer;
 
-import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.context.SpringBootTest;
 import org.springframework.boot.test.util.TestPropertyValues;
 import org.springframework.context.ApplicationContextInitializer;
 import org.springframework.context.ConfigurableApplicationContext;
 import org.springframework.data.redis.core.RedisTemplate;
 import org.springframework.data.redis.serializer.GenericJackson2JsonRedisSerializer;
+import org.springframework.session.data.redis.config.annotation.web.http.SpringSessionRedisOperations;
 import org.springframework.test.context.ContextConfiguration;
 import org.springframework.test.context.junit4.SpringRunner;
 
@@ -48,7 +48,7 @@ public class RedisSerializerTest {
 	public static GenericContainer redisContainer = new GenericContainer(DOCKER_IMAGE)
 			.withExposedPorts(6379);
 
-	@Autowired
+	@SpringSessionRedisOperations
 	private RedisTemplate<Object, Object> sessionRedisTemplate;
 
 	@Test
--- a/spring-session-data-redis/src/integration-test/java/org/springframework/session/data/redis/RedisOperationsSessionRepositoryITests.java
+++ b/spring-session-data-redis/src/integration-test/java/org/springframework/session/data/redis/RedisOperationsSessionRepositoryITests.java
@@ -38,6 +38,7 @@ import org.springframework.session.Session;
 import org.springframework.session.data.SessionEventRegistry;
 import org.springframework.session.data.redis.RedisOperationsSessionRepository.RedisSession;
 import org.springframework.session.data.redis.config.annotation.web.http.EnableRedisHttpSession;
+import org.springframework.session.data.redis.config.annotation.web.http.SpringSessionRedisOperations;
 import org.springframework.session.events.SessionCreatedEvent;
 import org.springframework.session.events.SessionDestroyedEvent;
 import org.springframework.test.context.ContextConfiguration;
@@ -61,7 +62,7 @@ public class RedisOperationsSessionRepositoryITests extends AbstractRedisITests 
 	@Autowired
 	private SessionEventRegistry registry;
 
-	@Autowired
+	@SpringSessionRedisOperations
 	private RedisOperations<Object, Object> redis;
 
 	private SecurityContext context;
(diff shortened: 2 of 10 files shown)


rwinch:
Thanks @vpavic! This is merged into master

rohitkrishna094:
Are there any docs or tutorials on how to use this annotation? Thanks

rohitkrishna094:
So my understanding is this. If you have two RedisConnectionFactories, things blow up. Hence we need this annotation to say to Spring Session what to use for session. 

Consider this example:  in my configuration class I have two beans of same type and now spring is confused as to what to use for session when `@EnableRedisHttpSession` is enabled. One solution suggested in other thread is to add `@Primary` on any one of these annotations. 

```java
// Spring will be confused by below code as there are two beans of same type

@Bean
RedisConnectionFactory cloudRedisConnectionFactory() {
    RedisStandaloneConfiguration redisStandaloneConfiguration = new RedisStandaloneConfiguration();
// ...do stuff
    return new JedisConnectionFactory(redisStandaloneConfiguration);
}

@Bean
@SpringSessionRedisConnectionFactory
RedisConnectionFactory localRedisConnectionFactory() {
    RedisStandaloneConfiguration redisStandaloneConfiguration = new RedisStandaloneConfiguration();
    redisStandaloneConfiguration.setHostName("localhost");
    redisStandaloneConfiguration.setPort(6379);
    return new JedisConnectionFactory(redisStandaloneConfiguration, JedisClientConfiguration.builder().build());
}
```

However I can also add `@SpringSessionRedisConnectionFactory` on any one of these instead of `@Primary` right? So I basically added `@SpringSessionRedisConnectionFactory` on the method named `cloudRedisConnectionFactory` above. However now I am getting another error called:

`Parameter 0 of method redisTemplate in org.springframework.boot.autoconfigure.data.redis.RedisAutoConfiguration required a single bean, but 2 were found:`

Looks like this PR did not fix everything? or am I misusing  `@SpringSessionRedisConnectionFactory` annotation?

thaingo:
I am having the same issue as @rohitkrishna094 mentioned.
Could you please advice @vpavic ? Thanks.

eleftherias:
@rohitkrishna094 @thaingo The class you are referring to (`RedisAutoConfiguration`) is a Spring Boot class.

The qualifier `@SpringSessionRedisConnectionFactory` will only differentiate between `RedisConnectionFactory` beans in the Spring Session configuration, for example `RedisHttpSessionConfiguration`.

If you are using the Spring Boot auto-configuration with multiple `RedisConnectionFactory` beans  you will still need to declare one of the beans as `@Primary`.

thaingo:
Thanks @eleftherias for your response.

I really want to use Spring boot auto-configuration with multiple `RedisConnectionFactory` beans: 1 for Spring data redis and 1 for Spring session redis.

Could you please advise how to properly defines these 2 beans? I am still struggling. Thanks.

vpavic:
@thaingo It sounds to me that your use case with 2 `RedisConnectionFactory` is that the _default_ one wouldn't be used for Spring Session but would for everything else (at least that's how I read your _1 for Spring data redis and 1 for Spring session redis_ comment).

If that's the case, you should annotate the _default_ one with `@Primary` and the one intended for use with Spring Session with `@SpringSessionRedisConnectionFactory`.

There's a test case in Spring Session that validates this approach:
https://github.com/spring-projects/spring-session/blob/5acb307a54c7a5347ffce8b4be7adb9bd3339aa5/spring-session-data-redis/src/test/java/org/springframework/session/data/redis/config/annotation/web/http/RedisHttpSessionConfigurationTests.java#L194-L208
https://github.com/spring-projects/spring-session/blob/5acb307a54c7a5347ffce8b4be7adb9bd3339aa5/spring-session-data-redis/src/test/java/org/springframework/session/data/redis/config/annotation/web/http/RedisHttpSessionConfigurationTests.java#L378-L394

thaingo:
It works. Thanks @vpavic and the team.

PR #906, opened 2017-10-25
Align `HttpSessionStrategy` with `WebSessionIdResolver`

### LINKED ISSUE #275: CookieHttpSessionStrategy should look at all cookies not just the first

vpavic:
This PR simplifies `HttpSessionStrategy` API by aligning it with Spring Framework's `WebSessionIdResolver`. As a part of this, support for managing multiple users' sessions has been removed.

This resolves #275 and #362.


rwinch:
Thanks @vpavic! This is merged into master

Issue #775, opened 2017-05-06
Question about PropertySourcesPlaceholderConfigurer Bean in JdbcHttpSession


sayan-chaliha:
Hey,

I'm looking at https://github.com/spring-projects/spring-session/blob/master/spring-session/src/main/java/org/springframework/session/jdbc/config/annotation/web/http/JdbcHttpSessionConfiguration.java#L192

This bean definition here is messing up placeholder resolution in my application. Is it really necessary?

vpavic:
Hi @sayan-chaliha - this was introduced while addressing #616 (see also #621 and 5ecf390).

Can you explain in more detail how is it messing up placeholder resolution in your app? Sample app would also be nice to have.

sayan-chaliha:
Hey @vpavic -- thanks for the pointers! I'll go through them.

I'll need some time before I can get a sample working for you. The libraries I'm using are internal to the company I'm working for. Briefly though, the company has it's own infrastructure for file based configuration of applications (which is injected based on the environment the app is deployed to). The library class in question extends PropertyPlaceholderConfigurer. On instantiating a bean of this library class, I expect that all placeholders are resolved by it. However, randomly (on some environments) https://github.com/spring-projects/spring-session/blob/master/spring-session/src/main/java/org/springframework/session/jdbc/config/annotation/web/http/JdbcHttpSessionConfiguration.java#L192 gets instantiated first and doesn't resolve any of the configuration placeholders that I define.

Would it be possible to ignore placeholders it cannot resolve so that another resolver can kick in?

vpavic:
@sayan-chaliha Could you use the same workaround as suggested in [#867 (comment)](https://github.com/spring-projects/spring-session/issues/867#issuecomment-328095462)?

> Can you attempt to override the `PropertySourcesPlaceholderConfigurer` provided by Spring Session configuration by assigning the `propertySourcesPlaceholderConfigurer` id to your `PropertySourcesPlaceholderConfigurer` bean?

sayan-chaliha:
Sure... works for me.

Thanks,
Sayan.

vpavic:
@sayan-chaliha There are proposed changes to Spring Session JDBC configuration which would eliminate the need for having `PropertySourcesPlaceholderConfigurer` bean registered. You can track that effort in #907.

PR #907, opened 2017-10-26
Add support for configuring JDBC session cleanup cron

vpavic:
While we currently do have _some_ configuration support for customizing expired session cleanup cron, this binds directly to specific property and is not exposed using standard configuration facilities (`JdbcHttpSessionConfiguration`/`@EnableJdbcHttpSession`). This makes it difficult to expose as application property in Spring Boot.

Customization of cron expression is desirable in scenarios with multiple active nodes, where cleanup should ideally run at different schedule on each node in order to reduce the possibility of deadlock occurrence.

Additionally, since current support is property based, it includes registration of `PropertySourcesPlaceholderConfigurer` bean in `JdbcHttpSessionConfiguration` which causes conflicts for some users (see #775).


rwinch:
Thanks @vpavic! Can you fix the conflicts and then merge this?

vpavic:
No problem, actually I was just about to address this.

PR #908, opened 2017-10-27
Harmonize Redis configurations

vpavic:
This PR improves reactive Redis configuration by adding support for `@SpringSessionRedisConnectionFactory` and `@SpringSessionRedisOperations` annotations that were introduced in #887.

Rather than introducing a new pair of annotations, the existing ones were moved from `org.springframework.session.data.redis.config.annotation.web.http` to `org.springframework.session.data.redis.config.annotation` so they could be shared.


rwinch:
Thanks @vpavic! I like that this moves the annotation to a shared package.  Once the tests pass, please merge this as well

PR #919, opened 2017-11-03
Make Redis namespace fully configurable

driverpt:
New Clean PR from #598

### DIFF
--- a/spring-session-data-redis/src/main/java/org/springframework/session/data/redis/RedisOperationsSessionRepository.java
+++ b/spring-session-data-redis/src/main/java/org/springframework/session/data/redis/RedisOperationsSessionRepository.java
@@ -588,7 +588,8 @@ public class RedisOperationsSessionRepository implements
 	}
 
 	public void setRedisKeyNamespace(String namespace) {
-		this.keyPrefix = DEFAULT_SPRING_SESSION_REDIS_PREFIX + namespace + ":";
+		Assert.hasText(namespace, "namespace cannot be null or empty");
+		this.keyPrefix = namespace.trim() + ":";
 	}
 
 	/**
--- a/spring-session-data-redis/src/test/java/org/springframework/session/data/redis/RedisOperationsSessionRepositoryTests.java
+++ b/spring-session-data-redis/src/test/java/org/springframework/session/data/redis/RedisOperationsSessionRepositoryTests.java
@@ -16,6 +16,7 @@
 
 package org.springframework.session.data.redis;
 
+import java.text.MessageFormat;
 import java.time.Duration;
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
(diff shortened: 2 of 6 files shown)


spring-session-data-redis/src/main/java/org/springframework/session/data/redis/RedisOperationsSessionRepository.java
```diff
@@ -588,7 +588,7 @@ private void publishEvent(ApplicationEvent event) {
 	}
 
 	public void setRedisKeyNamespace(String namespace) {
-		this.keyPrefix = DEFAULT_SPRING_SESSION_REDIS_PREFIX + namespace + ":";
+		this.keyPrefix = namespace + ":";
```
vpavic: We need to validate that provided namespace is not `null` or empty.

spring-session-data-redis/src/test/java/org/springframework/session/data/redis/RedisOperationsSessionRepositoryTests.java
```diff
@@ -709,9 +710,55 @@ public void setRedisFlushModeNull() {
 		this.redisRepository.setRedisFlushMode(null);
 	}
 
-	private String getKey(String id) {
-		return "spring:session:sessions:" + id;
-	}
+
+    @Test
+    public void testDefaultRedisNamespace() {
```
vpavic: No need to add test for default namespace, as other tests validate this already. OTOH we need to validate that namespace cannot be set to `null` or empty string as per previous comment.

spring-session-data-redis/src/main/java/org/springframework/session/data/redis/RedisOperationsSessionRepository.java
driverpt: How do you suggest to do that? Throw an Exception?

vpavic: Using `org.springframework.util.Assert#hasText` - I see you addressed that, thanks.

driverpt:
test this please

driverpt:
@vpavic , i've updated the code to include the tests.

How do i rebuild without adding commits on Travis ?

vpavic:
Thanks for the update @driverpt - you can trigger the new Travis CI build by amending your commit without updating its contents, and force pushing the changes to the remote:

```
$ git commit --amend --no-edit
$ git push --force
```

driverpt:
Is the master branch breaking ?

vpavic:
@driverpt Both `master` branch and your PR build are stable - thanks for the update.

vpavic:
Thanks for the contribution @driverpt. This is merged to `master` in 5f23a41 and polished in 36d349f.

driverpt:
👍

PR #995, opened 2018-02-13
Initialize SessionRepository lazy to avoid circular dependcies

mld-ger:
Hi,

I got a circular dependency issue using the `AbstractSessionWebSocketMessageBrokerConfigurer` together with Spring-Session-jdbc:

```
┌─────┐
|  entityManagerFactory
↑     ↓
|  webSocketBrokerConfig (field private org.springframework.session.SessionRepository org.springframework.session.web.socket.config.annotation.AbstractSessionWebSocketMessageBrokerConfigurer.sessionRepository)
↑     ↓
|  sessionRepository defined in class path resource [org/springframework/boot/autoconfigure/session/JdbcSessionConfiguration$SpringBootJdbcHttpSessionConfiguration.class]
↑     ↓
|  transactionManager defined in class path resource [org/springframework/boot/autoconfigure/orm/jpa/HibernateJpaAutoConfiguration.class]
↑     ↓
|  webSocketRegistryListener
└─────┘
```

Initializing the `SessionRepository` lazy fixed the problem for me.

### DIFF
--- a/spring-session-core/src/main/java/org/springframework/session/web/socket/config/annotation/AbstractSessionWebSocketMessageBrokerConfigurer.java
+++ b/spring-session-core/src/main/java/org/springframework/session/web/socket/config/annotation/AbstractSessionWebSocketMessageBrokerConfigurer.java
@@ -19,6 +19,7 @@ package org.springframework.session.web.socket.config.annotation;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.ApplicationEventPublisher;
 import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Lazy;
 import org.springframework.messaging.simp.config.ChannelRegistration;
 import org.springframework.session.Session;
 import org.springframework.session.SessionRepository;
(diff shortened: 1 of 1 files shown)


vpavic:
Thanks for the report @mld-ger.

I was unable to reproduce this problem with a simple app that uses Spring Boot, Spring Session JDBC and WebSockets. Can you provide a sample app to reproduce this?

mld-ger:
I was able to reproduce the issue with a simple app: https://github.com/mld-ger/Spring-Session-Circular-Dependency-Issue

The strange behavior is (I also had it in my production system that way) that you can start the app, but the integration test fails.

vpavic:
Thanks for the sample app @mld-ger, I was able to reproduce the described issue.

Since the problem only occurs in a `@SpringBootTest` I've opened spring-projects/spring-boot#12280 to get some feedback from the Boot team.

vpavic:
@mld-ger Let's move this to spring-projects/spring-boot#12280 as I don't believe the change proposed here should be applied. In any case, thanks for raising this.

As noted in the Boot issue comments, you can use the following workaround:

```java
@SpringBootTest(webEnvironment = WebEnvironment.RANDOM_PORT)
```

I've tested this with the test in your sample app, and it works.

Issue #1008, opened 2018-03-01
Upgrade to Gradle 4.6


jzheaux:

vpavic:
Closing in favor of PR #1009.

Note that if you intend to immediately submit the PR, the issue isn't needed.

PR #1009, opened 2018-03-02
Upgrade to Gradle 4.6

jzheaux:
Issue: gh-1008


vpavic:
Thanks for the PR @jzheaux. I've merged this to `master` in be1d3d3 and polished in a7a30da.

Note that it's generally preferable to upgrade Gradle wrapper using target Gradle version because that way:

- you get a wrapper binary that's generated by target version
- subsequent `gradle wrapper` invocations yield no change

PR #1037, opened 2018-04-04
Split read-only operations through their own DataSource

LouisVN:
If specified (through new annotations), the `SELECT` SQL statements will be going through a specific `DataSource` and `PlatformTransactionManager`. 
If not, the same objects will be used for all queries.

This would release load if the traffic is split between different database nodes for read and write operations.

I only added tests around `JdbcHttpSessionConfiguration` but please let me know if you think that this is a valid pull requests and want me to extend my work.

### DIFF
--- a/spring-session-jdbc/src/main/java/org/springframework/session/jdbc/JdbcOperationsSessionRepository.java
+++ b/spring-session-jdbc/src/main/java/org/springframework/session/jdbc/JdbcOperationsSessionRepository.java
@@ -186,7 +186,11 @@ public class JdbcOperationsSessionRepository implements
 
 	private final JdbcOperations jdbcOperations;
 
+	private final JdbcOperations jdbcOperationsReadOnly;
+
 	private final TransactionOperations transactionOperations;
+
+	private final TransactionOperations transactionOperationsReadOnly;
 
 	private final ResultSetExtractor<List<JdbcSession>> extractor = new SessionResultSetExtractor();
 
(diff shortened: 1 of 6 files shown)


vpavic:
Thanks for the PR @LouisVN.

While I understand the motivation behind the change you're proposing, it feels that this adds too much complexity for what it seems to be very specific use case.

I also think that read-write splitting is a problem that should be addressed at JDBC driver level, or even below, by some middleware solution that sits between DB server and (JDBC) clients.

Do you have any thoughts on this @rwinch?

rwinch:
Thanks for your contribution @LouisVN!

I agree with @vpavic that this adds too much complexity for a very specific use case. I'm closing this as won't fix.

LouisVN:
Thanks for your feedback @rwinch @vpavic.

I understand your point and just want to elaborate a little bit more the reasons that made me suggest this PR.

Coming back to your statement :
> I also think that read-write splitting is a problem that should be addressed at JDBC driver level

You are somehow correct. There is the [JDBC replication driver](https://dev.mysql.com/doc/connector-j/5.1/en/connector-j-master-slave-replication-connection.html) that is able to switch write/read traffic.
However, you still have to programmatically set your connection acquired from the `DataSource` with the following :
```
    Connection conn =
        driver.connect("jdbc:mysql:replication://master,slave1,slave2,slave3/test",
            props);
    conn.setReadOnly(false);
```
However, once connections (write/read mixed) are managed by a connection pool, switching and managing mixed connections is 'quite' costly in terms of performance.
It it usually a good practice -IMHO- to just use two distinct connections pools to separate the traffic.

> or even below, by some middleware solution that sits between DB server and (JDBC) clients.

There are some solutions indeed such as [Galera Cluster](http://galeracluster.com/) or [Percona XtraDB Cluster](https://www.percona.com/software/mysql-database/percona-xtradb-cluster) .
But those ones might also be configure with different IPs for write/read traffic to enhance resiliency.
You also have quite a few by nature Master/Slave solutions (e.g [MMM](http://mysql-mmm.org/mmm2:guide)) that will result in the same situation.

In those cases you are back to the what I just mentioned before (i.e either you use two distinct `DataSource` to separate write operations from the read ones, or use the JDBC replication driver + programmatic hint).

Sorry for the wall of text 😅 
I did not mean to argue but hope this clarifies my original intent.

Issue #1048, opened 2018-04-12
SessionRepositoryFilter is fetching session multiple (4) times per request 


cemo:
We are using JDBCSessionRepository and started to profile our application. On line 303 and 306 session is fetched twice: 

https://github.com/spring-projects/spring-session/blob/d97ad2ca3e2315f03203116a3ec74c1f7dab4fb5/spring-session-core/src/main/java/org/springframework/session/web/http/SessionRepositoryFilter.java#L303-L313

Here is the relevant `getRequestedSessionId` part. 

https://github.com/spring-projects/spring-session/blob/d97ad2ca3e2315f03203116a3ec74c1f7dab4fb5/spring-session-core/src/main/java/org/springframework/session/web/http/SessionRepositoryFilter.java#L354-L361

@vpavic I think this is a regression since I do not remember twice select before.

cemo:
@rwinch @vpavic 

I think this issue is very important since `getRequestedSessionId` is buggy now. This method is used in many places. For a simple request `findById` is called at least called 4 times.

vpavic:
Thanks for the report @cemo.

Why do you think the `SessionRepositoryRequestWrapper#getRequestedSessionId` is buggy? It does what it has to do in order to accommodate the changes to `HttpSessionIdResolver` (the successor to `HttpSessionStrategy`). For background on those changes see #275, #362 and #906.

That being said, I see your point with multiple retrieval of session as a part of `SessionRepositoryRequestWrapper#getSession(boolean)`. I'll see what can we do to optimize that.

cemo:
Sorry @vpavic. I did not mean buggy. In fact it has some performance issues. I think it can cache session in `getRequestedSessionId` as well.

vpavic:
Thanks for clarification @cemo. I think the `SessionRepositoryRequestWrapper#getRequestedSessionId` itself is fine, however some of its usages within the `SessionRepositoryFilter` could be optimized to avoid double retrieval from the session repository. That's the direction I'm currently looking into, will post back here when I have further updates.

cemo:
Thanks @vpavic. Please let me know after you optimized it. I can test for you.

cemo:
Please also note that it was called not only beginning of the request. At the end of request lifecycle it was called again multiple times. 

`org.springframework.session.web.http.SessionRepositoryFilter.SessionRepositoryRequestWrapper#commitSession`

`commitSession` session is called twice.

vpavic:
[...]

PR #1050, opened 2018-04-13
Optimize session resolution in SessionRepositoryFilter

vpavic:
This is a WIP PR to address #1048.

/cc @cemo


cemo:
I could not be sure that this is addressing this issue. From my poor view currently subsequent `getRequestedSessionId` calls makes db call. (I will check in detail soon)

I think you may consider to caching session in `getRequestedSession` by `org.springframework.session.web.http.SessionRepositoryFilter.SessionRepositoryRequestWrapper?

Issue #1064, opened 2018-05-05
Improve JdbcSession#setAttribute for updates


cemo:
I propose to add additional check when updating session:

https://github.com/spring-projects/spring-session/blob/751375338c73136b9b758e069539790620e51c5e/spring-session-jdbc/src/main/java/org/springframework/session/jdbc/JdbcOperationsSessionRepository.java#L738-L740

This check can be improved as this:

```
			else if (this.delegate.getAttribute(attributeName) != null) {
                                // attributeValue is not null here since it is checked before
                                // calling attributeValue's equals is important since it can be override by user
                                if(!attributeValue.equals(this.delegate.getAttribute(attributeName))){
				      this.delta.put(attributeName, DeltaValue.UPDATED);
                                 }
			}
```

This improvement may prevent update calls to DB.

candrews:
This change assumed that attribute values are immutable, which is not necessary the case.

For example, what if the value is a StringBuffer. A controller does getAttribute to get the StringBuffer object, calls .append on it, then calls setAttribute to persist it. Currently, this would behave as expected. With this proposed changed, the appended data would be lost, as the object reference has not changed (and the same object reference .equals itself).

cemo:
@candrews Would you give me please an example. I have thought on this a little bid but seems I am missing something.

candrews:
In request 1:
```
request.getSession().setAttribute("test", new StringBuilder("a"));
```

In request 2:
```
StringBuilder sb = (StringBuilder) request.getSession().getAttribute("test");
Assert.isTrue(sb.toString().equals("a"));
sb.append("b");
Assert.isTrue(sb.toString().equals("ab"));
request.getSession().setAttribute("test", sb);
```


In request 3:
```
StringBuilder sb = (StringBuilder) request.getSession().getAttribute("test");
Assert.isTrue(sb.toString().equals("ab"));
```

Currently, if you do 1, then 2, then 3 (in separate requests, so using separate controller methods, for example), all assertions will pass. Using your proposed change, the assertion in request 3 will fail.

vpavic:
For the reasons outlined by @candrews the proposed approach isn't something we can implement in a reliable manner. Therefore I'm closing this as invalid.

cemo:
Given the fact that such bad implementations can exist, this issue can be implemented for common types such as String and Number classes of JDK?

cemo:
@candrews primitives types are ok but I think It would make sense to support `String` as well since it is immutable as well.

Issue #648, opened 2016-10-13
Update an object stored 


crisarmen:
Hello,

I am trying to use spring-session-data-redis:1.2.2 on an existing project built with Spring/Spring MVC.

At the moment we have java objects stored in the session, and such objects are retrieved a modified in different parts of the code. Just as a simplified example, consider the simple object:
```
public class TestObject implements Serializable{
    private static final long serialVersionUID = -1283966188531894027L;  
    private Integer val;
    public TestObject(Integer val){
        this.val = val;
    }
    
    public Integer getVal() {
        return val;
    }

    public void setVal(Integer val) {
        this.val = val;
    }

    @Override
    public int hashCode() {
        final int prime = 31;
        int result = 1;
        result = prime * result + ((val == null) ? 0 : val.hashCode());
        return result;
    }

    @Override
    public boolean equals(Object obj) {
        if (this == obj)
            return true;
        if (obj == null)
            return false;
        if (getClass() != obj.getClass())
            return false;
        TestObject other = (TestObject) obj;
        if (val == null) {
            if (other.val != null)
                return false;
        } else if (!val.equals(other.val))
            return false;
        return true;
    }   
}
```

At a certain point in the flow we retrieve the object an modify the internal value doing something like:
```
TestObject t = (TestObject)WebUtils.getSessionAttribute(request, "testAttribute");
t.setVal(99);
```

And after in the flow we retrieve it and read the value doing something like:
```
TestObject t = (TestObject)WebUtils.getSessionAttribute(request, "testAttribute");
int value = t.getVal();  //here we expect to see 99
```
Before trying to use spring-session, here value was 99.
But now when I retrieve the value, I don't see 99, but I see the original value that the TestObject had when he was initially stored in the session.

Is there something I can configure in order to be able to  replicate the behaviour I had before using spring-session?

Few notes:
* redisFlushMode is configured to IMMEDIATE
* I appreciate that I could force a rewrite of the TestObject after the value is modified doing something like:
```
TestObject t = (TestObject)WebUtils.getSessionAttribute(request, "testAttribute");
t.setVal(99);
WebUtils.setSessionAttribute(request, "testAttribute", t);
``` 
but I was curious to know if there was something I could do without modifying the existing code.

Thanks in advance for your help.

candrews:
I also encountered this issue, here's how I'm working around it:

In an `@Configuration` class:
```
	@Bean
[...]

PR #1071, opened 2018-05-10
Attribute values that are unchanged and primitives are not changes

candrews:
If an attribute value stays the same by reference and is a primitive, then we know the data hasn't changed (because primitives are immutable) so there's no need to record a delta for that attribute.

This change is an optimization.

See https://github.com/spring-projects/spring-session/issues/1064

### DIFF
--- a/spring-session-jdbc/src/main/java/org/springframework/session/jdbc/JdbcOperationsSessionRepository.java
+++ b/spring-session-jdbc/src/main/java/org/springframework/session/jdbc/JdbcOperationsSessionRepository.java
@@ -16,6 +16,8 @@
 
 package org.springframework.session.jdbc;
 
+import java.math.BigDecimal;
+import java.math.BigInteger;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
@@ -57,6 +59,7 @@ import org.springframework.transaction.support.TransactionCallbackWithoutResult;
 import org.springframework.transaction.support.TransactionOperations;
 import org.springframework.transaction.support.TransactionTemplate;
 import org.springframework.util.Assert;
+import org.springframework.util.ClassUtils;
 import org.springframework.util.StringUtils;
 
 /**
@@ -732,7 +735,8 @@ public class JdbcOperationsSessionRepository implements
 
 		@Override
 		public void setAttribute(String attributeName, Object attributeValue) {
-			boolean attributeExists = (this.delegate.getAttribute(attributeName) != null);
+			final Object existingAttributeValue = this.delegate.getAttribute(attributeName);
+			boolean attributeExists = (existingAttributeValue != null);
 			boolean attributeRemoved = (attributeValue == null);
 			if (!attributeExists && attributeRemoved) {
 				return;
@@ -744,7 +748,7 @@ public class JdbcOperationsSessionRepository implements
 									? null
 									: deltaValue);
 				}
-				else {
+				else if (existingAttributeValue.getClass() != attributeValue.getClass() || !isImmutable(existingAttributeValue.getClass()) || !attributeValue.equals(existingAttributeValue)) {
 					this.delta.merge(attributeName, DeltaValue.UPDATED,
 							(oldDeltaValue, deltaValue) -> oldDeltaValue == DeltaValue.ADDED
 									? oldDeltaValue
@@ -762,6 +766,10 @@ public class JdbcOperationsSessionRepository implements
 					SPRING_SECURITY_CONTEXT.equals(attributeName)) {
 				this.changed = true;
 			}
+		}
+
+		private boolean isImmutable(Class<?> clazz) {
+			return ClassUtils.isPrimitiveOrWrapper(clazz) || clazz == BigDecimal.class || clazz == BigInteger.class || clazz == String.class || clazz == Class.class;
 		}
 
 		@Override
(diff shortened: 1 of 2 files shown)


spring-session-jdbc/src/main/java/org/springframework/session/jdbc/JdbcOperationsSessionRepository.java
```diff
@@ -735,11 +736,16 @@ public void setAttribute(String attributeName, Object attributeValue) {
...
 				this.delta.put(attributeName, DeltaValue.REMOVED);
 			}
-			else if (this.delegate.getAttribute(attributeName) != null) {
-				this.delta.put(attributeName, DeltaValue.UPDATED);
-			}
 			else {
-				this.delta.put(attributeName, DeltaValue.ADDED);
+				final Object existingAttributeValue = this.delegate.getAttribute(attributeName);
+				if (existingAttributeValue != null) {
+					if (existingAttributeValue == attributeValue && ClassUtils.isPrimitiveOrWrapper(attributeValue.getClass())) {
```
cemo: Your implementation seems not correct at first glance. You have not handled `existingAttributeValue != attributeValue` which has to update as well.

```
					if (existingAttributeValue != attributeValue) {
						this.delta.put(attributeName, DeltaValue.UPDATED);
					}else if(!ClassUtils.isPrimitiveOrWrapper(attributeValue.getClass() || !attributeValue.getClass().equals(String.class))){
                        this.delta.put(attributeName, DeltaValue.UPDATED);
                    }
```

candrews:
Added tests demonstrating that this change works.

I also check for String.class - the only immutable Java type that is not a primitive, as recommended at https://github.com/spring-projects/spring-session/issues/1064#issuecomment-388153131 by @cemo

vpavic:
Thanks for the PR @candrews. My initial thoughts on this are that if this kind of optimization is to be added to Spring Session, the proposed logic looks like the other `SessionRepository` implementation could benefit from it as well, which means that this could be considered in `SessionRepositoryFilter` as well.

Having said that, I don't think this is a `2.0.x` candidate so we'll take a closer look in due course.

candrews:
> My initial thoughts on this are that if this kind of optimization is to be added to Spring Session, the proposed logic looks like the other SessionRepository implementation could benefit from it as well, which means that this could be considered in SessionRepositoryFilter as well.

That sounds logical... But all the `delta` logic is in JdbcOperationsSessionRepository. Doing as you suggested would require a major refactor to hoist the delta functionality to the SessionRepository or SessionRepositoryFilter levels. I agree that such a refactor may not be in the 2.0.x scope, but as they say, the enemy of the good is the perfect - the approach in this PR is good, and in my opinion, it's best to get this improvement in (despite it not being perfect) instead of wait for a refactor that may never happen.

vpavic:
Either way, I'm not in favor of getting this merged in a maintenance release. Recently introduced regression with batch operation should be a lesson in that regard.

candrews:
FWIW, I updated this PR, fixing the conflicts.

vpavic:
We'll review and consider this for `2.1.x` @candrews, thanks for keeping the PR up to date.

Regarding the change itself, I believe (as previously stated) that this logic qualifies as generally reusable and therefore should be placed in `HttpSessionAdapter#setAttribute`. That way all the `Session` /`SessionRepository` implementations would benefit of this optimization and maintenance would be much easier.

vpavic:
Thanks for the PR again @candrews, however after discussing it with @rwinch we've decided not to merge it as the general feeling is that it adds complexity for a very limited amount of use cases.

candrews:
Can you please reconsider your position, @rwinch and @vpavic?

I'm currently working on porting a decent number of (very legacy) applications. The existing architecture of these applications is to use session affinity and in-memory session storage. In order to be more fault tolerant (among other reasons), the new architecture does not use session affinity and uses Spring Session JDBC to ensure session availability on any application server.

Since the old architecture uses in memory storage, if a mutable object is retrieved from the session (using `HttpSession.getAttribute`) then modified, the modifications are "saved" without actually have to call `HttpSession.setAttribute`.

Using Spring Session JDBC, that's not the behavior - you must `setAttribute` in order for changes to be saved (even if those changes are made to a mutable object already in session). See #648.

As a workaround, I devised `SessionSetWhenGetAttributesFilter` - see https://github.com/spring-projects/spring-session/issues/648#issuecomment-388057938

Under these circumstances, there are a _lot_ of updates to attribute values that haven't changed. Hence, the changes made in this PR result in a significant performance improvement by appreciably reducing the number of JDBC queries executed per request.

This PR is tested, complies with the contribution guidelines, fixes a real problem, and probably provides a potentially small performance improvement for all users (and a decent-sized performance improvement for some users).

rwinch:
I am pretty firm on this PR being rejected. The problem is it adds complexity without addressing the entire problem. If we accept the PR, users will come back asking why other immutable objects are not supported.

Perhaps an approach you could take is decorating the Session that is returned. If setSession is invoked, you can compare to the original value and see if you really want to set it on the delegate Session. This would achieve the same performance gains  you are wanting.

candrews:
Thanks for the info, I greatly appreciate you taking the time to provide it.

candrews:
I humbly request that this PR be reopened for consideration once again.

As I'm doing more performance testing, I've determined that unnecessary session attribute updates are significant problem that would be really nice to eliminate. I also suspect such an improvement would benefit other Spring Session users, such as anyone who uses a workaround or eventual included solution to https://github.com/spring-projects/spring-session/issues/1405 or https://github.com/spring-projects/spring-session/issues/648.



> If we accept the PR, users will come back asking why other immutable objects are not supported.

Then perhaps we Spring Session could then accept such changes to its immutable object detection.

I think having a 95% percent solution is better than not having a solution at all.

> Perhaps an approach you could take is decorating the Session that is returned. If setSession is invoked, you can compare to the original value and see if you really want to set it on the delegate Session. This would achieve the same performance gains you are wanting.

Sure, but why have everyone have to discover the need to do so and then reimplement it? Wouldn't it be better to include this optimization that benefits everyone, even if there are some cases that it doesn't perfectly optimize?

PR #1116, opened 2018-07-11
Fix garbled syntax relating to dropped APIs

dsyer:
I assume this is what it was meant to say. The language in the published guide isn't really in proper English.


rwinch:
Thanks for the PR @dsyer! This is merged into master

Issue #1005, opened 2018-02-26
Support SameSite Cookie


rwinch:
We should add support for using SameSite Cookie for session cookie to mitigate CSRF attacks. See https://scotthelme.co.uk/csrf-is-dead/

richard1122:
Hi, is there any plan to support SameSite cookie in spring-session 1.x.x ?

rwinch:
@richard1122 If someone sent a PR that was passive with tests we would consider it.

jmpavlec:
Was this included in a migration guide anywhere or mentioned in the docs? Making `SameSite=lax` for the SESSION cookie as the new default for Spring Session has some implications depending on the browser used. It potentially breaks some cross-domain SSO/SLO functionality that was dependent on the SESSION cookie. Safari also has a bug related to overriding it to be `SameSite=none` https://bugs.webkit.org/show_bug.cgi?id=198181

vpavic:
You're right @jmpavlec, we should provide some info about SameSite support in our reference manual. I've opened #1517 to take care of that.

If you're facing any issues due to presence of SameSite directive in our session cookie, you could disable it by providing a `null` value to `DefaultCookieSerializer#setSameSite`. Note that this requires overriding the default cookie serializer by providing your own `DefaultCookieSerializer` `@Bean`.

PR #1132, opened 2018-07-20
Add support for SameSite cookie directive

vpavic:
This is a WIP branch for #1005.

There are still some failing tests, due to pain of dealing with cookie directives that don't have support in Servlet spec (lik `SameSite` directive is), but please take a look on the general approach @rwinch.


vpavic:
Note to self: Since we write `Set-Cookie` header manually, `org.apache.tomcat.util.http.Rfc6265CookieProcessor#generateHeader` is probably something worth looking at for reference.

PR #1162, opened 2018-08-19
Switch to Spring Boot 2.1.0.M1 Dependencies BOM for dependency manage…

jxblum:
…ment.

Closes gh-1162.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,14 +7,20 @@ buildscript {
 		maven { url 'https://repo.spring.io/plugins-release' }
 	}
 }
+
 apply plugin: 'io.spring.convention.root'
 
 group = 'org.springframework.session'
 description = 'Spring Session'
 
+repositories {
+	maven { url "https://repo.spring.io/libs-milestone" }
+}
+
+ext['groovy.version'] = "2.4.15"
+ext['reactor-bom.version'] = "Californium-M2"
+ext['spring.version'] = "5.1.0.RC2"
 
 ext.releaseBuild = version.endsWith('RELEASE')
 ext.snapshotBuild = version.endsWith('SNAPSHOT')
 ext.milestoneBuild = !(releaseBuild || snapshotBuild)
-
-
--- a/gradle/dependency-management.gradle
+++ b/gradle/dependency-management.gradle
@@ -1,10 +1,6 @@
 dependencyManagement {
 	imports {
-		mavenBom 'com.fasterxml.jackson:jackson-bom:2.9.6'
-		mavenBom 'io.projectreactor:reactor-bom:Californium-M2'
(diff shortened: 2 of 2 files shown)


rwinch:
Thanks for the PR @jxblum! Unfortunately, we cannot use Spring Boot for our dependency management because this causes a cycle. Spring Boot already relies on Spring Session so Spring Session cannot rely on Spring Boot

jxblum:
I thought about this (too) and it does not create a direct cycle (since `gradle/dependency-manage.gradle` is only ever used to resolve the dependencies and build _Spring Session_).

The generated POMs and resulting artifacts from a _Spring Session_ build will not have any trace of _Spring Boot_.  In this case, you are simply using _Spring Boot's_ Dependencies BOM to align with versions of dependencies that overlap between _Spring Session_ and the version of Boot that will include _Spring Session_ (e.g. Jackson, etc).  It is much cleaner than declaring each dependency with a higher chance of a conflict if the dependency gets missed during the upgrade.

Anyway, I suspected this would get rejected for this very reason and it is not super critical, just a convenience.

rwinch:
It is a cycle. How would we perform a release with the latest dependencies of Spring Session without getting a release of Spring Boot?

jxblum:
It's a baseline.  

_Spring Session_ overrides the dependencies (from _Spring Boot_) that it requires (e.g. _Spring Boot_ `2.1.0.M1` currently pulls _Spring Framework_ `5.1.0.RC1`; _Spring Session_ overrides and sets the version to `5.1.0.RC2`).  You release _Spring Session_ `2.1.0.M2`.  Then _Spring Boot_ releases (`2.1.0.M2`) with _Spring Session_ BOM `Bean-M2`, which pulls in _Spring Session_ `2.1.0.M2`.  Immediately after the _Spring Boot_ `2.1.0.M2`, you rebase _Spring Session_ `2.1.0.BUILD-SNAPSHOT` on _Spring Boot_ Dependencies BOM, `2.1.0.M2` (which would have the upgraded _Spring Framework_ `5.1.0.RC2` version) allowing you to remove `ext['spring.version'] = "5.1.0.RC2`.

Anyway.  Like I already said, this is not a necessity, it is a convenience.  Agreed that there is a cycle, but it is not 1 that affects _Spring Boot_.

vpavic:
I agree with @rwinch, regardless of the fact that it doesn't reflect in published POMs, technically it's still a cycle.

Also keep in mind that the only reason this repository has any reference to Spring Boot is due to the sample apps, and IMO we should strive to keep it that way. Dependency management, especially with libraries, is something that should be as straightforward as possible. Introducing an intermediary BOM just to immediately override the versions it provides isn't straightforward IMO.

Issue #831, opened 2017-07-12
Hazelcast implementation of `ReactorSessionRepository`


vpavic:
We should provide a Hazelcast based implementation of `ReactorSessionRepository`.

vpavic:
While Hazelcast currently has reactive support, it is based on its own [`ICompletableFuture`](http://docs.hazelcast.org/docs/3.8/javadoc/com/hazelcast/core/ICompletableFuture.html) which extends from `Future` and not `CompletableFuture` since Hazelcast 3.x maintains compatibility with Java 6.

The plan is to replace `ICompletableFuture` with `CompletableFuture` in Hazelcast 4.0 when Java 8 becomes baseline - see hazelcast/hazelcast#3622.

Perhaps we should hold off our efforts of providing Hazelcast backed `ReactorSessionRepository` until Hazelcast 4.0 becomes available (ATM there's no target date for [the milestone](https://github.com/hazelcast/hazelcast/milestone/85)). WDYT @rwinch?

hantsy:
The hazelcast itself have Spring integration in the hazelcast  core, I am not sure if it works when implementing a custom MapReactiveSessionRepository with a hazelcast managed `Map` as the constructor arguments ?

d3bbrower:
Any plans to resurrect this @vpavic @rwinch now that Hazelcast 4 is out.

AndreasKasparek:
Hi @vpavic @eleftherias,
I have implemented a `ReactiveSessionRepository` based on Hazelcast (following the general approach of the existing non-reactive Hazelcast4 session repo and reusing the existing entry processor) and was wondering if I should integrate this with Spring? I am a bit short on time at the moment and not sure if I can provide a full pull request. But I could at least publish my branch and let others fork from it and complete it, if you think that is worth it.

Hazelcast 4.2.4 returns `CompletionStage` from its async `IMap` methods, so the integration with the reactor chain is not that complicated. My solution does not support `FlushMode` because I don't see a good way how to properly implement immediate flushing via async saves in blocking methods like session.setAttribute(). And the `FindByIndexNameSessionRepository` interface is also not reactive and thus I omitted it.

Thanks for your feedback.

eleftherias:
Thanks @AndreasKasparek.

We're still planning to wait on hazelcast/hazelcast#3622 before we add this feature to Spring Session. 
We want to make sure we can provide users with the full functionality of Spring Session and Hazelcast, rather than adding a partial solution.

[...]

PR #1174, opened 2018-08-29
darft of reactive Hazelcast session repository

OlegDokuka:
Provided a draft implementation of `HazelcastReactiveSessionRepository` which gives an asynchronous access to stored in Hazelcast sessions and may benefits resources utilization and apps performance

### DIFF
--- a/spring-session-hazelcast/spring-session-hazelcast.gradle
+++ b/spring-session-hazelcast/spring-session-hazelcast.gradle
@@ -1,6 +1,9 @@
 apply plugin: 'io.spring.convention.spring-module'
 
 dependencies {
+
+  	optional "io.projectreactor:reactor-core"
+
 	compile project(':spring-session-core')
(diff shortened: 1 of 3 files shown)


taylancapraz (approved):

taylancapraz:
is there any plan for this task?

DidierLoiseau:
Oh man, I hadn’t noticed this PR existed when I created #2222, you didn’t reference #831…

Our implementations are very similar, except yours is for Hazelcast 3 and you chose to remove the support of application events.

I guess both could be merged if they want to support both Hazelcast 3 and 4+ in 2.x (Hazelcast 3 support was removed in spring-session 3.x).

StevenPG:
Is there further decision making on this? Using Spring Boot 3.1.4 with spring session and spring cloud gateway and this is a blocker, since gateway uses Webflux.

Issue #1242, opened 2018-11-02
SessionRepositoryFilter should commit session before an include dispatch


jzheaux:
Related to https://github.com/spring-projects/spring-security/issues/5499

Headers cannot be written after an include dispatch (e.g. `request.getRequestDispatcher(path).include`), meaning that any header-based session id resolution strategy will fail to write the session id to the response.

vpavic:
Thanks @jzheaux - I believe this is a duplicate of #1214. Also note that there's no need to open an issue if you plan to submit a PR straight away.

vpavic:
Duplicate of #1214

PR #1243, opened 2018-11-02
Commit Session on Include Dispatch

jzheaux:
The servlet spec disallows any writing of headers after an include has
been issued.

This commit intercepts the include and commits the session, then
allowing the include to proceed.

Fixes: gh-1242


vpavic:
Thanks @jzheaux!

Issue #1253, opened 2018-11-09
Cookie http-only setting has no effect when using Spring Session


hellowenqi:
Hi, it still returns HttpOnly in Set-Cookie even if i set server.servlet.session.cookie.http-only=false in application.properties.

Spring Session version:2.1.1.RELEASE
Sample: https://github.com/hellowenqi/HttpOnly.git

reference issue described in https://github.com/spring-projects/spring-boot/issues/15139

rwinch:
Thanks for the report. I created #1254 to fix this. It appears for some reason we did not look at the HttpOnly setting.

In the meantime, you can work around this by creating an instance of `DefaultCookieSerializer` and exposing it as a `@Bean`.

vpavic:
Yes, we don't map `httpOnly` and `secure` attributes from `SessionCookieConfig` because they are `boolean` and both default to `false` there.

This is in contrast to our defaults where we have `httpOnly` set to `true` and for `secure` we use what `javax.servlet.ServletRequest#isSecure` says unless explicitly configured. So we cannot map those safely as it would override our defaults, which we consider to be more secure.

Also see #713 for background.

hellowenqi:
@vpavic  thanks, it works to create an instance of DefaultCookieSerializer . Here is my config:

```
package com.cmbc.bigdata;

import org.springframework.context.annotation.Bean;
import org.springframework.session.data.redis.config.annotation.web.http.EnableRedisHttpSession;
import org.springframework.session.web.http.CookieSerializer;
import org.springframework.session.web.http.DefaultCookieSerializer;

@EnableRedisHttpSession
public class Config {

  @Bean
  public CookieSerializer cookieSerializer() {
    DefaultCookieSerializer serializer = new DefaultCookieSerializer();
    serializer.setUseHttpOnlyCookie(false);
    return serializer;
  }

}

```

vpavic:
Removing `bug` label as this is a limitation of our capability to configure `DefaultCookieSerializer` off `SessionCookieConfig`. Unfortunately there isn't much we can do about it.

I'll open a Spring Boot issue to propose improvement of Spring Session's auto-configuration support with `DefaultCookieSerializer` configuration straight from configuration properties, thus bypassing the `SessionCookieConfig`.

vpavic:
Closing this in favor of spring-projects/spring-boot#15139, which was just re-opened.

PR #1254, opened 2018-11-09
Fix SessionCookieConfig.isHttpOnly

rwinch:
Previously, SessionCookieConfig.isHttpOnly was ignored by the default
CookieSerializer.

This commit updates the default CookieSerializer to respect the
SessionCookieConfig.isHttpOnly flag.

Fixes: gh-1253


rwinch:
Thanks @vpavic! For the sake of transparency, as discussed offline, I agree we cannot make these changes.

Issue #1395, opened 2019-03-27
RedisSessionExpirationPolicy's cleanExpiredSessions() doesn't remove the session id from Index


cuckoo3:
RedisSessionExpirationPolicy's cleanExpiredSessions():

	public void cleanExpiredSessions() {
		long now = System.currentTimeMillis();
		long prevMin = roundDownMinute(now);

		if (logger.isDebugEnabled()) {
			logger.debug("Cleaning up sessions expiring at " + new Date(prevMin));
		}

		String expirationKey = getExpirationKey(prevMin);
		Set<Object> sessionsToExpire = this.redis.boundSetOps(expirationKey).members();
		this.redis.delete(expirationKey);
		for (Object session : sessionsToExpire) {
			String sessionKey = getSessionKey((String) session);
			touch(sessionKey);
		}

	}

It only remove the session but not the session key from the index PRINCIPAL_NAME_INDEX_NAME

Since the TTL of the index in Redis is -1, all the expired session key will be kept forever and the index will keep growing in size.

yevgenb:
Will this fix get to 2.4.x branch?

vpavic:
@cuckoo3 I'm afraid I can't reproduce the problem - can you provide more details about the how to reproduce it? A sample app would be most helpful.

Also, did you read the documentation on [events support for `RedisOperationsSessionRepository`](https://docs.spring.io/spring-session/docs/current/reference/html5/#api-redisoperationssessionrepository-sessiondestroyedevent) and Redis documentation on [keyspace notifications](https://redis.io/topics/notifications)? Note that with automatic registration (that Spring Session does by default) if you restart Redis server the subscription will be lost, until you restart your application too.

cuckoo3:
@vpavic Thanks for your explanation. I will double check if it's related to the AWS setting for missing the event notification when the session is expired. We are using the cronjob cleanupExpiredSessions() to remove the session before it's expired from the Redis server. However the cronjob didn't remove the session id from the index directly and we are facing a problem in glowing up of the index size.

BTW [1.4.x](https://github.com/spring-projects/spring-session/branches) cannot be built due to the logout.html are using  "https://www.thymeleaf.org/dtd/xhtml1-strict-thymeleaf-spring4-3.dtd". http should be used instead.

vpavic:
I assume this was a configuration error @cuckoo3?

Regarding `1.4.x`, that's branch is obsolete and we never actually made a release from there.

[...]

PR #1396, opened 2019-03-28
Fix issue #1395. 

cuckoo3:
When remove the expired key, call the SessionRepository to remove the session ID from the index too
Fix issue: https://github.com/spring-projects/spring-session/issues/1395


cuckoo3:
Some test case will return error:

Root cause: redis.clients.jedis.exceptions.JedisDataException: ERR Unsupported CONFIG parameter: notify-keyspace-events

Need to add the 

```
@Bean
public ConfigureRedisAction configureRedisAction() {
	return ConfigureRedisAction.NO_OP;
}
```

Then for the logout.html, after chnaged to https:
```
<!DOCTYPE html SYSTEM "https://www.thymeleaf.org/dtd/xhtml1-strict-thymeleaf-spring4-3.dtd">
```

The test case will return error as below:

```
org.openqa.selenium.NoSuchElementException: Unable to locate element using css: com.gargoylesoftware.htmlunit.TextPage@7275be99
For documentation on this error, please visit: http://seleniumhq.org/exceptions/no_such_element.html

```

Either remove the URL from DOCTYPE:
```
<!DOCTYPE html>
```

or use the **http** instead of **https**

vpavic:
@cuckoo3 Thanks for the PR but let's discuss the problem in #1395 first, and try to understand it before attempting to fix it. There seem to be a lot of unrelated changes here.

PR #1398, opened 2019-03-29
1.3.x

cuckoo3:
When remove the expired key, call the SessionRepository to remove the session ID from the index too
Fix issue: https://github.com/spring-projects/spring-session/issues/1395


cuckoo3:
custom-cookie.integration-test will throw exception "first visit no attributes"  when 
PrincipalNameResolver.resolvePrincipal(...){
... session.getAttribute(...)

Issue #1278, opened 2018-12-04
Consider adding a simple Redis SessionRepository implementation


vpavic:
A lot of complexity in `RedisOperationsSessionRepository` is focused around support for Servlet API's `HttpSessionEvent`. Since many of users do not actually use session events, we can consider providing an alternative, simpler implementation of Redis based `SessionRepository` that wouldn't support session event handling.

/cc @rwinch

herveDarritchon:
👍 

I am very very interested in a simpler repo which doesn't use any events/messaging (pub/sub REDIS stuff) because we have Dynomite over REDIS to insure the cross datacente replication and Dynomite doesn't support pub/sub :)
I tried to do it myself but I am stuck in an issue around redefining the sessionRepository bean with my custom implementation.
I ask on gitter and stackoverflow (https://stackoverflow.com/questions/55818188/custom-spring-session-repository-redis-with-dynomite) for help with no solution.
If anyone can help me, it would be great.

I have difficulties to override the sessionRepository implementation from org.springframework.session:spring-session-data-redis with my custom one.

I tried to set up a config class such as : 

```
@Configuration
@EnableRedisHttpSession
public class SessionConfig extends AbstractHttpSessionApplicationInitializer {

    @Bean
    @Primary
    public SessionRepository sessionRepository() {
        return new RedisDynoSessionRepository();
    }
}
```

And the signature of my Custom Implementation of the SessionRepository is :

```
public class RedisDynoSessionRepository implements
        FindByIndexNameSessionRepository<RedisDynoSessionRepository.RedisSession> {
```

I tried to put the override bean configuration (`spring.main.allow-bean-definition-overriding=true`) or not with no success.

vpavic:
Closing in favor of PR #1408.

Issue #1277, opened 2018-12-04
Reconsider handling of SessionRepository#save for invalidated sessions


vpavic:
At present, the `SessionRepository` implementations are silent when an invalid session is passed to `#save` operation. This can have unexpected consequences in further processing of HTTP request, since it operates under assumption that the saved session is still valid - i.e. there's no difference in `#save` when passed in session is valid vs invalid.

One option could be (depending on the nature of the underlying data store) to do a read before saving a session, and throw an error if session is missing/invalid.

We could look at `WebSession#save` for inspiration - see #1135.

/cc @rwinch @jxblum @gregturn

chrisgolle:
Just wanted to comment that I see opposite behaviour using RedisSessionRepository and would prefer it were silent haha. Rarely, i have requests come in with an existing session and they are unlucky enough to be gone from Redis at the time the session is saved, triggering the IllegalStateException below.

`	@Override
	public void save(RedisSession session) {
		if (!session.isNew) {
			String key = getSessionKey(session.hasChangedSessionId() ? session.originalSessionId : session.getId());
			Boolean sessionExists = this.sessionRedisOperations.hasKey(key);
			if (sessionExists == null || !sessionExists) {
				throw new IllegalStateException("Session was invalidated");
			}
		}
		session.save();
}`

PR #1408, opened 2019-05-01
Add simple Redis SessionRepository implementation

vpavic:
This is a PR to address #1278.


vpavic:
Some things to point out about this implementation, since they differ from `RedisOperationsSessionRepository` but I believe are the way to go in future:

- the `#setKeyNamespace` is transparent in this implementation whereas `RedisOperationsSessionRepository` appends `:`, which both adds complexity and is obviously not so transparent
- it uses `RedisOperations<String, Object>` vs `RedisOperations<Object, Object>` in `RedisOperationsSessionRepository` (FWIW `ReactiveRedisOperationsSessionRepository` uses `ReactiveRedisOperations<String, Object>`)
- it aligns with #1277 by throwing `IllegalStateException` if `#save` is attempted for an invalid (i.e. not present in data store) session

The following open points are:
- naming: something like `DefaultRedisSessionRepository` or `BasicRedisSessionRepository` comes to mind - we can also rename `RedisOperationsSessionRepository` to something more appropriate
- configuration support: IMO this doesn't need to be imminent for `2.2`

vpavic:
> @vpavic Can you add a sample that uses this?

Yes, that would be a good thing to have.

vpavic:
I've added a sample application to showcase the new implementation.

PR #1439, opened 2019-06-01
In Oracle ignore non-existant tables when dropping

candrews:
If tables don't exist, ignore that error when dropping.

This aligns Oracle with the existing mysql, postgresql, h2, and hsqldb behavior.


candrews:
@vpavic can you please take a gander at this one?

rwinch:
Thanks for the PR @candrews! This is now merged into master

PR #1449, opened 2019-06-10
Add nohttp

rwinch:


gregturn:
Isn’t it at 0.0.3?

build.gradle
```diff
@@ -15,10 +15,16 @@ buildscript {
 	dependencies {
 		classpath 'io.spring.gradle:spring-build-conventions:0.0.25.RELEASE'
 		classpath "org.springframework.boot:spring-boot-gradle-plugin:$springBootVersion"
+		classpath 'io.spring.nohttp:nohttp-gradle:0.0.2.RELEASE'
```
vpavic: Can we add this to `spring-build-conventions` so that we only need to apply the plugin in the project build? Spring Security would benefit of this too.

rwinch: Yes. I just haven't done it yet.

rwinch:
@gregturn The snapshot is, but I want to use the latest stable

PR #1522, opened 2019-10-03
Use wnameless/oracle-xe-11g-r2 for Oracle testing

candrews:
According to https://github.com/wnameless/docker-oracle-xe-11g the DMCA issues have been resolved.

This change allows Oracle integration tests to be run.

### DIFF
--- a/spring-session-jdbc/src/integration-test/resources/testcontainers.properties
+++ b/spring-session-jdbc/src/integration-test/resources/testcontainers.properties
@@ -1,1 +1,2 @@
 ryuk.container.timeout=120
+oracle.container.image=wnameless/oracle-xe-11g-r2


vpavic:
I see the use of this image as highly problematic and risky, as there's no license acceptance involved, and you do require to accept one when downloading from Oracle's website. See:

- https://www.oracle.com/database/technologies/xe-prior-releases.html
- https://www.oracle.com/downloads/licenses/database-11g-express-license.html

This is in contrast to what we have with e.g. DB2 and SQL Server:
https://github.com/spring-projects/spring-session/blob/e2abe36fa8bc28b5ee094bb421a6a5b61db4f9e2/spring-session-jdbc/src/integration-test/resources/container-license-acceptance.txt#L1-L2

candrews:
I've asked if they've resolved the issues as their README seems to indicate they have: https://github.com/wnameless/docker-oracle-xe-11g/issues/118#issuecomment-538030143

vpavic:
The point is that I still don't see how are we accepting the license terms by using this image? This is problematic, because as indicated in my previous comment, those are required when you download from Oracle's website.

PR #1539, opened 2019-10-17
Parse expression only once in PrincipalNameIndexResolver

dreis2211:
Hi,

I noticed in some profiling runs of my application that a small share of cycles goes into parsing the expression in PrincipalNameIndexResolver, which seems redundant. This PR puts the expression into a field and reuses it.

Let me know what you think.
Cheers,
Christoph


spring-session-core/src/main/java/org/springframework/session/PrincipalNameIndexResolver.java
```diff
@@ -34,6 +34,8 @@
 
 	private static final SpelExpressionParser parser = new SpelExpressionParser();
 
+	private static final Expression expression = parser.parseExpression("authentication?.name");
```
vpavic: I think we should also get rid of `parser` as member variable and simply instantiate it here as it won't be reused.

dreis2211:
@vpavic Thanks for the review. Done

Issue #1517, opened 2019-09-27
Document support for SameSite cookie directive


vpavic:

vpavic:
Closing in favor of PR #1543.

PR #1543, opened 2019-11-08
Document support for SameSite cookie directive

eleftherias:
Resolves: #1517

### DIFF
--- a/spring-session-docs/src/docs/asciidoc/index.adoc
+++ b/spring-session-docs/src/docs/asciidoc/index.adoc
@@ -596,6 +596,7 @@ You can browse the complete link:../../api/[Javadoc] online. The key APIs are de
 * <<api-reactivemapsessionrepository>>
 * <<api-jdbcindexedsessionrepository>>
 * <<api-hazelcastindexedsessionrepository>>
+* <<api-cookieserializer>>
 
 [[api-session]]
 === Using `Session`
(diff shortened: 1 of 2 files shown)


eleftherias:
@vpavic Sounds good. 
I will add an entry in the API Documentation section about `DefaultCookieSerialzer`, is that what you had in mind?

vpavic:
Yes, that's what I was aiming for. Thanks.

eleftherias:
@vpavic I have added the new section to the API Documentation. Thanks for your feedback and let me know any further thoughts.

spring-session-docs/src/docs/asciidoc/index.adoc
```diff
@@ -1230,6 +1231,67 @@ Note that if you use Hazelcast's `MapStore` to persist your sessions `IMap`, the
...
+With some implementations (that is, Redis) this option provides no performance benefit.
+However, it can help with tracing logs of a particular user.
+* `domainName`: Allows specifying a specific domain name to be used for the cookie.
+This option is simple to understand but often requires a different configuration between development and production environments.
+See `domainNamePattern` as an alternative.
+* `domainNamePattern`: A case-insensitive pattern used to extract the domain name from the `HttpServletRequest#getServerName()`.
+The pattern should provide a single grouping that is used to extract the value of the cookie domain.
+If the regular expression does not match, no domain is set and the existing domain is used.
+If the regular expression matches, the first https://docs.oracle.com/javase/tutorial/essential/regex/groups.html[grouping] is used as the domain.
+* `sameSite`: The value for the `SameSite` cookie directive.
```
vpavic: I think we should mention here that `null` could be used as value to disable the serialization of `SameSite` directive.

eleftherias: Done

PR #1544, opened 2019-11-12
Backport of #1545: Session creation with pre-specified id.

zapp88:
This pull request address specific use-case of creating session with pre-specified id.

While standard `SessionRepository` interfaces provides means of creating session via `createSession()` - it does noting to let you control the 'id' creation process. All known to me implementation of `SessionRepository` (`Redis,Hazelcast,Jdbc`) fallback to calling `MapSession` and what is peculiar `MapSession` supports creating session object with externally defined id - via its secondary construtor - `new MapSession(id)`. 

However all implementations (`Redis,Hazelcast,Jdbc`) fallback to using the default construtor which in its default bahaviour calls `UUID.randomUUID().toString()`. While this provides necessary uniqueness to id (provied low probability of UUID ever repeating) it leaves us without controll over this process.

There are quite a few business casses when internaly created id would not be desired :
1. UUID randomnes not meeting business criteria
2. Integration with existing session managment systems where id creation is delegated to external system

We don't want to break any existing functionality (keeping `SessionRepository` as clean as possible) - here comes `CreateWithIdSessionRepository` -  it expands default implementation with ability to create session with pre-specified id via `S createSession(@Nullable String id);`
the same way `FindByIndexNameSessionRepository` expands session repository with ability to find session with principal/indexName. 

I took upon myself implementing this interfaces in 3 most common used implementations `(Redis,Hazelcast,Jdbc)` since it does not break any existing functionality while still providing benefit on controlling this process by adhering to proper interface.

There are no reasons it could not be supported in other available implementations.

PS. Sorry for my english. English is not my primary language.
If you have any suggestions on how i could improve my approach or would like to ask me a question - feel free to contact me. My email - zapp88@gmail.com


zapp88:
The travis build seems to be failing be cause of :

`Expected feature release number in range of 9 to 14, but got: 8
The command "~/bin/install-jdk.sh --target "/home/travis/oraclejdk8" --workspace "/home/travis/.cache/install-jdk" --feature "8" --license "BCL"" failed and exited with 3 during .`

No sure if it is mistake on my part.

zapp88:
Cancelled untill we decide weather we want #1545 or #1547

PR #1545, opened 2019-11-12
Session creation with pre-specified id.

zapp88:
This pull request address specific use-case of creating session with pre-specified id.

While standard `SessionRepository` interfaces provides means of creating session via `createSession()` - it does noting to let you control the 'id' creation process. All known to me implementation of `SessionRepository` (`Redis,Hazelcast,Jdbc`) fallback to calling `MapSession` and what is peculiar `MapSession` supports creating session object with externally defined id - via its secondary construtor - `new MapSession(id)`. 

However all implementations (`Redis,Hazelcast,Jdbc`) fallback to using the default construtor which in its default bahaviour calls `UUID.randomUUID().toString()`. While this provides necessary uniqueness to id (provied low probability of UUID ever repeating) it leaves us without controll over this process.

There are quite a few business casses when internaly created id would not be desired :
1. UUID randomnes not meeting business criteria
2. Integration with existing session managment systems where id creation is delegated to external system

We don't want to break any existing functionality (keeping `SessionRepository` as clean as possible) - here comes `CreateWithIdSessionRepository` -  it expands default implementation with ability to create session with pre-specified id via `S createSession(@Nullable String id);`
the same way `FindByIndexNameSessionRepository` expands session repository with ability to find session with principal/indexName. 

I took upon myself implementing this interfaces in 3 most common used implementations `(Redis,Hazelcast,Jdbc)` since it does not break any existing functionality while still providing benefit on controlling this process by adhering to proper interface.

There are no reasons it could not be supported in other available implementations.

PS. Sorry for my english. English is not my primary language.
If you have any suggestions on how i could improve my approach or would like to ask me a question - feel free to contact me. My email - zapp88@gmail.com

Issue #11, opened 2014-07-03
Provide Session Id Generation Strategy


rwinch:
We are currently using UIDs for session id generation. It may be better to include other characters to increase the entropy. We should be careful in how we use SecureRandom See http://www.cigital.com/justice-league-blog/2009/08/14/proper-use-of-javas-securerandom/ Last, we may want to make generating ids a strategy interface so that users can swap this out
- See PR #204

artgramlich:
+1
Thanks for closing the other issue, somehow I didn't see this issue.

artgramlich:
I submitted a pull request #204.

derylspielman:
This would be very helpful so that we can customize the session ID with a suffix of the jvmRoute. We have legacy code that requires the .jvmRoute at the end and currently calling `httpRequest.getSession().getId()` 
 comes without the jvmRoute. Any progress on this?

abhishekjoy1:
Hi, I am trying to overwrite the session-id generated through framework as follows:


`
       @EnableSpringHttpSession
        public class CustomSessionRepository implements SessionRepository<MapSession> {

            private Integer defaultMaxInactiveInterval;
            private  Map<String, Session> sessions = new ConcurrentHashMap<>();

            public CustomSessionRepository() {
                 this.sessions = sessions;
            }

            public void setDefaultMaxInactiveInterval(int defaultMaxInactiveInterval) {
                 this.defaultMaxInactiveInterval = defaultMaxInactiveInterval;
            }
       
            public void save(MapSession session) {
                if (!session.getId().equals(session.getOriginalId())) {
                    this.sessions.remove(session.getOriginalId());
               }

              this.sessions.put(session.getId(), new MapSession(session));
            }

            public MapSession findById(String id) {
                 Session saved = (Session)this.sessions.get(id);
                 if (saved == null) {
                     return null;
                 } else if (saved.isExpired()) {
                      this.deleteById(saved.getId());
                      return null;
                 } else {
                     return new MapSession(saved);
                 }
            }

            public void deleteById(String id) {
                this.sessions.remove(id);
            }

           public MapSession createSession() {
               MapSession result = new MapSession(myCustomId());
               if (this.defaultMaxInactiveInterval != null) {
                  result.setMaxInactiveInterval(Duration.ofSeconds((long)this.defaultMaxInactiveInterval));
             }

           return result;
         }
       }

`

PvanHengel:
[...]

PR #1547, opened 2019-11-14
Feature/id with strategy

zapp88:
After creating PR : #1545 I became aware of **long** ongoing disscusion about 
session generation strategy mainly : #204 which has merge conflicts and author is not present for quite a while, and a general issue #11 going back to 3 Jul 2014.

I tought i would try to takle the problem myself with the strategy approach (differently than in #1545). This PR is my attempt at assimilating strategy based way of controlling id generation process.

While I still prefer #1545 way of doing things 'this' PR is closer the expected behaviour of #11. 
But not without a few pros and cons. 

 #1545 is less invasive - it provides means of controlling session id without changing too much. #1545 doesn't break any existing functionality - especially it doesn't remove any existing methodes - it only adds new.

 At the same time `this` PR while gives us more elegant way of handling session id. It requires us to change a few things , mainly :

- Session creation is not session class reponsibility - it is session-repository reponsibility and as a result - `MapSession` should get valid id from _session-repository instead_ of creating it by itself. While this is easly guaranteed by removal of no-args constructor in `MapSession` it requires us to modify `changeId` methode signature in MapSession. The new id for session should be provided to session externally.

- Existing session repository implementations should provide a way of changing session generation strategy and as a result all session repository should implement additional method "changeSessionId".

Additionally since spring-session is a library - we don't want to introduce any new unwanted behaviour to projects rellying on it. Thats why i additionaly introduce UUIDSessionIdStrategy which is a default go-to behaviour for all session repositories if no strategy is provided.

- As a additional feature it would be nice to expand configuration with a means to customize session id strategy . For example add additional config field to EnableHazelcastHttpSession. 

For now i dont want this PR to grow to large or out of scope so I willy probably try to add the aformentioned feature as a separate PR if `this` gets accepted.

### DIFF
--- a/spring-session-core/src/main/java/org/springframework/session/MapSession.java
+++ b/spring-session-core/src/main/java/org/springframework/session/MapSession.java
@@ -23,7 +23,6 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
-import java.util.UUID;
 
 /**
  * <p>
@@ -44,6 +43,7 @@ import java.util.UUID;
  *
  * @author Rob Winch
  * @author Vedran Pavic
+ * @author Jakub Maciej
  * @since 1.0
  */
 public final class MapSession implements Session, Serializable {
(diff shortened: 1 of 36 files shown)


zapp88:
This feature is a work in progress untill i can fix integration tests.

zapp88:
Integration erros have been resolved

barryspearce:
I have a multiple concurrent file uploader which suffers from deadlocks on the session table with alarming frequency, resulting in the data having to be resent.

Does this change resolve the issue of deadlocks occuring where multiple requests are received for the same session using mariadb and innodb tables? 

For instance would this let me change the spring session table to the strategy used by the rest of my database which is db generated integer IDs?

If so is there any plan for including this in a release? I am sure a lot of people are keen to resolve the deadlock issue.

quaff:
@zapp88 I've created similar PR [2251](https://github.com/spring-projects/spring-session/pull/2251), could you take a look at that.

Issue #1531, opened 2019-10-08
why RedisSessionExpirationPolicy#onDelete remove something not exist in the expirations SET?


Alwayswithme:
the spring:session:expirations:${roundUpToNextMinute}  redis SET is store value call keyToExpire:
```
		String expireKey = getExpirationKey(toExpire);
		BoundSetOperations<Object, Object> expireOperations = this.redis.boundSetOps(expireKey);
		expireOperations.add(keyToExpire);
```
the keyToExpire is a string like expires:${sessionId}
https://github.com/spring-projects/spring-session/blob/927008bdc89701555f2f8f900f4f898df23b1476/spring-session-data-redis/src/main/java/org/springframework/session/data/redis/RedisSessionExpirationPolicy.java#L97

however, when delete the session, it actually remove the session id not expires:${sessionId}
```
	this.redis.boundSetOps(expireKey).remove(session.getId());
```

https://github.com/spring-projects/spring-session/blob/927008bdc89701555f2f8f900f4f898df23b1476/spring-session-data-redis/src/main/java/org/springframework/session/data/redis/RedisSessionExpirationPolicy.java#L70

Is this a defact or My misunderstanding?

vpavic:
The behavior of session expiration handling for Redis session repository is described both in the `RedisOperationsSessionRepository` javadoc and [reference manual](https://docs.spring.io/spring-session/docs/current/reference/html5/#api-redisoperationssessionrepository-storage).

Are you experiencing any particular issue here?

Alwayswithme:
org.springframework.session.data.redis.RedisOperationsSessionRepository#delete will delete the expireKey,

Alwayswithme:
@vpavic please take a look

eleftherias:
Thanks for the report @Alwayswithme. 
I'm going to close this issue as a duplicate of gh-585.
Please follow gh-585 for updates.

PR #1555, opened 2019-12-12
delete the member in the expire set

Alwayswithme:
When add to the redis set, prefix the sessionId with "expires:", but delete occur, it didn't have "expires:" prefix. So I create this PR.

#1531


eleftherias:
Thanks for the PR @Alwayswithme.
I'm going to close this in favour of gh-1800 which is very similar but a little bit closer to the format we're looking for.

PR #1576, opened 2020-02-03
Fix example in JavaDocs of EnableSpringHttpSession

Wrent:
The examples in JavaDocs of @EnableSpringHttpSession, SpringHttpSessionConfiguration and @EnableSpringWebSession were creating MapSessionRepository / ReactiveMapSessionRepository using a constructor, which no longer exists in the classes. This should allow the examples to be used out of the box.

### DIFF
--- a/spring-session-core/src/main/java/org/springframework/session/config/annotation/web/http/EnableSpringHttpSession.java
+++ b/spring-session-core/src/main/java/org/springframework/session/config/annotation/web/http/EnableSpringHttpSession.java
@@ -40,7 +40,7 @@ import org.springframework.session.events.SessionDestroyedEvent;
  *
  *     {@literal @Bean}
  *     public MapSessionRepository sessionRepository() {
- *         return new MapSessionRepository();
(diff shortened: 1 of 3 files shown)


spring-session-core/src/main/java/org/springframework/session/config/annotation/web/http/EnableSpringHttpSession.java
```diff
@@ -40,7 +40,7 @@
  *
  *     {@literal @Bean}
  *     public MapSessionRepository sessionRepository() {
- *         return new MapSessionRepository();
+ *         return new MapSessionRepository(new HashMap<>());
```
eleftherias: Please change this to use `new ConcurrentHashMap<>()`.

Wrent: thanks, I fixed that and also found similar case in `EnableSpringWebSession`

eleftherias:
Thanks for the PR @Wrent! This is now merged into master.

Issue #1424, opened 2019-05-14
SessionRepositoryFilter's getRequestedSession is called unnecessarily.


hosoitk:
When Spring Session used in the application was upgraded from **1.3.1.RELEASE** to **2.1.3.RELEASE**, performance was verified and it was confirmed that the CPU utilization has increased by about 10%.
When profiling was performed using the function `-agentlib: hprof = cpu = samples`, it was confirmed that the execution time of `RedisOperationsSessionRepository.getSession (String, boolean)` has increased after version upgrade.

The increase in the number of executions of `RedisOperationsSessionRepository.getSession (String, boolean)` is due to the following two updates.

## Update point 1
After Spring Session **1.3.4.RELEASE**, with implementation of `SessionCommittingRequestDispatcher` which introduced the change that when screen transition is made to a JSP page containing `jsp: include`, `commitSession` is executed each time `jsp: include` is executed.

## Update point 2
`getRequestedSessionId` has been modified since Spring Session **2.0.0.RELEASE**, `getRequestedSessionId` has been changed to check whether the session ID obtained from cookie by `RedisOperationsSessionRepository.findById` exists in the session.

`commitSession` calls `getRequestedSessionId`, `getRequestedSessionId` calls `RedisOperationsSessionRepository.findById` via `getRequestedSession`.
Because `RedisOperationsSessionRepository.findById` has become is a process that calls `RedisOperationsSessionRepository.getSession(String, boolean)`, increase in the number of executions of `commitSession` leads to an increase of `RedisOperationsSessionRepository.getSession (String, boolean)`

This is `getRequestedSessionId` but due to the modification from Spring Session **2.1.1.RELEASE** onwards, `getRequestedSession` can be executed to update `requestedSessionId` only when `requestedSessionId` is null, by caching the `requestedSessionId`.

However, at the same time in `clearRequestedSessionCache` a process is added to assign null to `requestedSessionId` as well and in `commitSession` in order to execute `clearRequestedSessionCache` before executing `requestedSessionId` it has become necessary to execute `getRequestedSession`.

https://github.com/spring-projects/spring-session/blob/c5fc4b57ad2eb369564927e4fd13850c0f5a94f2/spring-session-core/src/main/java/org/springframework/session/web/http/SessionRepositoryFilter.java#L357-L363

[...]

PR #1578, opened 2020-02-12
Use cached requestedSession info if available

oxc:
Prevent fetching the session multiple times when checking whether the
sessionId has to be sent to the client, by comparing values before
clearing the requestedSessionCache.

Resolves: #1424

### DIFF
--- a/spring-session-core/src/main/java/org/springframework/session/web/http/SessionRepositoryFilter.java
+++ b/spring-session-core/src/main/java/org/springframework/session/web/http/SessionRepositoryFilter.java
@@ -221,10 +221,11 @@ public class SessionRepositoryFilter<S extends Session> extends OncePerRequestFi
 			}
 			else {
 				S session = wrappedSession.getSession();
+				String sessionId = session.getId();
+				boolean isRequestedSession = isRequestedSessionIdValid() && sessionId.equals(getRequestedSessionId());
 				clearRequestedSessionCache();
 				SessionRepositoryFilter.this.sessionRepository.save(session);
-				String sessionId = session.getId();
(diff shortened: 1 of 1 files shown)


oxc:
Apologies for the failed tests. Why is the findById invocation after session commit expected, or rather, what purpose does it serve? Am I missing something here?
Is it legitimate to reduce the number of expected invocations, since this is exactly what this pull request is trying to fix.

spring-session-core/src/main/java/org/springframework/session/web/http/SessionRepositoryFilter.java
```diff
@@ -221,10 +221,11 @@ private void commitSession() {
 			}
 			else {
 				S session = wrappedSession.getSession();
+				String sessionId = session.getId();
+				boolean isRequestedSession = isRequestedSessionIdValid() && sessionId.equals(getRequestedSessionId());
 				clearRequestedSessionCache();
 				SessionRepositoryFilter.this.sessionRepository.save(session);
-				String sessionId = session.getId();
```
oxc: Is the sessionId allowed to change on save? In that case, the comparison should still be done after save, but requestedSessionId has to be retrieved before the clearRequestedSessionCache()

quaff:
This PR should be superseded by https://github.com/spring-projects/spring-session/pull/2194

bianjp:
> This PR should be superseded by #2194

But #2194 was not backported to 2.7.x

oxc:
This has been open for over 3 years. Feel free to open if anybody really thinks this is going to get merged somewhere.

PR #1579, opened 2020-02-13
Fixes the duplicate index name in the example SQL script

v-antech:
The change is in the comments for the JdbcIndexedSessionRepository.

The script does not works, as it is, as the index names collide, so the change was required.


eleftherias:
Thanks for the PR @v-antech! This is now merged into master.

Issue #1321, opened 2019-01-21
Spring Session Oracle jdbc temp space issue ORA-01652


chkrishna411:
We are using spring-session with jdbc backed by Oracle database. 

we are using  1.3.4 version of spring-session.
    compile "org.springframework.session:spring-session-jdbc:1.3.4.RELEASE"
    compile "org.springframework.session:spring-session:1.3.4.RELEASE"

We are seeing temp space being utilized heavily by spring-session and it's keep on increasing gradually. It's not cleaning up automatically and after 2 to 3 days, it has consumed the entire TEMP space and our app was going down.

After researching on oracle database side,  it looks like the following outer join query is the one which is using temporary LOB.

  private static final String GET_SESSION_QUERY =
              "SELECT S.SESSION_ID, S.CREATION_TIME, S.LAST_ACCESS_TIME, S.MAX_INACTIVE_INTERVAL, SA.ATTRIBUTE_NAME, SA.ATTRIBUTE_BYTES " +
                        "FROM %TABLE_NAME% S " +
                        "LEFT OUTER JOIN %TABLE_NAME%_ATTRIBUTES SA ON S.SESSION_ID = SA.SESSION_ID " +
                        "WHERE S.SESSION_ID = ?";
 
Once after killing those sessions directly from database, entire temp space usage came down to 0.

**Failed Logs in the application:**
org.springframework.jdbc.UncategorizedSQLException: PreparedStatementCallback; uncategorized SQLException for SQL [UPDATE SPRING_SESSION_ATTRIBUTES SET ATTRIBUTE_BYTES = ? WHERE SESSION_ID = ? AND ATTRIBUTE_NAME = ?]; SQL state [72000]; error code [1652]; ORA-01652: unable to extend temp segment by 128 in tablespace TEMP
; nested exception is java.sql.SQLException: ORA-01652: unable to extend temp segment by **128 in tablespace TEMP**

        at org.springframework.jdbc.support.AbstractFallbackSQLExceptionTranslator.translate(AbstractFallbackSQLExceptionTranslator.java:90)
        at org.springframework.jdbc.support.AbstractFallbackSQLExceptionTranslator.translate(AbstractFallbackSQLExceptionTranslator.java:82)
        at org.springframework.jdbc.support.AbstractFallbackSQLExceptionTranslator.translate(AbstractFallbackSQLExceptionTranslator.java:82)
[...]

PR #1654, opened 2020-06-29
Add try-with-resources to methods to insert BLOB

k-tamura:
## Problem

If using Oracle DB as a session store of Spring Session and idle connection timeout is not set (or connections do not become idle for a long  time), there is a possibility that Oracle's TEMP space is increased and finally `SQLException` (`ORA-01652`) is thrown. 

*Note:* This problem has already reported on #1321.

## Description

When inserting a record to `SPRING_SESSION_ATTRIBUTES` table (which has Blob column), the following steps are executed:

1. Create a temporary LOB in Oracle's TEMP space
2. Insert a record to `SPRING_SESSION_ATTRIBUTES` table

The created temporary LOB is freed when the connection (which created the temporary LOB) is closed by idle timeout or Spring application is restarted, but these events do not always occur.

## Solution

So the following step should be added just after step 2.
 
3. Free the temporary LOB

To achieve this, try-with-resources statement should be added.

This problem is caused by leaks of calling `TemporaryLobCreator.close()`. `TemporaryLobCreator` implements `LobCreator` (which extends `Closeable`) and has `close()` (which frees temporary LOB), but `TemporaryLobCreator` is created without try-with-resources statement. Therefor `close()` is not called and temporary LOB is not freed.

## Steps to reproduce:

1. Create a Web application which uses Oracle DB as a session store of Spring Session.
2. Access to the application (for example, 1000 times by JMeter).
3. Run the SQL:
   ```
   select * from V$TEMPORARY_LOBS;
   ```
If you keep accessing to the application, `ORA-01652` occurs and `SQLException` is thrown.

## Result before applying my commit (CACHE_LOBS will increase):

   ```
      SID CACHE_LOBS NOCACHE_LOBS ABSTRACT_LOBS     CON_ID
---------- ---------- ------------ ------------- ----------
        18	  371		 0	       0	  0
        23	    0		 0	       0	  0
        25	  229		 0	       0	  0
       173	  144		 0	       0	  0
       178	  723		 0	       0	  0
       346	  533		 0	       0	  0
   ```

## Result after applying my commit (CACHE_LOBS will not increase):

   ```
       SID CACHE_LOBS NOCACHE_LOBS ABSTRACT_LOBS     CON_ID
---------- ---------- ------------ ------------- ----------
        23	    0		 0	       0	  0
   ```

### DIFF
--- a/spring-session-jdbc/src/main/java/org/springframework/session/jdbc/JdbcIndexedSessionRepository.java
+++ b/spring-session-jdbc/src/main/java/org/springframework/session/jdbc/JdbcIndexedSessionRepository.java
@@ -44,6 +44,7 @@ import org.springframework.jdbc.core.BatchPreparedStatementSetter;
 import org.springframework.jdbc.core.JdbcOperations;
 import org.springframework.jdbc.core.ResultSetExtractor;
 import org.springframework.jdbc.support.lob.DefaultLobHandler;
+import org.springframework.jdbc.support.lob.LobCreator;
 import org.springframework.jdbc.support.lob.LobHandler;
 import org.springframework.session.DelegatingIndexResolver;
 import org.springframework.session.FindByIndexNameSessionRepository;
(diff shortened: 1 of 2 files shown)


k-tamura:
Hi @rwinch ,

Thank you for your reviewing. 
My changes will only affect Oracle's TEMP space. So I think it is difficult to create test cases to confirm it. Do you have a good idea?

rwinch:
I'm guessing you could add a unit test to ensure that the temporary LOG is freed?

k-tamura:
Some test methods in `JdbcIndexedSessionRepositoryTests` calls two methods which is changed by me. However, it seems that variable `this.jdbcOperations` does not include information which indicates the temporary LOG is freed even if `createTemporaryLob` flag is `true`. So I think a unit test to ensure it cannot be added.

Screenshot from 2020-07-02 22-07-35

rwinch:
I think you would need to verify that the lob that was created (which would be a mock also) is then closed proeprly.

k-tamura:
I am sure that freeing temporary LOB cannot be tested. If not using Oracle DB, we need to test if `TemporaryLobCreator.close()` is called while calling `this.repository.save(session);` but, as far as I investigated, it is not possible on this testing framework.

k-tamura:
@rwinch

Do you believe it is possible? If yes, could you give me some more specific advice?

rwinch:
I think it should be possible by injecting a mock LobHandler that returns a mock LobCreator. Then you would verify the LobCreator is closed.

k-tamura:
@rwinch 

Thank you for your advice. I added a test method. Is this what you intended?

k-tamura:
Hi @rwinch 

It would be very helpful if you could comment my test case. I added it on 22 Jul. 

Best regards,

pivotal-issuemaster:
@k-tamura Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-session&pullRequestId=1654)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-session&pullRequestId=1654) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@k-tamura Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-session&pullRequestId=1654)!

k-tamura:
Hi @rwinch 

It would be very helpful if you could comment my test case. I added it on 22 Jul. 

Best regards,

yoshimi0227:
How's the review going?
I'd really appreciate it if this request could be merged !

k-tamura:
@rwinch @vpavic 

Is it difficult to include this commit in the next release?

rwinch:
We are already in the release process. Given this is a bug, we can get it in the next patch release (Dec 9th). That will be in time for Spring Boot's GA release.

eleftherias:
Thanks for the PR @k-tamura! This is merged into master.

PR #1657, opened 2020-07-16
Update OncePerRequestFilter to match with spring-web

bboyz269:
Call "getAlreadyFilteredAttributeName()" instead of direct usage if private "this.alreadyFilteredAttributeName"

Fix the error of the property value is null when proxied by CGLIB (bean from extending SessionRepositoryFilter)

### DIFF
--- a/spring-session-core/src/test/java/org/springframework/session/web/http/OncePerRequestFilterAopTests.java
+++ b/spring-session-core/src/test/java/org/springframework/session/web/http/OncePerRequestFilterAopTests.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2014-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.session.web.http;
+
+import org.aspectj.lang.annotation.AfterReturning;
+import org.aspectj.lang.annotation.Aspect;
+import org.junit.jupiter.api.Test;
+import org.mockito.Mockito;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.EnableAspectJAutoProxy;
+import org.springframework.mock.web.MockFilterChain;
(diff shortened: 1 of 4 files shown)


bboyz269:
I'm not sure what would be suffice for the test you mention but I'll look it up as soon as I get available time at a PC.On Jul 17, 2020 9:28 PM, Eleftheria Stein-Kousathana <notifications@github.com> wrote:
@eleftherias requested changes on this pull request.

Thanks for the PR @bboyz269!
Could you add a test demonstrating the bug?
Also, please add Closes gh-1658 to the commit message, so that the issue you reported is automatically closed when this is merged.

rwinch:
I think a unit test demonstrating that a subclass of OncePerRequestFilter that has overridden getAlreadyFilteredAttributeName will still work is sufficient.

bboyz269:
> I think a unit test demonstrating that a subclass of OncePerRequestFilter that has overridden getAlreadyFilteredAttributeName will still work is sufficient.

Not quite, it has to be a subclass being proxied by Spring AOP.

@eleftherias I add a new test class `OncePerRequestFilterAopTests`, would you check it out?

eleftherias:
Thanks for the updates @bboyz269!

In the original issue that you reported, gh-1658, was the class also being proxied by Spring AOP?

bboyz269:
> In the original issue that you reported, [gh-1658](https://github.com/spring-projects/spring-session/issues/1658), was the class also being proxied by Spring AOP?

Yes, it was being proxied by Spring AOP with CGLIb.

spring-session-core/src/test/java/org/springframework/session/web/http/OncePerRequestFilterAopTests.java
```diff
@@ -0,0 +1,70 @@
+/*
+ * Copyright 2014-2019 the original author or authors.
```
eleftherias, line 2: The copyright should go up to 2020.

spring-session-core/src/test/java/org/springframework/session/web/http/OncePerRequestFilterAopTests.java
```diff
@@ -0,0 +1,70 @@
...
+			return Mockito.mock(SessionRepository.class);
+		}
+
+		@Bean
+		public SessionRepositoryFilter filter() {
+			return new SessionRepositoryFilter(sessionRepository());
+		}
+
+		@AfterReturning("execution(* SessionRepositoryFilter.doFilterInternal(..))")
+		public void doInternalFilterPoincut() {
```
eleftherias, line 64: Small typo
```suggestion
		public void doInternalFilterPointcut() {
```

bboyz269: Fixed as suggested.

spring-session-core/src/test/java/org/springframework/session/web/http/OncePerRequestFilterAopTests.java
bboyz269: Fixed as suggested.

eleftherias:
Thanks for the PR @bboyz269! This is now merged into master.

Issue #1584, opened 2020-02-20
Spring Session Support for Hazelcast 4


mesutcelik:
Hazelcast 4 has recently been released and it is not currently supported.
related : https://stackoverflow.com/questions/60303477/spring-session-with-hazlecast-4

fisco-unimatic:
I've [forked the repo](https://github.com/fisco-unimatic/spring-session), upgraded the hazelcast dependencies, and fixed the code so that it compiles, but there's a failing integration test, `SessionEventHazelcastIndexedSessionRepositoryTests.updateMaxInactiveIntervalTest`. Hz is not firing an entry expired event after we update the time-to-live. (It does if I put breakpoints in, or a thread sleep between updating the ttl on the map and updating the session with the entry processor.)

eleftherias:
@fisco-unimatic Thanks for looking at this issue. 
However, this is schedule for the Spring Session 3.x release because it is a major version upgrade to one of our dependencies.
Since Spring Session 3.x does not have a scheduled release date, I would advise pausing any work on this issue, since it may be a while before it can be merged and any code changes may become stale by then.

rwinch:
Another option is ensuring that we can work with both versions of Hazelcast.

mesutcelik:
Spring Sessions 2.4+ can support Hazelcast 4+. 
Spring Sessions 2.3 and earlier can still support Hazelcast 3.x. 

This can be documented so that everyone can be aware of compatibility requirement.

rwinch:
We can add support for Hazelcast 4, but we must continue to support Hazelcast 3. Our tests should verify that it continues to work as well.

vpavic:
> Spring Sessions 2.4+ can support Hazelcast 4+.

@mesutcelik Spring Session primarily has to align with Spring Boot in terms of managed dependency versions, and since Spring Boot manages those taking semver into account, it's unlikely that upgrade to Hazelcast 4.x will happen before Spring Boot 3.0

It is unfortunate that Hazelcast opted to do so many breaking changes in 4.0 release without ensuring users have a smooth transition period by keeping the old APIs in deprecated form for some time.

mesutcelik:
@vpavic Thanks for the feedback.

A major release had to be done at some point. Every product does that but good news is that 3.12 will be actively supported and this support can be considered as deprecation policy. [See active branches for 3.x](https://github.com/hazelcast/hazelcast/branches/active)

[...]

Issue #1131, opened 2018-07-18
Improve Hazelcast serialization


vpavic:
`HazelcastSessionRepository` currently relies on standard Java serialization, which isn't the ideal option performance wise. We should look into leveraging Hazelcast's serialization mechanism like `DataSerializable` which should yield a nice performance improvement.

oxc:
In order to fix the issue mentioned in #1319 (client-server topology with custom-class session entries), the solution must ensure that no session entries are deserialized on the server-side. 
Switching to another form of serialization will probably not provide this without changing what is stored inside the session map.

enozcan:
As @oxc pointed out
> the solution must ensure that no session entries are deserialized on the server-side.

This can be partially done via Portable serialization (more information here at #1671). No user code deployments or classes on the server's classpath would be needed if Portable applies. Moreover, it will have a considerable impact on the query performance of `findByPrincipal` case. But this will require to break backward compatibility. It's suggested to have this in the next major version (https://github.com/spring-projects/spring-session/pull/1671#issuecomment-692561036) and I will be happy to implement this. 

Leaving the comment here to make this a sort of reminder for the issue.

PR #1671, opened 2020-08-07
Add optional Hazelcast session serializer

enozcan:
Related: #1131 

Provides custom serializer optimized for Hazelcast rather than native Java serialization.

#### Backward Compatibility

This is an optional serializer. If Hazelcast instances are not explicitly configured to use, the serialization mechanism will remain unchanged. However, it's not possible to use this serializer for an existing session store serialized without this.

#### Requirements to Use

In embedded mode, it can be configured directly just before a member starts. In client mode, the serializer - and hence the MapSession, need to be present on member's classpath and configured. Both Hazelcast client and member must use the same serialization mechanism for consistency.

#### About Hazelcast's Portable Serialization

Portable (not used in this PR) is another way to serialize the objects in Hazelcast way. It also fits well with the indexed repository. Because in the current scenario when a query is made to fetch sessions by the principal, the objects on the distributed map are deserialized to extract the value, then serialized and transferred to the caller side, and then deserialized again. Portable serialization, on the other hand, allows map values to be queried without deserializing the whole value but only the necessary one - in this case, `PRINCIPAL_NAME_ATTRIBUTE`.

However, Portable requires a serialization interface to be implemented for the objects to be stored on the map. In that case - as MapSession is final, a wrapper for sessions could be good to go. But this would require some changes in HazelcastIndexedSessionRepository and more importantly, would break the backward compatibility for the sessions serialized in the old fashion. So I think it's better for now to keep this serialization logic optional - maybe until 2.4.0.

#### More on Portable

Also, with the next versions of Hazelcast, it will be possible to run entry processors (in spring-session-hazelcast, sessions are updated via entry processors on the member side, without the need for fetching and then putting the value on the caller side) without including a dependency on members. When this is the case, it will prevent users to be required to adding their classes when using client/server topology. The current version uses user code deployment for this, which is actually not the best way of. (related: #1584)

### DIFF
--- a/spring-session-hazelcast/src/main/java/org/springframework/session/hazelcast/HazelcastSessionSerializer.java
+++ b/spring-session-hazelcast/src/main/java/org/springframework/session/hazelcast/HazelcastSessionSerializer.java
@@ -1,0 +1,87 @@
+package org.springframework.session.hazelcast;
+
+import com.hazelcast.nio.ObjectDataInput;
+import com.hazelcast.nio.ObjectDataOutput;
+import com.hazelcast.nio.serialization.StreamSerializer;
+import org.springframework.session.MapSession;
+
+import java.io.EOFException;
+import java.io.IOException;
+import java.time.Duration;
+import java.time.Instant;
+
+public class HazelcastSessionSerializer implements StreamSerializer<MapSession> {
+
+	private static final int SERIALIZER_TYPE_ID = 12345;
+
+	@Override
+	public void write(ObjectDataOutput out, MapSession session) throws IOException {
+		out.writeUTF(session.getOriginalId());
+		out.writeUTF(session.getId());
+		writeInstant(out, session.getCreationTime());
+		writeInstant(out, session.getLastAccessedTime());
+		writeDuration(out, session.getMaxInactiveInterval());
+		for (String attrName : session.getAttributeNames()) {
+			Object attrValue = session.getAttribute(attrName);
+			if (attrValue != null) {
+				out.writeUTF(attrName);
+				out.writeObject(attrValue);
+			}
+		}
+	}
+
+	private void writeInstant(ObjectDataOutput out, Instant instant) throws IOException {
+		out.writeLong(instant.getEpochSecond());
+		out.writeInt(instant.getNano());
+	}
+
+	private void writeDuration(ObjectDataOutput out, Duration duration) throws IOException {
+		out.writeLong(duration.getSeconds());
+		out.writeInt(duration.getNano());
+	}
+
+	@Override
+	public MapSession read(ObjectDataInput in) throws IOException {
+		String originalId = in.readUTF();
+		MapSession cached = new MapSession(originalId);
+		cached.setId(in.readUTF());
+		cached.setCreationTime(readInstant(in));
+		cached.setLastAccessedTime(readInstant(in));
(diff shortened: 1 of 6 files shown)


eleftherias:
@enozcan Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-session&pullRequestId=1671)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-session&pullRequestId=1671) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@enozcan Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-session&pullRequestId=1671)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-session&pullRequestId=1671) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@enozcan Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-session&pullRequestId=1671)!

spring-session-hazelcast/src/main/java/org/springframework/session/hazelcast/HazelcastSessionSerializer.java
```diff
@@ -0,0 +1,158 @@
...
+ *
+ * SerializerConfig serializerConfig = new SerializerConfig();
+ * serializerConfig.setImplementation(new HazelcastSessionSerializer()).setTypeClass(MapSession.class);
+ * clientConfig.getSerializationConfig().addSerializerConfig(serializerConfig);
+ *
+ * HazelcastInstance hazelcastClient = HazelcastClient.newHazelcastClient(clientConfig);
+ * </pre>
+ *
+ * @author Enes Ozcan
+ */
```
eleftherias, line 82: Please add `@since 2.4.0`

eleftherias (changes requested):
Thanks for the PR @enozcan!
I have left one small comment inline and then this can be merged.

spring-session-hazelcast/src/main/java/org/springframework/session/hazelcast/HazelcastSessionSerializer.java
enozcan: Done

enozcan:
Thanks for your review @eleftherias. 

As I explained, the serialization issue for Hazelcast sessions could have been even more efficient via some breaking changes. But my main concern was to conform backward compatibility and hence made the serializer in this PR optional. However, now that I see you added Hazelcast 4 support as another module, I wonder if we can perform some changes in that Hazelcast 4 module which are not backward compatible at all - considering this is a newly introduced module. The only outcome that makes me worried is having two different types of modules for Hazelcast 3 and 4 might be somehow complicated to users. 

WDYT? I also wonder what the expected release date of 2.4.0 is.

eleftherias:
@enozcan 2.4.0 is scheduled for October 28, 2020.

Thank you for the suggestion.
For now, I would like to keep the Hazelcast 4 module as minimal as possible.
It is meant to make Hazelcast 4 useable, but any enhancements are better suited for the next major release (3.0).

3.0 isn't scheduled yet, but I will update https://github.com/spring-projects/spring-session/milestones when we have an estimated date.

PR #1676, opened 2020-08-28
Move Gradle enterprise cache secrets to top level of CI build

elliedori:
While updating the SpringSecurity build, I found a way to set env vars used by all steps at a higher level to prevent duplication. This PR simplifies our usage of the Gradle enterprise secrets by moving them to the top level of the build and enabling us to get rid of duplication.

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -7,6 +7,11 @@ on:
   schedule:
     - cron: '0 10 * * *' # Once per day at 10am UTC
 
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous-integration-workflow.yml
```diff
@@ -3,10 +3,15 @@ name: CI
 on:
   push:
     branches:
-      - master
+      - simplify-build-env-vars
```
eleftherias, line 6: Is this change meant to be part of this PR?

PR #1708, opened 2020-09-26
Delay allocating Strings for Redis keyspace notifications

theigl:
### Problem

Spring Session Redis allocates new Strings for `channel` and `body` of all incoming messages. In my case, Redis is used to store other session related data and `RedisIndexedSessionRepository` receives a lot of key expiration and deletion messages that it doesn't need to handle. Hundreds of megabytes of Strings are allocated unnecessarily every minute.

In such a case, it makes sense to delay allocating new Strings until we are sure the message should be handled by Spring session.

### Solution

This PR directly uses the message's `byte[]` to check if a message should be handled. String allocation is delayed until we know that the message is intended for Spring Session.

### DIFF
--- a/spring-session-data-redis/src/main/java/org/springframework/session/data/redis/RedisIndexedSessionRepository.java
+++ b/spring-session-data-redis/src/main/java/org/springframework/session/data/redis/RedisIndexedSessionRepository.java
@@ -18,6 +18,7 @@ package org.springframework.session.data.redis;
 
 import java.time.Duration;
 import java.time.Instant;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


theigl:
I also created a related PR in Spring Data Redis that can further reduce the allocation rate for messages that should be handled by Spring session:

https://github.com/spring-projects/spring-data-redis/pull/559

vpavic (approved):
This looks like a nice optimization to me @theigl. :+1:

theigl:
@vpavic: I have a version of this in production already and it saves a couple of hundred MB of string allocations a minute. Our Redis is not really busy so this might make even more of a difference for other users. 

Unfortunately, this is only half of the story because {{Message.getChannel()}} and {{Message.getBody()}} create new byte arrays. So filtering incoming messages is still not allocation free even after merging this PR. But it is a step in the right direction.

I'll ping the Spring Data guys to see if there is a chance my other PR will get merged. With both PRs applied, message filtering wouldn't allocate at all.

theigl:
@vpavic: I discussed the situation with @mp911de and there is not much that can be done to improve the situation on the Spring Data side. So please go ahead and merge this PR if it is OK with you.

PR #1710, opened 2020-10-01
Improve RedisSessionRepository-based sample configuration

vpavic:
This improves the `RedisSessionRepository`-based sample to showcase how to reuse Spring Boot's existing configuration support for Spring Session Redis so that the usual configuration properties can be used.


eleftherias:
Thanks @vpavic! This is now merged.

PR #1712, opened 2020-10-02
Harmonize Redis key namespace configurations

vpavic:
At present, the `RedisSessionRepository#setKeyNamespace` expects users to provide the trailing colon (`:`) character that is used as separator between namespace segments. This is unlike `RedisIndexedSessionRepository` and `ReactiveRedisSessionRepository` that apply the separator implicitly in their respective `#setRedisKeyNamespace` methods.

This PR harmonizes the Redis key namespaces configurations across all Redis-backed session repository implementations.

I've noticed this while applying the approach from #1710 in one project.
Ideally, this should be merged before #1710 which should then updated to pick up the new `#setRedisKeyNamespace` method and use it in the sample.

### DIFF
--- a/spring-session-data-redis/src/main/java/org/springframework/session/data/redis/RedisSessionRepository.java
+++ b/spring-session-data-redis/src/main/java/org/springframework/session/data/redis/RedisSessionRepository.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2019 the original author or authors.
+ * Copyright 2014-2020 the original author or authors.
  *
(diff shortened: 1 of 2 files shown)


eleftherias:
Thanks @vpavic!
I will merge this after the 2.4.0 release, in order to avoid API changes between RC1 and GA.

vpavic:
As labeled, this is IMO a bug and the change to address it is passive.

If you're that concerned about the API change (or an addition, to be more preciese) I'd suggest to do an `RC2` instead of postponing this for a full cycle.

Issue #838, opened 2017-07-25
Deadlock found when trying to get lock with SQL [UPDATE SPRING_SESSION SET LAST_ACCESS_TIME = ?, MAX_INACTIVE_INTERVAL = ?, PRINCIPAL_NAME = ? WHERE SESSION_ID = ?]; 


MichoFelipe:
Hi All,

We are having the following issue in our production environment.

Subject: ERROR: PreparedStatementCallback; SQL [UPDATE SPRING_SESSION SET LAST_ACCESS_TIME = ?, MAX_INACTIVE_INTERVAL = ?, PRINCIPAL_NAME = ? WHERE SESSION_ID = ?]; Deadlock found when trying to get lock; try restarting transaction; nested exception is com.mysq...

Thread's Attributes Found: 

DeadlockLoserDataAccessException Message: PreparedStatementCallback; SQL [UPDATE SPRING_SESSION SET LAST_ACCESS_TIME = ?, MAX_INACTIVE_INTERVAL = ?, PRINCIPAL_NAME = ? WHERE SESSION_ID = ?]; Deadlock found when trying to get lock; try restarting transaction; nested exception is com.mysql.jdbc.exceptions.jdbc4.MySQLTransactionRollbackException: Deadlock found when trying to get lock; try restarting transaction

org.springframework.jdbc.support.SQLErrorCodeSQLExceptionTranslator.doTranslate(SQLErrorCodeSQLExceptionTranslator.java:263)
org.springframework.jdbc.support.AbstractFallbackSQLExceptionTranslator.translate(AbstractFallbackSQLExceptionTranslator.java:73)
org.springframework.jdbc.core.JdbcTemplate.execute(JdbcTemplate.java:649)
org.springframework.jdbc.core.JdbcTemplate.update(JdbcTemplate.java:870)
org.springframework.jdbc.core.JdbcTemplate.update(JdbcTemplate.java:931)
org.springframework.session.jdbc.JdbcOperationsSessionRepository$2.doInTransactionWithoutResult(JdbcOperationsSessionRepository.java:318)
org.springframework.transaction.support.TransactionCallbackWithoutResult.doInTransaction(TransactionCallbackWithoutResult.java:34)
org.springframework.transaction.support.TransactionTemplate.execute(TransactionTemplate.java:133)
org.springframework.session.jdbc.JdbcOperationsSessionRepository.save(JdbcOperationsSessionRepository.java:314)
[...]

PR #1726, opened 2020-10-31
Provide database specific JdbcIndexedSessionRepository customizers

### LINKED ISSUE #1213: Concurrent requests adding the same session attribute result in duplicate key constraint violation
We are using Spring session 2.0.5.RELEASE. We are using JDBC session persistence.
We have migrated those applications that were previously using container in memory session (Wildfly/Undertow). The code has stayed as-is.

We have html pages that fire up multiple calls in AJAX simultaneously to a controller.
In the controller we have code that set an attribute into the session.

The issue is that multiple threads of the application server are busy in parallel treating those AJAX calls, the first thread will insert the new attribute in the database. The other threads are then supposed to call an update of the attribute in the database because the attribute already exists. But the issue is that the other threads have not detected (yet) that the attribute already exists and try to do an insert as well resulting in a duplicate key exception.

I must say that as a workaround we have patched the application to do a sort of initialisation AJAX call at page load and this initialisation just set the attribute with a blank value, this allows the attribute row to be created and when further calls are being made in parallel, the JDBC implementation of Spring session sees that only updates are needed.

However I wanted to report this in order to support those use cases natively (without workarounds) using JDBC persistence.

### LINKED ISSUE #1550: Deadlock in SQL Server when INSERT INTO SPRING_SESSION_ATTRIBUTES by JdbcOperationsSessionRepository
I have a deadlock in SQL Server.

**LOG**
```
Forwarding to error page from request [/login] due to exception [PreparedStatementCallback; SQL [INSERT INTO SPRING_SESSION_ATTRIBUTES(SESSION_PRIMARY_ID, ATTRIBUTE_NAME, ATTRIBUTE_BYTES) SELECT PRIMARY_ID, ?, ? FROM SPRING_SESSION WHERE SESSION_ID = ?]; ***deadlock error message from sql server is omitted because Japanese language***
org.springframework.dao.DeadlockLoserDataAccessException: PreparedStatementCallback; SQL [INSERT INTO SPRING_SESSION_ATTRIBUTES(SESSION_PRIMARY_ID, ATTRIBUTE_NAME, ATTRIBUTE_BYTES) SELECT PRIMARY_ID, ?, ? FROM SPRING_SESSION WHERE SESSION_ID = ?]; ***deadlock error message from sql server is omitted because Japanese language***
	at org.springframework.jdbc.support.SQLErrorCodeSQLExceptionTranslator.doTranslate(SQLErrorCodeSQLExceptionTranslator.java:266)
	at org.springframework.jdbc.support.AbstractFallbackSQLExceptionTranslator.translate(AbstractFallbackSQLExceptionTranslator.java:72)
	at org.springframework.jdbc.core.JdbcTemplate.translateException(JdbcTemplate.java:1442)
	at org.springframework.jdbc.core.JdbcTemplate.execute(JdbcTemplate.java:632)
	at org.springframework.jdbc.core.JdbcTemplate.update(JdbcTemplate.java:861)
	at org.springframework.jdbc.core.JdbcTemplate.update(JdbcTemplate.java:916)
	at org.springframework.session.jdbc.JdbcOperationsSessionRepository.insertSessionAttributes(JdbcOperationsSessionRepository.java:548)
	at org.springframework.session.jdbc.JdbcOperationsSessionRepository.access$300(JdbcOperationsSessionRepository.java:135)
```

I read all the previous issues about this problem.

As #1083 @TorosyanV said:

we need to always keep order to avoid deadlock on parent and child relationship table's operation.

like as below:

parnet = SPRING_SESSION
child  = SPRING_SESSION_ATTRIBUTES

insert parent first, then child
update parent first, then child
delete parent first, then child
select parent first, then child

I looked at the JdbcOperationsSessionRepository code, 
and I understood it looks like keeping transaction order.

but major problem here is DELETE_SESSIONS_BY_EXPIRY_TIME_QUERY is depending on each relational database's CASCADE ON DELETE algorithm
which might not be the same order deleting parent and child.
and also, we are difficult position to know or change relational database's CASCADE ON DELETE algorithm.

It is better not rely on to database's CASCADE ON DELETE.
I wish, Spring framework will ensure the every session table's transaction operation order by own.

vpavic:
This PR addresses several `JdbcIndexedSessionRepository` concerns.

This first commit of this PR provides `JdbcIndexedSessionRepository` customizers for the following SQL dialects:

- PostgreSQL
- MySQL (also used by MariaDB)
- SQL Server
- IBM DB2
- Oracle

These customizers are intended to address the concurrency issues occurring on insert of new session attribute by applying SQL dialect specific SQL upsert/merge statement instead of a generic insert.

This closes #1213.

The second commit optimizes the SQL statement used to insert session attribute so that it doesn't perform a nested select statement.

This closes #1550.

Final commit reduces the JDBC integration tests to only single (latest) version per RDBMS vendor, due to a growing number of integration tests. Additionally, the configuration of most containers is simplified due to improved defaults within the Testcontainers library.


vpavic:
@jkuipers, since you also went at depth with this issue, I'd appreciate your feedback on the changes proposed by this PR?

jkuipers:
> @jkuipers, since you also went at depth with this issue, I'd appreciate your feedback on the changes proposed by this PR?

Hey Vedran,

I just had a look (sorry, wasn't able to get to this earlier). I like the approach that there can now be different SQL for different DBMSs. I also noticed a commit that makes the code use the session's primary key again instead of a nested select by session ID, which I think is the right approach.
What is still missing is that the new SQL to upsert the attributes is only used for session _creation_ but not _updating_. That means that two requests adding the same session attribute still causes errors. I believe that this code should use the same query: it's an upsert, so it works for both new AND existing sessions. That fixes another long-standing problem, then. 
I ran my tests from https://github.com/jkuipers/spring-session-jdbc-sql-server-bugs against the latest snapshot, and they clearly show that the session update is still an issue. 

On another note: the new MS SQL Server SQL is I think semantically equivalent to what I came up with in my own fork. Both my version and the one currently on master still sporadically cause deadlocks in SQL Server, so even though it's a big improvement over the old code, the problem isn't solved 100% yet. 
If I come up with a fix for that I'll send it a merge request.

vpavic:
Thanks for following up on this @jkuipers, good to hear you're seeing these changes as an improvement in practice. :+1:

Regarding the remaining reasons for deadlocks, they are IMO due to expired session clean-up job. See #838 - this is something that to my knowledge affects MySQL/MariaDB and SQL Server users.

I opened #1739 yesterday to take a closer look at improving the clean-up job, so feel free to track that one as well.

jkuipers:
Using my tests, after updating to a snapshot of Spring Session and registering the SqlServerJdbcIndexedSessionRepositoryCustomizer as a bean, I can trigger the deadlock by simply creating several sessions in parallel still. I therefore do not think that this is related to the cleanup, I don't think that cleanup ever runs during these tests. 
My customer, who's using my fork that includes a MERGE statement for upserting the session attributes that's almost the same as what you have on master right now, ran into this problem in production yesterday when their application was under heavy load.

I've examined the problem yesterday evening. It appears that for some reason, the combination of INSERTing a new session and then using the MERGE statement to insert the corresponding session attributes can trigger deadlocks in MS SQL Server. I tried several things first like making the primary key clustered, changing isolation levels and some other things that didn't help.
Then I updated the code to only use the MERGE statement when updating the attributes of an _existing_ session, and to use a simple INSERT for attributes of a newly created session as there's no actual need for an upsert there.
For my test, this completely eliminated the problem.

vpavic:
That's interesting, thanks for sharing @jkuipers.
I'll reopen #1550 to take another look at that. Limiting the use of upsert to existing sessions does make sense.

dlamblin:
Pardon which release was this merged into? I'd like to try the MySQL specific customizations, but can't quite see whether a version is out that includes them.

I'm confused a bit because while I used the `mavenBom 'org.springframework.session:spring-session-bom:2020.0.3'` I'm not exactly sure how that relates to Spring or Spring Boot releases much less spring-session releases. Thank you.

--
For context, when I'm using spring-session-jdbc with an aws aurora mysql 5.7 single instance RDS which is connected by a `spring.datasource.url=jdbc:mariadb://`... string to my Spring Boot 2.3.9.RELEASE app, I can't complete a single request to a static resource, nor a simple controller that returns "hello"; they hang and eventually time out.

The clean up job deadlocks, like #838, but requests timeout for a lock wait on updating the SPRING_SESSION too, which I think is related to the clean up deadlocking.
```
HTTP Status 500 – Internal Server Error
Type Exception Report

Message PreparedStatementCallback; SQL [UPDATE SPRING_SESSION SET SESSION_ID = ?, LAST_ACCESS_TIME = ?, MAX_INACTIVE_INTERVAL = ?, EXPIRY_TIME = ?, PRINCIPAL_NAME = ? WHERE PRIMARY_ID = ?]; (conn=426748) Lock wait timeout exceeded; try restarting transaction; nested exception is java.sql.SQLTransientConnectionException: (conn=426748) Lock wait timeout exceeded; try restarting transaction

Description The server encountered an unexpected condition that prevented it from fulfilling the request.

Exception

org.springframework.dao.CannotAcquireLockException: PreparedStatementCallback; SQL [UPDATE SPRING_SESSION SET SESSION_ID = ?, LAST_ACCESS_TIME = ?, MAX_INACTIVE_INTERVAL = ?, EXPIRY_TIME = ?, PRINCIPAL_NAME = ? WHERE PRIMARY_ID = ?]; (conn=426748) Lock wait timeout exceeded; try restarting transaction; nested exception is java.sql.SQLTransientConnectionException: (conn=426748) Lock wait timeout exceeded; try restarting transaction
	org.springframework.jdbc.support.SQLErrorCodeSQLExceptionTranslator.doTranslate(SQLErrorCodeSQLExceptionTranslator.java:267)
	org.springframework.jdbc.support.AbstractFallbackSQLExceptionTranslator.translate(AbstractFallbackSQLExceptionTranslator.java:72)
	org.springframework.jdbc.core.JdbcTemplate.translateException(JdbcTemplate.java:1443)
	org.springframework.jdbc.core.JdbcTemplate.execute(JdbcTemplate.java:633)
	org.springframework.jdbc.core.JdbcTemplate.update(JdbcTemplate.java:862)
	org.springframework.jdbc.core.JdbcTemplate.update(JdbcTemplate.java:917)
	org.springframework.session.jdbc.JdbcIndexedSessionRepository$JdbcSession.lambda$save$10(JdbcIndexedSessionRepository.java:807)
	org.springframework.transaction.support.TransactionOperations.lambda$executeWithoutResult$0(TransactionOperations.java:68)
	org.springframework.transaction.support.TransactionTemplate.execute(TransactionTemplate.java:140)
	org.springframework.transaction.support.TransactionOperations.executeWithoutResult(TransactionOperations.java:67)
	org.springframework.session.jdbc.JdbcIndexedSessionRepository$JdbcSession.save(JdbcIndexedSessionRepository.java:802)
	org.springframework.session.jdbc.JdbcIndexedSessionRepository$JdbcSession.access$200(JdbcIndexedSessionRepository.java:634)
	org.springframework.session.jdbc.JdbcIndexedSessionRepository.save(JdbcIndexedSessionRepository.java:410)
	org.springframework.session.jdbc.JdbcIndexedSessionRepository.save(JdbcIndexedSessionRepository.java:131)
	org.springframework.session.web.http.SessionRepositoryFilter$SessionRepositoryRequestWrapper.commitSession(SessionRepositoryFilter.java:225)
	org.springframework.session.web.http.SessionRepositoryFilter$SessionRepositoryRequestWrapper.access$100(SessionRepositoryFilter.java:192)
	org.springframework.session.web.http.SessionRepositoryFilter.doFilterInternal(SessionRepositoryFilter.java:144)
	org.springframework.session.web.http.OncePerRequestFilter.doFilter(OncePerRequestFilter.java:82)
	org.springframework.boot.actuate.metrics.web.servlet.WebMvcMetricsFilter.doFilterInternal(WebMvcMetricsFilter.java:93)
	org.springframework.web.filter.OncePerRequestFilter.doFilter(OncePerRequestFilter.java:119)
	org.springframework.web.filter.CharacterEncodingFilter.doFilterInternal(CharacterEncodingFilter.java:201)
	org.springframework.web.filter.OncePerRequestFilter.doFilter(OncePerRequestFilter.java:119)
	org.springframework.web.filter.OncePerRequestFilter.doFilter(OncePerRequestFilter.java:103)
```

vpavic:
@dlamblin This will be a part of upcoming Spring Session `2.5.0`, which in turn is a part of Spring Session BOM `2021.0.0`.

PR #1732, opened 2020-11-21
[master] Avoid extraneous call to SessionRespository.findById(...) in SessionRepositoryRequestWrapper.commitSession()

### LINKED ISSUE #1731: SessionRepositoryFilter.SessionRepositoryRequestWrapper.commitSession() triggers extraneous call to SessionRespository.findById(...)
**Describe the bug**
For a valid session, SessionRepositoryFilter.SessionRepositoryRequestWrapper.commitSession() does the following:

1. Clears cached values for references for requestedSessionCache, requestedSession, and requestedSessionId.
1. Saves the state of the session
1. If the requested session ID is not valid, or if the ID of the session changed from the requested session ID, then the session ID is sent to the client

However, since #1 already cleared the requested session cached values #3 requires a completely redundant call to SessionRepository.findById(...).  Depending on the implementation, this can be unnecessarily expensive.

**To Reproduce**
Use a tracer to detect calls to SessionRepository.findById(...).  For a requested session, this is triggered twice, once at the beginning of the request, and again after the session is saved.

**Expected behavior**
SessionRepository.findById(...) should be considered a potentially expensive operation, and only triggered when necessary.

pferraro:
Resolves #1731 
Clearing the requested session cache *after* we've written the session ID to the response avoid the need for a redundant lookup of the session.

### DIFF
--- a/spring-session-core/src/main/java/org/springframework/session/web/http/SessionRepositoryFilter.java
+++ b/spring-session-core/src/main/java/org/springframework/session/web/http/SessionRepositoryFilter.java
@@ -221,11 +221,16 @@ public class SessionRepositoryFilter<S extends Session> extends OncePerRequestFi
 			}
 			else {
 				S session = wrappedSession.getSession();
-				clearRequestedSessionCache();
-				SessionRepositoryFilter.this.sessionRepository.save(session);
 				String sessionId = session.getId();
-				if (!isRequestedSessionIdValid() || !sessionId.equals(getRequestedSessionId())) {
-					SessionRepositoryFilter.this.httpSessionIdResolver.setSessionId(this, this.response, sessionId);
+				try {
+					boolean sendSessionId = !isRequestedSessionIdValid() || !sessionId.equals(getRequestedSessionId());
+					SessionRepositoryFilter.this.sessionRepository.save(session);
+					if (sendSessionId) {
+						SessionRepositoryFilter.this.httpSessionIdResolver.setSessionId(this, this.response, sessionId);
+					}
+				}
+				finally {
+					clearRequestedSessionCache();
 				}
 			}
 		}
--- a/spring-session-core/src/test/java/org/springframework/session/web/http/SessionRepositoryFilterTests.java
+++ b/spring-session-core/src/test/java/org/springframework/session/web/http/SessionRepositoryFilterTests.java
@@ -1313,8 +1313,8 @@ class SessionRepositoryFilterTests {
 			}
 		});
 
-		// 3 invocations expected: initial resolution, after invalidation, after commit
-		verify(sessionRepository, times(3)).findById(eq(session.getId()));
+		// 3 invocations expected: initial resolution, after invalidation
+		verify(sessionRepository, times(2)).findById(eq(session.getId()));
 		verify(sessionRepository).deleteById(eq(session.getId()));
 		verify(sessionRepository).createSession();
 		verify(sessionRepository).save(any());


pivotal-issuemaster:
@pferraro Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-session&pullRequestId=1732)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-session&pullRequestId=1732) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@pferraro Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-session&pullRequestId=1732)!

eleftherias (changes requested):
Thanks for the PR @pferraro! 

Could you add `Closes gh-1731` to the commit message so that it will automatically close the issue you created?

pferraro:
@eleftherias Done.

vpavic:
The lookup might seem redundant when one considers only what happens within the scope of this method (or even JVM), but have in mind that one of the core challenges Spring Session solves is session management in a cluster.

With that in mind, this needs to be looked at having in mind the scenarios with concurrent requests. The session id validity check that guards the invocation of `HttpSessionIdResolver#setSessionId` needs to know the up-to-date state, otherwise I think we could expose ourself to a number of subtle issue in concurrent scenarios, especially since things have been this way since the early days of Spring Session. Hence the unit test that clearly describes that 3 invocations are expected.

An important bit of context here is that `SessionRepository#save` has historically been implemented in such a way to be as performant as possible, specifically in this case it's relevant that invoking `#save` on a session that was already deleted from the store by some other request will result in a no-op. This is done to prevent reads within the implementation of `#save`. I believe that different implementations of `SessionRepository` aren't consistent here but at the same time there isn't any specific contract here to adhere to. That's why I opened #1277 a while ago.

So the change proposed here IMO shouldn't be done without considering a more substantial review of the whole arrangement.

pferraro:
> The lookup might seem redundant when one considers only what happens within the scope of this method (or even JVM), but have in mind that one of the core challenges Spring Session solves is session management in a cluster.

Concurrent requests to a given Session by multiple cluster members are forbidden by section 7.7.2 of the Servlet specification:

    Within an application marked as distributable, all requests that are part of a session must be handled by one JVM at a time.

Thus, we only need to be concerned with concurrent requests within a single JVM.

> With that in mind, this needs to be looked at having in mind the scenarios with concurrent requests. The session id validity check that guards the invocation of HttpSessionIdResolver#setSessionId needs to know the up-to-date state...

Assuming that the state referenced by a Session returned by SessionRepository.findById(...) for a given request is, by definition, up-to-date until the corresponding Session#save completes, can we not simply reorder this logic such that we perform the session id validity check *before* Session#save?

> I believe that different implementations of SessionRepository aren't consistent here but at the same time there isn't any specific contract here to adhere to. That's why I opened #1277 a while ago.

It seems to me that a SessionRepository implementation that persists an invalid session is incorrect, no?

vpavic:
> Thus, we only need to be concerned with concurrent requests for a given Session within a single JVM.

I'm afraid the excerpt you're quoting isn't really applicable to modern environments, as nowadays we typically have a round robin based load-balancers that distribute requests evenly between the nodes in a cluster. Spring Session is exactly here to address that scenario so that users don't have to resort to anti-patterns like sticky sessions (which is something Servlet spec seems to suggest there).

> can we not simply reorder this logic such that we perform the session id validity check before Session#save?

Logically, the concern I'm expressing here would still be present, because `#save` precedes `#setSessionId`. Historically, we've seen many issues reported due to concurrency scenarios and that's why I'm hesitant to make any changes that could lead to subtle issues. We need to review the overall approach here and that looks like a `3.0` target to me.

> It seems to me that a SessionRepository implementation that persists an invalid session is incorrect, no?

Yes, I agree with you there and that's exactly why I opened #1277.

pferraro:
> Logically, the concern I'm expressing here would still be present, because #save precedes #setSessionId. Historically, we've seen many issues reported due to concurrency scenarios and that's why I'm hesitant to make any changes that could lead to subtle issues. We need to review the overall approach here and that looks like a 3.0 target to me.

Can you describe a scenario where the proposed change is problematic - that is not already problematic using the current logic?

lnxmad:
Any update on this?

quaff:
@pferraro @lnxmad Could you try #1909?

rwinch:
Closing in favor of gh-1909

PR #1733, opened 2020-11-21
[2.4.x] Avoid extraneous call to SessionRespository.findById(...) in SessionRepositoryRequestWrapper.commitSession()

### LINKED ISSUE #1731: SessionRepositoryFilter.SessionRepositoryRequestWrapper.commitSession() triggers extraneous call to SessionRespository.findById(...)

pferraro:
Backports https://github.com/spring-projects/spring-session/pull/1732 to 2.4.x
Resolves #1731

### DIFF
--- a/spring-session-core/src/main/java/org/springframework/session/web/http/SessionRepositoryFilter.java
+++ b/spring-session-core/src/main/java/org/springframework/session/web/http/SessionRepositoryFilter.java
@@ -221,11 +221,15 @@ public class SessionRepositoryFilter<S extends Session> extends OncePerRequestFi
 			}
 			else {
 				S session = wrappedSession.getSession();
-				clearRequestedSessionCache();
-				SessionRepositoryFilter.this.sessionRepository.save(session);
-				String sessionId = session.getId();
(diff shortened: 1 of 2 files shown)


eleftherias:
Thanks for the suggestion @pferraro. I consider this an enhancement, rather than a bug, so I will not be backporting it 2.4.x.

PR #1801, opened 2021-03-14
Principal Name Index is not stored in Hazelcast Session when flush mode is immediate

s-londono:
We found out that the Principal Name attribute is not being stored in the Hazelcast Session Repository. This attribute should be written to the session's map, so that we could efficiently search sessions by principal name. 

This is the value used by the index _Hazelcast4IndexedSessionRepository.PRINCIPAL_NAME_ATTRIBUTE_, which is suggested to be set on the session's map.

As far as we can see the attribute is properly set, but the save method doesn't save it into the Hazelcast map, because the delta that would instruct the entry processor to do so is missing.

### DIFF
--- a/spring-session-hazelcast/hazelcast4/src/main/java/org/springframework/session/hazelcast/Hazelcast4IndexedSessionRepository.java
+++ b/spring-session-hazelcast/hazelcast4/src/main/java/org/springframework/session/hazelcast/Hazelcast4IndexedSessionRepository.java
@@ -450,6 +450,7 @@ public class Hazelcast4IndexedSessionRepository
 						.resolveIndexesFor(this);
 				String principal = (attributeValue != null) ? indexes.get(PRINCIPAL_NAME_INDEX_NAME) : null;
 				this.delegate.setAttribute(PRINCIPAL_NAME_INDEX_NAME, principal);
+				this.delta.put(PRINCIPAL_NAME_INDEX_NAME, principal);
 			}
 			flushImmediateIfNecessary();
 		}
--- a/spring-session-hazelcast/hazelcast4/src/main/java/org/springframework/session/hazelcast/Hazelcast4IndexedSessionRepository.java
+++ b/spring-session-hazelcast/hazelcast4/src/main/java/org/springframework/session/hazelcast/Hazelcast4IndexedSessionRepository.java
@@ -450,7 +450,7 @@ public class Hazelcast4IndexedSessionRepository
 						.resolveIndexesFor(this);
 				String principal = (attributeValue != null) ? indexes.get(PRINCIPAL_NAME_INDEX_NAME) : null;
 				this.delegate.setAttribute(PRINCIPAL_NAME_INDEX_NAME, principal);
-				this.delta.put(PRINCIPAL_NAME_INDEX_NAME, principal);
+//				this.delta.put(PRINCIPAL_NAME_INDEX_NAME, principal);
 			}
 			flushImmediateIfNecessary();
 		}
--- a/spring-session-hazelcast/hazelcast4/src/main/java/org/springframework/session/hazelcast/Hazelcast4IndexedSessionRepository.java
+++ b/spring-session-hazelcast/hazelcast4/src/main/java/org/springframework/session/hazelcast/Hazelcast4IndexedSessionRepository.java
@@ -450,7 +450,7 @@ public class Hazelcast4IndexedSessionRepository
 						.resolveIndexesFor(this);
 				String principal = (attributeValue != null) ? indexes.get(PRINCIPAL_NAME_INDEX_NAME) : null;
 				this.delegate.setAttribute(PRINCIPAL_NAME_INDEX_NAME, principal);
-//				this.delta.put(PRINCIPAL_NAME_INDEX_NAME, principal);
+				this.delta.put(PRINCIPAL_NAME_INDEX_NAME, principal);
 			}
 			flushImmediateIfNecessary();
 		}
--- a/spring-session-hazelcast/hazelcast4/src/test/java/org/springframework/session/hazelcast/Hazelcast4IndexedSessionRepositoryWithInstanceTests.java
+++ b/spring-session-hazelcast/hazelcast4/src/test/java/org/springframework/session/hazelcast/Hazelcast4IndexedSessionRepositoryWithInstanceTests.java
@@ -1,0 +1,72 @@
+package org.springframework.session.hazelcast;
+
+import com.hazelcast.config.Config;
+import com.hazelcast.core.Hazelcast;
+import com.hazelcast.core.HazelcastInstance;
+import com.hazelcast.map.IMap;
+import org.assertj.core.api.Condition;
+import org.junit.Before;
+import org.junit.Test;
+import org.springframework.security.authentication.UsernamePasswordAuthenticationToken;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.authority.AuthorityUtils;
+import org.springframework.security.core.context.SecurityContextImpl;
+import org.springframework.session.FlushMode;
+import org.springframework.session.MapSession;
+import org.springframework.session.SaveMode;
+import org.springframework.session.hazelcast.Hazelcast4IndexedSessionRepository.HazelcastSession;
+
+import java.util.Arrays;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class Hazelcast4IndexedSessionRepositoryWithInstanceTests {
+
+	private static final String SPRING_SECURITY_CONTEXT = "SPRING_SECURITY_CONTEXT";
+
+	private static final String SESSION_MAP_NAME = "spring:session:sessions";
+
+	private HazelcastInstance hzInstance;
+
+	private Hazelcast4IndexedSessionRepository repository;
+
+	@Before
+	public void initialize() {
+		this.hzInstance = Hazelcast.newHazelcastInstance(new Config());
+
+		this.repository = new Hazelcast4IndexedSessionRepository(this.hzInstance);
+		this.repository.setFlushMode(FlushMode.IMMEDIATE);
+		this.repository.setSaveMode(SaveMode.ALWAYS);
+		this.repository.setSessionMapName(SESSION_MAP_NAME);
+		this.repository.init();
+	}
+
+	@Test
+	public void hazelcastInstance() {
+		assertThat(hzInstance).withFailMessage("HazelcastInstance must not be null").isNotNull();
+	}
+
+	@Test
+	public void findByIndexNameAndIndexValuePrincipalIndexNameFound() {
+		String principal = "username";
+		Authentication authentication = new UsernamePasswordAuthenticationToken(principal, "notused",
+				AuthorityUtils.createAuthorityList("ROLE_USER"));
+		SecurityContextImpl securityCtx = new SecurityContextImpl(authentication);
+
+		HazelcastSession newSession = repository.createSession();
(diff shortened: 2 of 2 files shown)


pivotal-issuemaster:
@s-londono Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-session&pullRequestId=1801)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-session&pullRequestId=1801) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@s-londono Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-session&pullRequestId=1801)!

eleftherias:
Thanks for reaching out @s-londono.

I'm confused by the description and the proposed change.
In the issue description you mention setting `Hazelcast4IndexedSessionRepository.PRINCIPAL_NAME_ATTRIBUTE` while in the suggested change you are setting `FindByIndexNameSessionRepository.PRINCIPAL_NAME_INDEX_NAME`.

Furthermore, the `FindByIndexNameSessionRepository.PRINCIPAL_NAME_INDEX_NAME` attribute is currently being populated. 
If it is not being populated in your application, please let me know and we can try to figure out why that is the case for you.

s-londono:
Thank you so much for your reply @eleftherias.

I apologize, my description of the issue was convoluted. Here is a more concise version:
In the code we seee that the `FindByIndexNameSessionRepository.PRINCIPAL_NAME_INDEX_NAME` is set. 
I ran a debugging session and noticed that although said attribute was set, it didn't get written into the Hazelcast sessions map.

I suspect that what is missing, is the delta that tells the Hazelcast EntryProcessor to store the attribute into the sessions map. That's the only line I added in my suggested change:
`this.delta.put(PRINCIPAL_NAME_INDEX_NAME, principal);`

As far as I can see, the deltas are processed by method `Hazelcast4IndexedSessionRepository.save`.
By the way, is this working well for you? (I searched as much as I could for reports about any similar issues, but found nothing).

eleftherias:
Thanks for the clarification @s-londono.
In the example that I tested I can see the `FindByIndexNameSessionRepository.PRINCIPAL_NAME_INDEX_NAME` attribute set in the Hazelcast session map, when the session is created.
Perhaps your use-case is different than what I am using.

If you could provide a [minimal sample](https://stackoverflow.com/help/minimal-reproducible-example) that reproduces the issue you're seeing that would be a great starting point.

s-londono:
Big thanks to you @eleftherias! Sure, I will try to reproduce the case I witnessed in a minimal sample.

s-londono:
Hi @eleftherias. I managed to reproduce the issue by implemementing a test: `org.springframework.session.hazelcast.Hazelcast4IndexedSessionRepositoryWithInstanceTests`. The trick here is to use an actual Hazelcast Instance to store the session created in the test. As you may notice, the test fails unless we uncomment line 453 at `Hazelcast4IndexedSessionRepository`, which adds the delta I think is missing. 

Here is a link to my fork, where you will find the code: [https://github.com/s-londono/spring-session](url)
Thank you!

eleftherias:
Thanks @s-londono.
Could you provide a running application that reproduces the issue?
I would like to see the issue in an end to end situation so I can figure out how it gets into that state.

eleftherias:
Hey @s-londono, have you been able to reproduce this in a running application?

s-londono:
Hi @eleftherias,
I'm sorry for the late reply. Originally I reproduced the error on my job's application, but unfortunately it would be very difficult to make that one visible to you. So I thought about creating a small Spring webapp that can be easily run in your environment and that clearly reveals the case. Unfortunately I haven't managed to find the time to do so. Hopefully I'll be able to work on it next weekend :)

alienisty:
Hi @eleftherias,
I manged to create a simple reproducing project [spring-session-sample-boot-hazelcast4-1801.zip](https://github.com/spring-projects/spring-session/files/6622127/spring-session-sample-boot-hazelcast4-1801.zip).
From my analysis there are two main problems:

1. When the principal attribute is set, the session should be flushed, otherwise, I suspect, the index doesn't get updates/populated.
2. The save method in Hazelcast4IndexedSessionRepository skips saving the attributes if the session is new.

In my opinion, flush mode should be immediate and save mode should be always for the Hazelcast repository, otherwise changes will not be immediately visible in all nodes of a cluster, and that is a problem in load balanced environments not using sticky sessions.

Problem number 2. is clearly a bug because the HazelcastSession constructor actually forces all attributes to be saved when a session is new:
```java
HazelcastSession(MapSession cached, boolean isNew) {
	this.delegate = cached;
	this.isNew = isNew;
	this.originalId = cached.getId();
	if (this.isNew /* <-- Note this */ || (Hazelcast4IndexedSessionRepository.this.saveMode == SaveMode.ALWAYS)) {
		getAttributeNames()
				.forEach((attributeName) -> this.delta.put(attributeName, cached.getAttribute(attributeName)));
	}
}
```
But in the save method we have instead:
```java
@Override
public void save(HazelcastSession session) {
        if (session.isNew) {
	        this.sessions.set(session.getId(), session.getDelegate(), session.getMaxInactiveInterval().getSeconds(),
			        TimeUnit.SECONDS);
        }
        else if (session.sessionIdChanged) {
	        this.sessions.delete(session.originalId);
	        session.originalId = session.getId();
	        this.sessions.set(session.getId(), session.getDelegate(), session.getMaxInactiveInterval().getSeconds(),
			        TimeUnit.SECONDS);
        }
        else if (session.hasChanges()) { // <-- This block is going to be skipped if session.isNew == true
	        Hazelcast4SessionUpdateEntryProcessor entryProcessor = new Hazelcast4SessionUpdateEntryProcessor();
	        if (session.lastAccessedTimeChanged) {
		        entryProcessor.setLastAccessedTime(session.getLastAccessedTime());
	        }
	        if (session.maxInactiveIntervalChanged) {
		        if (SUPPORTS_SET_TTL) {
			        updateTtl(session);
		        }
		        entryProcessor.setMaxInactiveInterval(session.getMaxInactiveInterval());
	        }
	        if (!session.delta.isEmpty()) {
		        entryProcessor.setDelta(new HashMap<>(session.delta));
	        }
	        this.sessions.executeOnKey(session.getId(), entryProcessor);
        }
        session.clearChangeFlags();
}
```
We need to also note that we should add the principal name index into the delta, otherwise it gets saved accidentally because normally set when the session is new:
```java
@Override
public void setAttribute(String attributeName, Object attributeValue) {
	this.delegate.setAttribute(attributeName, attributeValue);
	this.delta.put(attributeName, attributeValue);
	if (SPRING_SECURITY_CONTEXT.equals(attributeName)) {
		Map<String, String> indexes = Hazelcast4IndexedSessionRepository.this.indexResolver
				.resolveIndexesFor(this);
		String principal = (attributeValue != null) ? indexes.get(PRINCIPAL_NAME_INDEX_NAME) : null;
		this.delegate.setAttribute(PRINCIPAL_NAME_INDEX_NAME, principal);
                // Not adding the attribute to delta will prevent it to be saved if changed when !session.isNew
	}
	flushImmediateIfNecessary();
}
```
I hope this helps

eleftherias:
Thanks for the sample @alienisty. 
Could you also provide instructions on how to reproduce the issue using the sample?
When I tried running the application I was not able to see the issue.

I noticed you have a test, but it is important to see the issue in an end to end situation, in a functioning application.
Furthermore, the proposed fix does not fix the test. 
I suspect the test failure is caused by mocking and is unrelated to what was originally mentioned here.

alienisty:
Hi @eleftherias,
I wasn't able to reproduce it in an end to end situation in the sample project, the test is the only way I was able to reproduce the symptoms and my proposed fix does actually make the test pass. The mocking is only for the MVC, so it should not affect how spring session.
I also monkeypatched my application with the same fix and it all works as expected.
I'm going to try digging a bit more and see what I can do.

s-londono:
Hi @eleftherias and @alienisty,
I have some further information about how to reproduce this case.

The PRINCIPAL_NAME_INDEX_NAME attribute set on the sessions stored in Hazelcast, gets lost only when we have set the FlushMode to IMMEDIATE in the HazelcastHttpSessionConfig. For instance, if we define the following HazelcastHttpSessionConfig:
```java
@Configuration
public class HazelcastHttpSessionConfig extends HazelcastHttpSessionConfiguration {
  private static final int MAX_INACTIVE_INTERVAL_IN_SECONDS = 28800;

  @Autowired
  public HazelcastHttpSessionConfig() {
    // Customize the Session Repository
    setFlushMode(FlushMode.IMMEDIATE);
    setSaveMode(SaveMode.ALWAYS);
    setSessionMapName("spring:session:sessions");
    setMaxInactiveIntervalInSeconds(MAX_INACTIVE_INTERVAL_IN_SECONDS);
  }
}
```
The PRINCIPAL_NAME_INDEX_NAME attribute set by org.springframework.session.hazelcast.HazelcastSession, at the setAttribute method will not actually be saved. According to our analysis and tests, this issue would be solved by putting a delta on the session as follows:
```java
@Override
public void setAttribute(String attributeName, Object attributeValue) {
	this.delegate.setAttribute(attributeName, attributeValue);
	this.delta.put(attributeName, attributeValue);
	if (SPRING_SECURITY_CONTEXT.equals(attributeName)) {
		Map<String, String> indexes = Hazelcast4IndexedSessionRepository.this.indexResolver
				.resolveIndexesFor(this);
		String principal = (attributeValue != null) ? indexes.get(PRINCIPAL_NAME_INDEX_NAME) : null;
		this.delegate.setAttribute(PRINCIPAL_NAME_INDEX_NAME, principal);
    // We believe adding this delta would prevent the PRINCIPAL_NAME_INDEX_NAME from getting lost 
		// when HazelcastHttpSessionConfig's flushMode is set to FlushMode.IMMEDIATE
		this.delta.put(PRINCIPAL_NAME_INDEX_NAME, principal);
	}
	flushImmediateIfNecessary();
}
```
In the demo app attached (thanks @alienisty, it's based on yours):
[spring-session-sample-boot-hazelcast4-1801-1.zip](https://github.com/spring-projects/spring-session/files/6989192/spring-session-sample-boot-hazelcast4-1801-1.zip) 
You wil find that we set Flush Mode to IMMEDIATE. Please run it and login with credentials user/password. This will create an Http Session, which will be stored in Hazelcast. The homepage lists all attributes set on the session and as you will notice, the PRINCIPAL_NAME_INDEX_NAME attribute will not be present (this is the scenario that should be fixed by the change we suggest in the Pull Request):
image

Note that if we remove the line setting Flush Mode to IMMEDIATE at sample.config.HazelcastHttpSessionConfig, the PRINCIPAL_NAME_INDEX_NAME will be stored as expected:
```java
@Configuration
public class HazelcastHttpSessionConfig extends HazelcastHttpSessionConfiguration {
  private static final int MAX_INACTIVE_INTERVAL_IN_SECONDS = 28800;

  @Autowired
  public HazelcastHttpSessionConfig() {
    // Customize the Session Repository
    //setFlushMode(FlushMode.IMMEDIATE);
    setSaveMode(SaveMode.ALWAYS);
    setSessionMapName("spring:session:sessions");
    setMaxInactiveIntervalInSeconds(MAX_INACTIVE_INTERVAL_IN_SECONDS);
  }
}
```
image

alienisty:
Hi @eleftherias, any news on this one considering there is now an end to end reproducing demo provided by @s-londono? I would really like to remove the mokey patching from my application.

eleftherias:
Thanks for the reminder @alienisty and thank you for the sample @s-londono. 
I have merged your proposed fix via 661ecaf and added an integration test via a64a11b.

PR #1864, opened 2021-06-03
Optimize the createDefaultCookieSerializer method of SpringHttpSessionConfiguration

weixsun:
If assemble `DefaultCookieSerializer` from `SessionCookieConfig`, we also need to consider the **Secure** and **HttpOnly** attributes of the cookie. Otherwise **Secure** will always be `null` and **HttpOnly** will always be the default value `true`.


eleftherias:
Thanks for reaching out @weixsun.
The current behavior is intentional.
See gh-1253 for more details.

weixsun:
@eleftherias Thanks for the explanation, so that's it, I get it.

PR #1900, opened 2021-08-23
Always set existing time to live on Hazelcast 4 map entry when Session is updated

### LINKED ISSUE #1899: Hazelcast4SessionUpdateEntryProcessor does not keep existing TTL of session map entry
**Describe the bug**
Hazelcast 4 has changed the way time-to-live on IMap entries are handled with hazelcast/hazelcast#10965. A set/put without a given TTL will reset the TTL of the entry to the value configured on the map or max value (it will not keep the existing TTL of the entry as Hazelcast 3 has done). That also applies to operations done on a MapEntry from within an EntryProcessor.

The `Hazelcast4SessionUpdateEntryProcessor` uses similar same logic as the old one for Hazelcast 3: it only specifies a TTL when updating the entry in the Hazelcast map if the TTL value of the session was actually changed.

> if (this.maxInactiveInterval != null) {
                ((ExtendedMapEntry)entry).setValue(value, this.maxInactiveInterval.getSeconds(), TimeUnit.SECONDS);
            } else {
                entry.setValue(value);
            }

The field `this.maxInactiveInterval` on the EntryProcessor is only set if someone changed the TTL of an existing Session (via `HazelcastSession.setMaxInactiveInterval(Duration interval)`). If that value was not changed, this field will be null and therefore `entry.setValue(value)` will be called. Which will reset any previously set TTL on this MapEntry back to the map config/max value, effectively removing any session timeout configured via Spring properties.

**To Reproduce**
Create a Hazelcast4IndexedSessionRepositry and set the `defaultMaxInactiveInterval` on it to e.g. 60 seconds. Create a new session and save it. This session now correctly has a TTL of 60 seconds in Hazelcast's IMap.

Now change anything on this session (e.g. add a new attribute) except for its maxInactiveInterval. On `Hazelcast4IndexedSessionRepository.save()` a new `Hazelcast4SessionUpdateEntryProcessor` will be created but its `maxInactiveInterval` will be null because `session.maxInactiveIntervalChanged` is false. This leads to the above behavior. If you check the EntryView of the session in the IMap after the EntryProcessor was executed you will see that the TTL is now set to max value (not 60 seconds anymore).

Note: the HazelcastSession object is still expired by the code checks in `Hazelcast4IndexedSessionRepositry.findById()` because the maxInactiveInterval value stored within the session object is not cleared, of course. But Hazelcast itself will not expire the map entry anymore in the expected amount of time. So if there is no further read access to this session after its timeout interval has passed, it will never get evicted and therefore no expiration event will be fired by Hazelcast. That breaks all listeners that depend on these events (e.g. auditing).

**Expected behavior**
Given the new behavior of Hazelcast 4, I would expect that the `Hazelcast4SessionUpdateEntryProcessor` always calls `((ExtendedMapEntry)entry).setValue(value, ttl.getSeconds(), TimeUnit.SECONDS);` to update the map entry if `Hazelcast4IndexedSessionRepositry .defaultMaxInactiveInterval` is not null.

It should use `this.maxInactiveInterval` as ttl if there was a change to this value, otherwise it should use the previous value from `session.getMaxInactiveInterval()` as ttl instead.

AndreasKasparek:
Due to the changes in Hazelcast 4, an EntryProcessor always has to define the intended time-to-live (ttl) when setting the value on an ExtendedMapEntry, otherwise any existing ttl on such an entry would be reset to the default time configured on the map itself or to Integer.MAX.

The `Hazelcast4SessionUpdateEntryProcessor` was adjusted to use either the updated or the existing `maxInactiveInterval` of the session being processed, so that the resulting Hazelcast IMap entry always has the right ttl for this session instance. In the previous implementation the ttl was only set if the `maxInactiveInterval` was actually updated on the session.

This commit fixes #1899

### DIFF
--- a/spring-session-hazelcast/hazelcast4/src/integration-test/java/org/springframework/session/hazelcast/AbstractHazelcast4IndexedSessionRepositoryITests.java
+++ b/spring-session-hazelcast/hazelcast4/src/integration-test/java/org/springframework/session/hazelcast/AbstractHazelcast4IndexedSessionRepositoryITests.java
@@ -16,6 +16,9 @@
 
 package org.springframework.session.hazelcast;
 
+import java.time.Duration;
+import java.time.Instant;
+
 import com.hazelcast.core.HazelcastInstance;
 import com.hazelcast.map.IMap;
 import org.junit.jupiter.api.Test;
@@ -196,6 +199,53 @@ abstract class AbstractHazelcast4IndexedSessionRepositoryITests {
 		assertThat(this.repository.findById(sessionId)).isNotNull();
 
 		this.repository.deleteById(sessionId);
+	}
+
+	@Test
+	void createAndUpdateSessionWhileKeepingOriginalTimeToLiveConfiguredOnRepository() {
+		final Duration defaultSessionTimeout = Duration.ofSeconds(1800);
+
+		final IMap<String, MapSession> hazelcastMap = this.hazelcastInstance
+				.getMap(Hazelcast4IndexedSessionRepository.DEFAULT_SESSION_MAP_NAME);
+
+		HazelcastSession session = this.repository.createSession();
+		String sessionId = session.getId();
+		this.repository.save(session);
+
+		assertThat(session.getMaxInactiveInterval()).isEqualTo(defaultSessionTimeout);
+		assertThat(hazelcastMap.getEntryView(sessionId).getTtl()).isEqualTo(defaultSessionTimeout.toMillis());
+
+		session = this.repository.findById(sessionId);
+		session.setLastAccessedTime(Instant.now());
+		this.repository.save(session);
+
+		session = this.repository.findById(sessionId);
+		assertThat(session.getMaxInactiveInterval()).isEqualTo(defaultSessionTimeout);
+		assertThat(hazelcastMap.getEntryView(sessionId).getTtl()).isEqualTo(defaultSessionTimeout.toMillis());
+	}
+
+	@Test
+	void createAndUpdateSessionWhileKeepingTimeToLiveSetOnSession() {
+		final Duration individualSessionTimeout = Duration.ofSeconds(23);
+
+		final IMap<String, MapSession> hazelcastMap = this.hazelcastInstance
+				.getMap(Hazelcast4IndexedSessionRepository.DEFAULT_SESSION_MAP_NAME);
+
+		HazelcastSession session = this.repository.createSession();
+		session.setMaxInactiveInterval(individualSessionTimeout);
+		String sessionId = session.getId();
+		this.repository.save(session);
+
+		assertThat(session.getMaxInactiveInterval()).isEqualTo(individualSessionTimeout);
+		assertThat(hazelcastMap.getEntryView(sessionId).getTtl()).isEqualTo(individualSessionTimeout.toMillis());
+
+		session = this.repository.findById(sessionId);
+		session.setAttribute("attribute", "value");
+		this.repository.save(session);
+
+		session = this.repository.findById(sessionId);
+		assertThat(session.getMaxInactiveInterval()).isEqualTo(individualSessionTimeout);
+		assertThat(hazelcastMap.getEntryView(sessionId).getTtl()).isEqualTo(individualSessionTimeout.toMillis());
 	}
 
 }
--- a/spring-session-hazelcast/hazelcast4/src/main/java/org/springframework/session/hazelcast/Hazelcast4SessionUpdateEntryProcessor.java
+++ b/spring-session-hazelcast/hazelcast4/src/main/java/org/springframework/session/hazelcast/Hazelcast4SessionUpdateEntryProcessor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2020 the original author or authors.
+ * Copyright 2014-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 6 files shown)


spring-session-hazelcast/hazelcast4/src/main/java/org/springframework/session/hazelcast/Hazelcast4SessionUpdateEntryProcessor.java
```diff
@@ -63,9 +63,9 @@ public Object process(Map.Entry<String, MapSession> entry) {
...
 		}
-		if (this.maxInactiveInterval != null) {
-			((ExtendedMapEntry<String, MapSession>) entry).setValue(value, this.maxInactiveInterval.getSeconds(),
-					TimeUnit.SECONDS);
+		final Duration timeToLive = value.getMaxInactiveInterval();
+		if (timeToLive != null) {
+			((ExtendedMapEntry<String, MapSession>) entry).setValue(value, timeToLive.getSeconds(), TimeUnit.SECONDS);
 		}
 		else {
 			entry.setValue(value);
```
AndreasKasparek, line 71: I actually think that this else-branch might also not be needed. At least the `Hazelcast4IndexedSessionRepository` always uses the `session.getMaxInactiveInterval()` during a `save()` to set the Hazelcast map entry if the session is new or the sessionId has changed. That looks like the assumption is that this duration is never null.

The assumption is basically true given that both the `HazelcastHttpSessionConfiguration` and the `MapSession` both have a non-null default. The configuration/properties only allow to set the session timeout as an integer, while the repository accepts an Integer but it uses the MapSession default if the Integer is null. However, the `HazelcastSession.setMaxInactiveInterval(Duration)` allows to set null. This is this only place where a null can actually happen and it would most probably break the `save()` method of the repository (I assume that Hazelcast throws an exception if the given ttl is null). So the best solution would be to add a null check to the `setMaxInactiveInterval(Duration)` method and remove the else-branch here in the EntryProcessor.

AndreasKasparek:
@eleftherias I cleaned up a bit more in the code than I originally had planned but I think the usage of the `maxInactiveInterval` property within the session repository and entry processor was not consistent. I also removed the explicit `setTtl` call because it seemed to be superfluous now that we are forced to give the time-to-live with each update of the entry processor anyway.

I saw that a previous build had failed. I ran the `format` Gradle command locally in my IDE and it succeeded, but there were no additional changes made to my files. I've never used Gradle before, so no idea if I did something wrong. In case the next build fails again with that error, I would appreciate a hint how to fix it on my end. Thank you very much!

spring-session-hazelcast/hazelcast4/src/test/java/org/springframework/session/hazelcast/Hazelcast4SessionUpdateEntryProcessorTest.java
```diff
@@ -0,0 +1,126 @@
...
+import org.springframework.session.MapSession;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.BDDMockito.given;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+
+class Hazelcast4SessionUpdateEntryProcessorTest {
+
+	private Hazelcast4SessionUpdateEntryProcessor sut;
```
eleftherias, line 37: We don't use the SUT naming convention in this project, I would prefer this variable be name `processor`, or something similar, for consistency.

spring-session-hazelcast/hazelcast4/src/integration-test/java/org/springframework/session/hazelcast/AbstractHazelcast4IndexedSessionRepositoryITests.java
```diff
@@ -198,4 +201,51 @@ void createSessionWithSecurityContextAndFindById() {
 		this.repository.deleteById(sessionId);
 	}
 
+	@Test // gh-1899
```
eleftherias, line 204: We typically add a comment referencing the GitHub issue when it is not immediately obvious what the purpose of the test is. I think this test is clear enough that we don't have to point to the associated issue.
```suggestion
	@Test
```

spring-session-hazelcast/hazelcast4/src/integration-test/java/org/springframework/session/hazelcast/AbstractHazelcast4IndexedSessionRepositoryITests.java
```diff
@@ -198,4 +201,51 @@ void createSessionWithSecurityContextAndFindById() {
...
+		session = this.repository.findById(sessionId);
+		session.setLastAccessedTime(Instant.now());
+		this.repository.save(session);
+
+		session = this.repository.findById(sessionId);
+		assertThat(session.getMaxInactiveInterval()).isEqualTo(defaultSessionTimeout);
+		assertThat(hazelcastMap.getEntryView(sessionId).getTtl()).isEqualTo(defaultSessionTimeout.toMillis());
+	}
+
+	@Test // gh-1899
```
eleftherias, line 227: No need to point to GitHub issue. See comment above.
```suggestion
	@Test
```

eleftherias (changes requested):
Thanks for the excellent analysis and PR @AndreasKasparek!

I agree with all of your proposed changes and I have added some minor feedback on the test classes below.

Please also update the license header to the current year in all modified classes.
```
Copyright 2014-2021 the original author or authors.
```
Please also update the commit message to follow the format
```
Short (50 chars or less) summary of changes

Closes gh-1899
```
This will make it more readable when looking through the commit history.

AndreasKasparek:
@eleftherias Requested changes pushed.

eleftherias:
Thanks @AndreasKasparek! 
This is now merged into the main branch.

To answer your question, it is recommended to squash the commits and force push.
You are still able to see the differences by clicking on "force-pushed" in the status update 
> AndreasKasparek force-pushed the AndreasKasparek:issue/hazelcast-session-processor-ttl branch from 23ae0a4 to a8dbe9f

AndreasKasparek:
@vpavic @eleftherias Is there anything I have to do for the backport or are you handling this yourselves (probably more efficient)? Thanks

vpavic:
Nothing is needed from the contributor side @AndreasKasparek, thanks!

PR #1909, opened 2021-08-31
Eliminate unnecessary sessionRepository::findById

quaff:
It is an obvious fix.


eleftherias:
Hi @quaff, this is a duplicate of gh-1732.
Please use the search functionality to check for existing issues before opening a new issue or pull request.

quaff:
@eleftherias This one is more simple and safe than that one.

quaff:
@vpavic Could you take a look at this?

vpavic:
After a closer review, we've decided to do this change in `3.0.0-RC1` as it is an elegant way of removing that extra round trip to session store and fetch the session. Thanks for the contribution @quaff.

PR #1988, opened 2021-12-28
Memory leak in Index for principal `null`

### LINKED ISSUE #1987: Memory leak in Index for principal `null`
**Describe the bug**

When a user lands at `/admin`, Spring generates an anonymous session for such user and redirects him to the login page.

When a user passes authentication process, `RedisIndexedSessionRepository` persists the session identifier in two indexes:
* `session:index:org.springframework.session.FindByIndexNameSessionRepository.PRINCIPAL_NAME_INDEX_NAME:null`
* `session:index:org.springframework.session.FindByIndexNameSessionRepository.PRINCIPAL_NAME_INDEX_NAME:principalUsername`

The issue is that `session:index:org.springframework.session.FindByIndexNameSessionRepository.PRINCIPAL_NAME_INDEX_NAME:null` never cleanup: I see hundreds of identifiers under that index.

**To Reproduce**

1. Start [the sample](https://github.com/spring-projects/spring-session/tree/main/spring-session-samples/spring-session-sample-boot-redis).
2. Sign-in using `user`/`password` credentials.
3. Click on the logout button.
4. Check the index, you will see the record under the key `spring:session:index:org.springframework.session.FindByIndexNameSessionRepository.PRINCIPAL_NAME_INDEX_NAME:null` that will never cleanup. 

**Expected behavior**

Do not persist session under `null` principal, or clean the identifier when the session changes its principal.

**Sample**

The issue can be reproduced using the sample you have in the repository: https://github.com/spring-projects/spring-session/tree/main/spring-session-samples/spring-session-sample-boot-redis.

ruslanys:
I believe this should be a solution for the reported issue.

The explanation is the following. There are only two places where we add `sessionId` into an index: [first](https://github.com/spring-projects/spring-session/blob/main/spring-session-data-redis/src/main/java/org/springframework/session/data/redis/RedisIndexedSessionRepository.java#L826), [second](https://github.com/spring-projects/spring-session/blob/main/spring-session-data-redis/src/main/java/org/springframework/session/data/redis/RedisIndexedSessionRepository.java#L864).

And the first one is under the validation: [if (principal != null)](https://github.com/spring-projects/spring-session/blob/main/spring-session-data-redis/src/main/java/org/springframework/session/data/redis/RedisIndexedSessionRepository.java#L824).
So, we do not create an index if a principal is `null`.

When a session changes its ID, we try to change it in the index as well.
But we do not check here that the principal is null. So, we do not check if the index exists.

In the end, instead of replacing sessionId in Index, we create a new one. And this leads to a memory leak.

Close #1987

### DIFF
--- a/spring-session-data-redis/src/integration-test/java/org/springframework/session/data/redis/RedisIndexedSessionRepositoryITests.java
+++ b/spring-session-data-redis/src/integration-test/java/org/springframework/session/data/redis/RedisIndexedSessionRepositoryITests.java
@@ -527,6 +527,30 @@ class RedisIndexedSessionRepositoryITests extends AbstractRedisITests {
 		assertThat(findByPrincipalName.keySet()).containsOnly(changeSessionId);
 	}
 
+	@Test // gh-1987
+	void changeSessionIdWhenPrincipalNameChangesFromNullThenIndexShouldNotBeCreated() {
+		String principalName = null;
+		String principalNameChanged = "findByChangedPrincipalName" + UUID.randomUUID();
+		RedisSession toSave = this.repository.createSession();
+		toSave.setAttribute(INDEX_NAME, principalName);
+
+		this.repository.save(toSave);
+
+		RedisSession findById = this.repository.findById(toSave.getId());
+		String changeSessionId = findById.changeSessionId();
+		findById.setAttribute(INDEX_NAME, principalNameChanged);
+		this.repository.save(findById);
+
+		Map<String, RedisSession> findByPrincipalName = this.repository.findByIndexNameAndIndexValue(INDEX_NAME,
+				principalName);
+		assertThat(findByPrincipalName).isEmpty();
+
+		findByPrincipalName = this.repository.findByIndexNameAndIndexValue(INDEX_NAME, principalNameChanged);
+
+		assertThat(findByPrincipalName).hasSize(1);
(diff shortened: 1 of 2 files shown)


ruslanys:
I found that this PR is the reason of the mistake: https://github.com/spring-projects/spring-session/pull/1941

quaff:
Verified this PR will fix unexpected key `spring:session:index:org.springframework.session.FindByIndexNameSessionRepository.PRINCIPAL_NAME_INDEX_NAME:null` creation.

eleftherias (changes requested):
Thanks for the PR @ruslanys! 
The changes look good to me.
Could you please squash the commits into one, and add a descriptive commit message, for example
```
Fix memory leak with null principal in Redis

Closes gh-1987
```

eleftherias:
Thanks @ruslanys! 
I have merged this into the main branch and I will also be backporting the fix to the 2.6.x branch, making it available in 2.6.2.

ruslanys:
Thank you @eleftherias !

jebeaudet:
@eleftherias Since this regression also reached the 2.5.x branch, could you backport it there too please? I can open a PR if you prefer.

eleftherias:
Thanks for pointing that out @jebeaudet. I will backport it to 2.5.x as well.

PR #2040, opened 2022-03-03
Fix bug in JDBC `SaveMode.ON_GET_ATTRIBUTE`

felixscheinost:
This fixes a bug in the `JDBC` implementation of `SaveMode.ON_GET_ATTRIBUTE`.

I included a testcase that triggers the bug.

The steps to reproduce the bug are:

  - Use `SaveMode.ON_GET_ATTRIBUTE`
  - Have an existing session
  - Do a `setAttribute` of a new attribute
  - Immediately after, before `save`, do a `getAttribute` of the same attribute
    - `ON_GET_ATTRIBUTE` causes this attribute to be set to  `DeltaValue.UPDATED`
    - **FIX**: Set to `UPDATED` iff not already set to `ADDED`
 - Later `save` tries to just `UPDATE` the new value which fails silently

### DIFF
--- a/spring-session-jdbc/src/main/java/org/springframework/session/jdbc/JdbcIndexedSessionRepository.java
+++ b/spring-session-jdbc/src/main/java/org/springframework/session/jdbc/JdbcIndexedSessionRepository.java
@@ -728,7 +728,8 @@ public class JdbcIndexedSessionRepository
 			T attributeValue = supplier.get();
 			if (attributeValue != null
 					&& JdbcIndexedSessionRepository.this.saveMode.equals(SaveMode.ON_GET_ATTRIBUTE)) {
-				this.delta.put(attributeName, DeltaValue.UPDATED);
+				this.delta.merge(attributeName, DeltaValue.UPDATED, (oldDeltaValue,
+						deltaValue) -> (oldDeltaValue == DeltaValue.ADDED) ? oldDeltaValue : deltaValue);
 			}
 			return attributeValue;
 		}
--- a/spring-session-jdbc/src/test/java/org/springframework/session/jdbc/JdbcIndexedSessionRepositoryTests.java
+++ b/spring-session-jdbc/src/test/java/org/springframework/session/jdbc/JdbcIndexedSessionRepositoryTests.java
@@ -656,6 +656,20 @@ class JdbcIndexedSessionRepositoryTests {
 	}
 
 	@Test
+	void saveWithSaveModeOnGetAttributeAndNewAttributeSetAndGet() {
+		this.repository.setSaveMode(SaveMode.ON_GET_ATTRIBUTE);
+		MapSession delegate = new MapSession();
+		delegate.setAttribute("attribute1", (Supplier<String>) () -> "value1");
+		JdbcSession session = this.repository.new JdbcSession(delegate, UUID.randomUUID().toString(), false);
+		session.setAttribute("attribute2", "value2");
+		session.getAttribute("attribute2");
+		this.repository.save(session);
+		verify(this.jdbcOperations, times(1)).update(startsWith("INSERT INTO SPRING_SESSION_ATTRIBUTES ("),
+				isA(PreparedStatementSetter.class));
+		verifyNoMoreInteractions(this.jdbcOperations);
+	}
+
+	@Test
 	void saveWithSaveModeAlways() {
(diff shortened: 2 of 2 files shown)


spring-session-jdbc/src/integration-test/java/org/springframework/session/jdbc/AbstractJdbcIndexedSessionRepositoryITests.java
```diff
@@ -819,6 +820,33 @@ void saveNewSessionAttributeConcurrently() {
 		}
 	}
 
+	@Test
+	void savesWhenSaveModeIsOnGetAttributeAndANewAttributeIsSetAndGet() {
```
eleftherias, line 824: This test would be better off in `JdbcIndexedSessionRepositoryTests`. Instead of checking the value of the attribute you can verify that it calls "INSERT INTO SPRING_SESSION_ATTRIBUTES ". The test would be similar to `saveWithSaveModeOnGetAttribute`.

eleftherias (changes requested):
Thanks for the PR @felixscheinost!

I've left a comment below. 

Please run `./gradlew format` once you've added the changes and squash the commits into a single one.

spring-session-jdbc/src/integration-test/java/org/springframework/session/jdbc/AbstractJdbcIndexedSessionRepositoryITests.java
felixscheinost: Thanks for your response!

I couldn't get the test to work inside `JdbcIndexedSessionRepositoryTests`. 

In the call to `this.repository.findById(initialSession.getId());` a NPE was thrown.
Because `List<JdbcSession> sessions = JdbcIndexedSessionRepository.this.jdbcOperations.query` `sessions` was `null`. Weird right?

Maybe because this test is using mocked `JdbcOperations`? 

---

I think testing that `INSERT INTO [...]` is executed rather than testing `getAttribute` is making the test weaker. 
As a end-user I care that after a `setAttribute` a `getAttribute` returns the correct result. I don't care how this is done internally. For this reason I think it is cleaner to test the actual API surface instead of testing what the implementation of the API does internally.

eleftherias: Because the integration tests (`AbstractJdbcIndexedSessionRepositoryITests`) run on multiple implementations, 16 at the moment, we avoid adding tests for every scenario in this class.
Testing method like `setFlushMode`, `setSaveMode` etc go in `JdbcIndexedSessionRepositoryTests` class so that we can test different variations efficiently.
Additionally, the integration tests are intended to test the default `JdbcIndexedSessionRepository` that is autowired. By modifying this object we are no longer testing the default implementation. 
Also, calling `setSaveMode` at the end of the test makes the test suite fragile, since it would produce unexpected behavior if the default saveMode changed in the future.

Testing that the new attribute is inserted rather than updated is sufficient to catch this bug if we had a regression and it appeared again.
I would expect the test in `JdbcIndexedSessionRepositoryTests` to look something like this:
```
@Test
void saveWithSaveModeOnGetAttributeAndNewAttributeSetAndGet() {
	this.repository.setSaveMode(SaveMode.ON_GET_ATTRIBUTE);
	MapSession delegate = new MapSession();
	delegate.setAttribute("attribute1", (Supplier<String>) () -> "value1");
	JdbcSession session = this.repository.new JdbcSession(delegate, UUID.randomUUID().toString(), false);
	session.setAttribute("attribute2", "value2");
	session.getAttribute("attribute2");
	this.repository.save(session);
	ArgumentCaptor<BatchPreparedStatementSetter> captor = ArgumentCaptor
			.forClass(BatchPreparedStatementSetter.class);
	verify(this.jdbcOperations, times(1)).update(startsWith("INSERT INTO SPRING_SESSION_ATTRIBUTES ("),
			isA(PreparedStatementSetter.class));
	verifyNoMoreInteractions(this.jdbcOperations);
}
```

eleftherias:
Thanks for the PR @felixscheinost!
This is now merged into the `5.7.x` branch and the `main` branch.
I will also backport it to `5.6.x` and `5.5.x`.

felixscheinost:
Thanks!

PR #2055, opened 2022-04-18
Reuse StringRedisSerializer instance

quaff:


eleftherias:
Thanks for the suggestion @quaff.
Is the existing configuration causing any problems for you?

Since this is a sample configuration, I don't see the need for this change if it doesn't cause any problems.

vpavic (changes requested):
I'd merge this. There's no need to instantiate additional serializer instances when Spring Data Redis offers `RedisSerializer#string` exactly to avoid that.

Note @quaff that there are more occurrence of this in `RedisHttpSessionConfiguration` and `RedisWebSessionConfiguration`.

quaff:
> I'd merge this. There's no need to instantiate additional serializer instances when Spring Data Redis offers `RedisSerializer#string` exactly to avoid that.
> 
> Note @quaff that there are more occurrence of this in `RedisHttpSessionConfiguration` and `RedisWebSessionConfiguration`.

@vpavic Updated.

PR #2081, opened 2022-05-15
Polish related to testcontainers

eddumelendez:
* Use getHost instead of getContainerIpAddress
* Rely on OracleContainer and it's improvements which involve the log
wait strategy


marcusdacoregio:
Thanks @eddumelendez, this is now merged into `2.7.x` and `main`

Issue #1913, opened 2021-09-07
Consider moving away from Java serialization as default serialization mechanism


vpavic:
At present, all `SessionRepository` implementations use Java serialization as the default serialization mechanism.

While convenient as a default, Java serialization has several well known limitations and the ecosystem appears to be getting ready to move away from it - see [_Towards Better Serialization_ document](https://cr.openjdk.java.net/~briangoetz/amber/serialization.html) by Brian Goetz.

The next major release seems like a good opportunity to reconsider Spring Session's general approach to serialization.

PR #2099, opened 2022-05-23
Treat a SerializationException retrieving a session as though there w…

dbyron-sf:
…ere no session.

This makes zero-downtime / seamless upgrades possible in more cases.  Currently, upgrading spring-security across versions that change SpringSecurityCoreVersion.SERIAL_VERSION_ID requires manual deletion of cached sessions.

See also https://github.com/spring-projects/spring-security/issues/9204 and https://github.com/spring-projects/spring-security/issues/3737

### DIFF
--- a/spring-session-data-redis/src/main/java/org/springframework/session/data/redis/RedisIndexedSessionRepository.java
+++ b/spring-session-data-redis/src/main/java/org/springframework/session/data/redis/RedisIndexedSessionRepository.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2021 the original author or authors.
+ * Copyright 2014-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -37,6 +37,7 @@ import org.springframework.data.redis.core.BoundHashOperations;
 import org.springframework.data.redis.core.RedisOperations;
 import org.springframework.data.redis.serializer.JdkSerializationRedisSerializer;
 import org.springframework.data.redis.serializer.RedisSerializer;
+import org.springframework.data.redis.serializer.SerializationException;
 import org.springframework.data.redis.util.ByteUtils;
 import org.springframework.session.DelegatingIndexResolver;
 import org.springframework.session.FindByIndexNameSessionRepository;
@@ -454,7 +455,15 @@ public class RedisIndexedSessionRepository
 	 * @return the Redis session
 	 */
 	private RedisSession getSession(String id, boolean allowExpired) {
-		Map<Object, Object> entries = getSessionBoundHashOperations(id).entries();
+		Map<Object, Object> entries;
+		try {
+			entries = getSessionBoundHashOperations(id).entries();
+		}
+		catch (SerializationException ex) {
+			// An error deserializing is equivalent to not having any information at all.
+			logger.warn("exception getting session " + id, ex);
+			return null;
+		}
 		if (entries.isEmpty()) {
 			return null;
 		}
--- a/spring-session-data-redis/src/test/java/org/springframework/session/data/redis/RedisIndexedSessionRepositoryTests.java
+++ b/spring-session-data-redis/src/test/java/org/springframework/session/data/redis/RedisIndexedSessionRepositoryTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2019 the original author or authors.
+ * Copyright 2014-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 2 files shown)


quaff:
Workaround: customizing your own `CacheErrorHandler` instead of default `SimpleCacheErrorHandler`

```java
@EnableCaching
@Configuration(proxyBeanMethods = false)
public class CacheConfiguration extends CachingConfigurerSupport {

	@Override
	public CacheErrorHandler errorHandler() {
		return new DefaultCacheErrorHandler();
	}
}

@Slf4j
public class DefaultCacheErrorHandler implements CacheErrorHandler {

	@Override
	public void handleCacheGetError(RuntimeException exception, Cache cache, Object key) {
		if (exception instanceof SerializationFailedException
				|| exception.getCause() instanceof SerializationFailedException) {
			return;
		}
		if (exception.getCause() instanceof JsonParseException) {
			return;
		}
		if (exception instanceof QueryTimeoutException) {
			log.error(exception.getMessage());
			return;
		}
		throw exception;
	}

	@Override
	public void handleCachePutError(RuntimeException exception, Cache cache, Object key, @Nullable Object value) {
		if (exception instanceof QueryTimeoutException) {
			log.error(exception.getMessage());
			return;
		}
		throw exception;
	}

	@Override
	public void handleCacheEvictError(RuntimeException exception, Cache cache, Object key) {
		if (exception instanceof QueryTimeoutException) {
			log.error(exception.getMessage());
			return;
		}
		throw exception;
	}

	@Override
	public void handleCacheClearError(RuntimeException exception, Cache cache) {
		if (exception instanceof QueryTimeoutException) {
			log.error(exception.getMessage());
			return;
		}
		throw exception;
	}

}

```

quaff:
Customize bean `springSessionDefaultRedisSerializer` as workaround if you are using redis as session store
```java
@Configuration(proxyBeanMethods = false)
@EnableSpringHttpSession
public class SessionConfiguration {

	@Bean
	public RedisSerializer<?> springSessionDefaultRedisSerializer() {
		return new JdkSerializationRedisSerializer(getClass().getClassLoader()) {
			@Override
			public Object deserialize(byte[] bytes) throws SerializationException {
				try {
					return super.deserialize(bytes);
				}
				catch (SerializationException ex) {
					if (ex.getCause() instanceof SerializationFailedException) {
						return null; // ignore SerializationFailedException
					}
					throw ex;
				}
			}
		};
	}

}

```

dbyron-sf:
@marcusdacoregio Hope it's OK that I'm tagging you here as things have been quiet on this PR for awhile.  I see some activity elsewhere and would love a review.

Thanks much!

marcusdacoregio:
Hi @dbyron-sf, thanks for the ping.

I'll tag @rwinch so we can talk about this issue and get back to you. It may take a couple of weeks for us to take a look at this so I will ask you for a little bit more patience.

dbyron-sf:
Thanks so much @marcusdacoregio for taking a look!

dbyron-sf:
Hey @marcusdacoregio (or maybe @rwinch).  Sorry to nag, but it's been awhile and I'm still hoping to get some traction here.  Thanks for your help.

vpavic:
I don't think this should be addressed as proposed here. The PR targets only a single `SessionRepository` implementation (and the one that isn't a default anymore in 3.0) while in reality this is a concern that affects all session repositories that support Java based session serialization (and that's most if not all). Duplicating the `try-catch` block proposed here across different `SessionRepository` implementations isn't an attractive option either, for obvious reasons.

Note that this concern (safe session deserialization) in general is tracked under https://github.com/spring-projects/spring-session/issues/529.

My opinion is that this should be addressed using a centralized strategy that would deal with deserialization failures and would then be reused across all session repositories that support Java serialization.

However, taking a step back, my suggestion to anyone affected by this is to strongly consider opting into a different (non-default) session serialization mechanism. JSON serialization is the most attractive option there, and is already supported by all Redis based `SessionRepository` implementations. This way, you shouldn't be affected by Spring Security's changes of serial version id while also standing a much better chance of doing major upgrades without affecting persisted sessions.

Note that there's also https://github.com/spring-projects/spring-session/issues/1913 tracking the effort to move away from Java serialization.

dbyron-sf:
Thanks for the comments @vpavic.  What you're saying makes sense...and especially if it wouldn't be backported to some probably consider ancient/closed versions like 2.5/2.6/2.7 it won't help those upgrades.  And if this isn't the default in 3.0 it won't help that upgrade either.

Any chance you can give some pointers on how to select JSON serialization?

PS I'm following both those issues, but they've been quiet for quite some time.

vpavic:
There's a [sample application](https://github.com/spring-projects/spring-session/tree/2.7.0/spring-session-samples/spring-session-sample-boot-redis-json) within the Spring Session codebase that shows how to set up JSON based serialization. It also uses Spring Security's Jackson mixins.

dbyron-sf:
Just saw https://github.com/spring-projects/spring-session/issues/2124 fly by which looks related.

marcusdacoregio:
Based on the [comment](https://github.com/spring-projects/spring-session/pull/2099#issuecomment-1217566705) from @vpavic, I'll close this as duplicate of https://github.com/spring-projects/spring-session/issues/529

Issue #2108, opened 2022-07-08
Adapt DuplicateKeyException on insertSessionAttributes to new SQL Exception Translator


marcusdacoregio:
The commit https://github.com/spring-projects/spring-framework/commit/083113d8a4ce1220ee4ee7e21395830e34ea8450 changed the way that exceptions are translated from the database exception. In some of our tests, what before was a `DuplicateKeyException` now is a `DataIntegrityViolationException` and we ignore the `DataIntegrityViolationException`. See:

https://github.com/spring-projects/spring-session/blob/abce2eb555eec90f548160a81cbb38077a74b667/spring-session-jdbc/src/main/java/org/springframework/session/jdbc/JdbcIndexedSessionRepository.java#L506-L520

The code should be adapted to the new behavior. It is important to note that in case of a duplicate key we should still consider throwing an exception since it is a clue to users that they need to expose a `SessionRepositoryCustomizer` Bean (see #1213)

marcusdacoregio:
Closed via https://github.com/spring-projects/spring-session/pull/2113

PR #2113, opened 2022-07-18
Adapt tests that expected a DuplicateKeyException

marcusdacoregio:
Closes gh-2108

### DIFF
--- a/spring-session-jdbc/src/main/java/org/springframework/session/jdbc/JdbcIndexedSessionRepository.java
+++ b/spring-session-jdbc/src/main/java/org/springframework/session/jdbc/JdbcIndexedSessionRepository.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2020 the original author or authors.
+ * Copyright 2014-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


vpavic:
Hi @marcusdacoregio - I'll try to take a closer look at this over the next day or two. Hopefully that's OK, as I see no due date set on the [3.0.0-M3](https://github.com/spring-projects/spring-session/milestone/125) milestone.

eleftherias (approved):

vpavic (commented):
After taking a look at this, my opinion is that due to the JDBC `SessionRepository` having some quite specific error handling scenarios that have been established over the years (based on user feedback, mostly involving high concurrency use cases), there is a need for a more sophisticated SQL exception translation than the one that's present by default in Spring Framework starting with 6.0.

So basically, what I'm saying is that the `JdbcTemplate` that's configured in Spring Session's JDBC configuration infrastructure should set up an exception translator that's capable of telling `DuplicateKeyException` apart from `DataIntegrityViolationException`.

Note that not ensuring such behavior out of the box effectively means that `catch` blocks expecting `DuplicateKeyException` are dead code, unless users configure `JdbcIndexedSessionRepository` by themselves supplying `JdbcTemplate` with appropriate exception handling configured. 

Also, I don't think this warrants being labeled as breaking change (label `type: breaks-passivity`) since there's really nothing breaking in the Spring Session change.

marcusdacoregio:
Thank you @eleftherias and @vpavic.

I've decided to use the `SQLErrorCodeSQLExceptionTranslator` class by default in `JdbcTemplate`, this way we keep the behavior and in the future, we can add a way to customize the exception translator if there is a need.

PR #2117, opened 2022-08-11
Disable Hazelcast network join auto-detection in tests

vpavic:
At present, Hazelcast configurations used in tests disable multicast join but leave network join auto-detection enabled. This can cause issues with parallel test execution on machines that have bigger number of CPU cores/threads.

This commit updates Hazelcast configurations used in tests to disable network join auto-detection and thus ensure no network join method ends up being enabled.

---

Before making these changes, a number of Hazelcast tests regularly fail on my desktop (Ryzen 5 5600X with 6 cores/12 threads). 

Update:
I would hold off merging this for now because I plan to also try and see if other supported branches are affected by the same issue.


rwinch:
Thanks @vpavic Can you try rebasing to see if that fixes the build?

vpavic:
It did for me locally. Let me force push that to the remote.

vpavic:
The build is green on the CI as well now.

rwinch (approved):

rwinch:
Thanks this is now merged into 3.0.x

Issue #2122, opened 2022-08-16
Restructure Redis HttpSession configuration support


vpavic:
The purpose of this issue is to revisit the Redis `HttpSession` configuration arrangement put in place in https://github.com/spring-projects/spring-session/issues/1711. The `@EnableRedisHttpSession` attribute `enableIndexingAndEvents` introduced in that issue is a binary choice and as such is very fragile because it won't tolerate any additional `SessionRepository` implementations.

Goals:
- provide configuration support for the new default (`RedisSessionRepository`) while allowing to easily opt into the previous default (`RedisIndexedSessionRepository`)
- aim to provide extensible configuration support that will be friendly to additional `SessionRepository` implementations
- consider Spring Boot's auto-configuration support and ensure it's also able to opt into the previous defaults using configuration properties

Related issues:
- https://github.com/spring-projects/spring-boot/issues/30673

Issue #1010, opened 2018-03-02
Improve Redis configuration documentation


vpavic:
We should improve our reference documentation on Redis configuration capabilities to cover more advanced configuration scenarios (including `@SpringSessionRedisConnectionFactory`, `@SpringSessionRedisOperations` and other means of configuration customization).

PR #2123, opened 2022-08-16
Restructure Redis HttpSession configuration support

vpavic:
This PR reverts 8582b97 and restructures configuration support for Redis-backed `HttpSession` with aim to enable users to easily select the `SessionRepository` implementation they prefer to use.

This is achieved by introducing `@EnableRedisIndexedHttpSession` annotation that can be used to configure `RedisIndexedSessionRepository`, while the existing `@EnableRedisHttpSession` will going forward configure `RedisSessionRepository` as the `SessionRepository` implementation used by Spring Session.

Additionally, this also introduces `AbstractRedisHttpSessionConfiguration` as the base configuration class that manages common aspects of Redis-backed `HttpSession` support, which is then extended by more specific configuration classes that provide specific `SessionRepository` implementation.

Closes https://github.com/spring-projects/spring-session/issues/2122


rwinch (changes requested):
Thanks for the PR @vpavic! Overall it looks good. Can you please rebase and add documentation for the new annotations?

vpavic:
I've rebased the PR on top of the current `main`.

I considered the documentation updates while putting the PR together, however with all the changes that happened during the lifecycle of 2.x (that were unfortunately not documented in timely manner) and now these changes I feel that the docs are better addressed via https://github.com/spring-projects/spring-session/issues/1010 as a part of larger overhaul. That's one topic I believe we should discuss when we jump on the next sync.

In the meantime, I wouldn't hold this up, especially since https://github.com/spring-projects/spring-boot/pull/32205 is likely to get merged soon as well.

spring-session-data-redis/src/main/java/org/springframework/session/data/redis/RedisIndexedSessionRepository.java
```diff
@@ -288,7 +288,7 @@
 
 	private byte[] expiredKeyPrefixBytes;
 
-	private final RedisOperations<Object, Object> sessionRedisOperations;
+	private final RedisOperations<String, Object> sessionRedisOperations;
```
vpavic, line 291: Forgot to point out this after opening the PR - this is technically a minor breaking change but it was needed in order to align the base configuration support - `RedisSessionRepository` uses `RedisOperations<String, Object>` from the start, and that's more exact since our ids are strings anyway. However, I don't think it's a big deal considering we're also changing the session repository which `@EnableRedisHttpSession` provides.

rwinch: Thanks I noticed this. I think this is an acceptable change especially given this is a major release anyway.

Issue #2136, opened 2022-08-30
Simplify session cleanup related scheduled jobs


vpavic:
At present, `RedisIndexedHttpSessionConfiguration` and `JdbcHttpSessionConfiguration` rely on `@EnableScheduling` annotated inner configuration classes to set up their session cleanup related scheduled jobs.

This feels a bit invasive, as it silently opts in users into a general purpose task scheduling support provided by Spring Framework. In reality, those session cleanup jobs should really need a single thread that's dedicated to their execution and does not eat into the thread pool used by general purpose task scheduling (and vice-versa actually).

We should look into simplifying this setup, so that `SessionRepository` implementations that required session cleanup jobs manage their own scheduled executor service with minimal resource footprint.

PR #2163, opened 2022-09-20
Simplify expired session cleanup jobs

vpavic:
At present, `RedisIndexedHttpSessionConfiguration` and `JdbcHttpSessionConfiguration` include `@EnableScheduling` annotated inner configuration classes that configure expired session cleanup jobs. This approach silently opts in users into general purpose task scheduling support provided by Spring Framework, which isn't something a library should do. Ideally, session cleanup jobs should only require a single thread dedicated to their execution and also one that doesn't compete for resources with general purpose task scheduling.

This PR updates `RedisIndexedSessionRepository` and `JdbcIndexedSessionRepository` to have them manage their own `ThreadPoolTaskScheduler` for purposes of running expired session cleanup jobs.

Closes gh-2136

### DIFF
--- a/spring-session-data-redis/src/main/java/org/springframework/session/data/redis/RedisIndexedSessionRepository.java
+++ b/spring-session-data-redis/src/main/java/org/springframework/session/data/redis/RedisIndexedSessionRepository.java
@@ -27,6 +27,8 @@ import java.util.Set;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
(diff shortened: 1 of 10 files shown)


rwinch (changes requested):
I'd prefer that we don't rely on `@PostConstruct`/`@PostDestroy` for Beans as this means that the class must be instantiated by a container in order to be used. Instead, I'd consider using `InitializingBean` and `DisposableBean`

rwinch (approved):

rwinch:

PR #2167, opened 2022-09-23
Replace JSR-250 annotations with standard Spring lifecycle callbacks

vpavic:
This aligns the codebase with remarks from https://github.com/spring-projects/spring-session/pull/2163#pullrequestreview-1117585647.


rwinch (changes requested):
Is it possible to remove the jakarta.annotation-api dependency at this point?

vpavic:
> Is it possible to remove the jakarta.annotation-api dependency at this point?

Good catch - I forgot to remove it from core, and subsequently, from the dependency management altogether.

That's done now.

rwinch (approved):

PR #2169, opened 2022-09-26
Use java.time for temporal values in all session repositories

vpavic:


spring-session-data-mongodb/src/main/java/org/springframework/session/data/mongo/MongoSession.java
```diff
@@ -66,7 +67,7 @@ public class MongoSession implements Session {
 	private Map<String, Object> attrs = new HashMap<>();
 
 	public MongoSession() {
-		this(MongoIndexedSessionRepository.DEFAULT_INACTIVE_INTERVAL);
+		this(MapSession.DEFAULT_MAX_INACTIVE_INTERVAL_SECONDS);
 	}
 
 	public MongoSession(long maxInactiveIntervalInSeconds) {
```
rwinch, line 73: Should this use java.time as well? It seems like there are other locations where int or long are still being used as well. I know that the annotations need to use primitives, but I think the setters on the configuration should still be java.time and we should convert it.

spring-session-data-mongodb/src/main/java/org/springframework/session/data/mongo/MongoIndexedSessionRepository.java
```diff
@@ -46,15 +47,19 @@
...
+ * @author Vedran Pavic
  * @since 2.2.0
  */
 public class MongoIndexedSessionRepository
 		implements FindByIndexNameSessionRepository<MongoSession>, ApplicationEventPublisherAware, InitializingBean {
 
 	/**
 	 * The default time period in seconds in which a session will expire.
+	 * @deprecated since 3.0.0 in favor of
+	 * {@link MapSession#DEFAULT_MAX_INACTIVE_INTERVAL_SECONDS}
```
rwinch, line 59: Why does MongoIndexedSessonRepository depend on MapSession?

spring-session-data-mongodb/src/main/java/org/springframework/session/data/mongo/MongoIndexedSessionRepository.java
```diff
@@ -178,8 +181,16 @@ private void publishEvent(ApplicationEvent event) {
...
-	public void setMaxInactiveIntervalInSeconds(final Integer maxInactiveIntervalInSeconds) {
-		this.maxInactiveIntervalInSeconds = maxInactiveIntervalInSeconds;
+	/**
+	 * Set the maximum inactive interval in seconds between requests before newly created
+	 * sessions will be invalidated. A negative time indicates that the session will never
+	 * time out. The default is 30 minutes.
+	 * @param defaultMaxInactiveInterval the default maxInactiveInterval
+	 */
+	public void setDefaultMaxInactiveInterval(Duration defaultMaxInactiveInterval) {
+		org.springframework.util.Assert.notNull(defaultMaxInactiveInterval,
```
rwinch, line 191: I see that Mongo's Assert.requireNonNull is already used which is why this must be fully qualified. This usage is inconsistent, but it brings the question of why Spring's Assert.notNull isn't used in the other case. Perhaps we should log a ticket to align this?

rwinch (changes requested):

spring-session-data-mongodb/src/main/java/org/springframework/session/data/mongo/MongoIndexedSessionRepository.java
vpavic: Actually, MongoDB implementation is in many ways an outlier in our codebase. All other session repositories keep their `org.springframework.session.Session` implementations as an internal implementation detail. Also, all those implementations delegate to `MapSession` rather then re-implementing all the logic.

spring-session-data-mongodb/src/main/java/org/springframework/session/data/mongo/MongoIndexedSessionRepository.java
vpavic: Along the lines of https://github.com/spring-projects/spring-session/pull/2169#discussion_r980348567, IMO we need to do a more general alignment of MongoDB session repository implementations. I'll open an issue to track that.

PR #2139, opened 2022-09-16
Update Gradle Enterprise plugin to 3.11.1

jprinet:

PR #2192, opened 2022-10-18
Upgrade Gradle to 7.5.1

vpavic:
Note that I've also added a commit that introduces `.gitattributes` in order for git to respect line endings in `gradlew.bat`.


vpavic:
While merging this, we should also consider #2139.

PR #2197, opened 2022-10-24
Update reference site link

candrews:
https://docs.spring.io/spring-session/docs/current/reference/html5/ goes to a 404 page

Use https://docs.spring.io/spring-session/reference/ instead which is the current documentation site.


rwinch:

Issue #2195, opened 2022-10-19
Move Spring Session Bom to spring-projects/spring-session


rwinch:
* [ ] Move the Source code to this repository
* [ ] Update spring.io/projects/ to have a single project page

PR #2199, opened 2022-10-25
Add Spring Session BOM module

vpavic:
With Spring Session Data Geode module being removed from the BOM, all of Spring Session's modules are now managed by this repository.

This means that the BOM itself can be moved to this repository, in order to simplify the overall project maintenance.

See #2195


spring-session-docs/modules/ROOT/nav.adoc
```diff
@@ -15,7 +15,6 @@
 *** xref:guides/boot-webflux-custom-cookie.adoc[Custom Cookie]
 ** Java Configuration
 ** XML Configuration
-* xref:modules.adoc[Modules]
```
vpavic, line 18: The documentation section describing modules arrangement was introduced in 2.0 when project was split into separate repos and the BOM was originally introduced. With everything now being in the same repo, I thought this section could safely be removed.

spring-session-docs/modules/ROOT/pages/whats-new.adoc
```diff
@@ -1,4 +1 @@
 = What's New
-
-Check also the Spring Session BOM https://github.com/spring-projects/spring-session-bom/wiki#release-notes[release notes]
```
vpavic, line 3: 3.0 needs _What's New_ updated, but in the meantime we should probably remove this reference to external BOM repo.

rwinch (approved):

PR #2251, opened 2023-02-23
Provide Session Id Generation Strategy

quaff:
Fix GH-11

### DIFF
--- a/spring-session-core/src/main/java/org/springframework/session/MapSession.java
+++ b/spring-session-core/src/main/java/org/springframework/session/MapSession.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2022 the original author or authors.
+ * Copyright 2014-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -21,9 +21,10 @@ import java.time.Duration;
 import java.time.Instant;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.Map;
+import java.util.ServiceLoader;
 import java.util.Set;
-import java.util.UUID;
 
 /**
  * <p>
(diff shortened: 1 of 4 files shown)


spring-session-core/spring-session-core.gradle
```diff
@@ -4,6 +4,7 @@ description = "Spring Session"
 
 dependencies {
 	api "org.springframework:spring-jcl"
+	implementation "org.springframework:spring-core"
```
vpavic, line 7: Historically, Spring Session has avoided the required dependency on `spring-core` so that the Servlet support can be used even in applications that are not Spring-based. See [`spring-session-sample-misc-hazelcast`](https://github.com/spring-projects/spring-session/tree/main/spring-session-samples/spring-session-sample-misc-hazelcast) as an example of such arrangement.

quaff:
@vpavic Switched from `SpringFactoriesLoader` to `java.util.ServiceLoader`, I don't know why CI failed, but all tests passed locally.

spring-session-core/src/main/java/org/springframework/session/MapSession.java
```diff
@@ -229,7 +231,14 @@ public int hashCode() {
 	}
 
 	private static String generateId() {
-		return UUID.randomUUID().toString();
+		return sessionIdGenerator.generateId();
+	}
+
+	private static final SessionIdGenerator sessionIdGenerator;
```
trydofor, line 237: `static` is not good idea

quaff: Then what's your suggestion?

trydofor: At present, this PR can solve the problem easily. 
I think `static` is structural problem (not this PR), I prefer PR #1547, using spring DI is better than SPI.

quaff: That huge changeset is risky, this one is simple and reliable.

marcusdacoregio:
Closing in favor of https://github.com/spring-projects/spring-session/pull/2286

Issue #2219, opened 2022-12-07
Compile sources with '-parameters' to retain method parameter names


quaff:
Warning show up when upgrade to spring boot 3.0
```
o.s.c.LocalVariableTableParameterNameDiscoverer WARN Using deprecated '-debug' fallback for parameter name resolution. Compile the affected code with '-parameters' instead or avoid its introspection: org.springframework.session.config.annotation.web.http.SpringHttpSessionConfiguration
```

quaff:
Superseded  by https://github.com/spring-projects/spring-session/pull/2252

marcusdacoregio:
Hi @quaff,

I am not seeing this warning, can you help me reproduce it?

In addition to that, if you are using IntelliJ IDEA, can you check under Settings > Build, Execution, Deployment > Compiler > Java Compiler, if you are not using `-debug` as compilation options?

quaff:
> Hi @quaff,
> 
> I am not seeing this warning, can you help me reproduce it?
> 
> In addition to that, if you are using IntelliJ IDEA, can you check under Settings > Build, Execution, Deployment > Compiler > Java Compiler, if you are not using `-debug` as compilation options?

It's recommended to use `-parameters` for reflection, my own code always complied with parameters, the warning reported by spring is come from 3rd party jars such as spring-session, it should be fixed at their side.

quaff:
I didn't notice this warning recently, It probably fixed by spring-framework.

PR #2252, opened 2023-02-24
Compile sources with '-parameters' to retain method parameter names

quaff:
Fix GH-2219


marcusdacoregio:
Closing since it seems to be fixed by the framework.

https://github.com/spring-projects/spring-session/issues/2219#issuecomment-1524723459

Issue #2256, opened 2023-03-08
Introduce Utility Method to Replace Default Table Name


marcusdacoregio:
Related to https://github.com/spring-projects/spring-session/issues/2225

PR #2257, opened 2023-03-08
Introduce Utility Method to Replace Default Table Name

marcusdacoregio:
Closes gh-2256

### DIFF
--- a/spring-session-jdbc/src/main/java/org/springframework/session/jdbc/util/JdbcSchemaUtils.java
+++ b/spring-session-jdbc/src/main/java/org/springframework/session/jdbc/util/JdbcSchemaUtils.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2014-2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


marcusdacoregio:
Hey @philwebb, it would be great to get your feedback on this to see how it fits what Spring Boot expects.

spring-session-jdbc/src/main/java/org/springframework/session/jdbc/util/JdbcSchemaUtils.java
```diff
@@ -0,0 +1,49 @@
...
+
+	/**
+	 * Loads the content of the provided schema resource and replaces the
+	 * {@link JdbcIndexedSessionRepository#DEFAULT_TABLE_NAME} by the provided table name.
+	 * @param schemaResource the schema resource
+	 * @param tableName the table name to replace
+	 * @return the schema resource with the table name replaced
+	 */
+	public static Resource replaceDefaultTableName(Resource schemaResource, String tableName) throws IOException {
+		String schemaScript = new String(schemaResource.getInputStream().readAllBytes());
```
philwebb, line 44: We should probably do `try (InputStream inputStream = schemaResource.getInputStream()) {...}` to ensure the stream is closed.

spring-session-jdbc/src/main/java/org/springframework/session/jdbc/util/JdbcSchemaUtils.java
```diff
@@ -0,0 +1,49 @@
...
+		String newSchema = schemaScript.replace(JdbcIndexedSessionRepository.DEFAULT_TABLE_NAME, tableName);
+		return new ByteArrayResource(newSchema.getBytes());
```
philwebb, line 46: We might want to read and write with `StandardCharsets.UTF_8`.

philwebb:
@marcusdacoregio It looks good to me, we should be able to make use of it without issue. I've added a couple of minor review comments.

PR #1910, opened 2021-08-31
Make SessionRepositoryRequestWrapper::commitSession idempotent

quaff:
Before this commit, commitSession could be called twice, which cause unnecessary access to sessionRepository
1. closing response output stream by other filter
```java
	SessionRepositoryFilter$SessionRepositoryRequestWrapper.commitSession() line: 217
	SessionRepositoryFilter$SessionRepositoryResponseWrapper.onResponseCommitted() line: 180
	SessionRepositoryFilter$SessionRepositoryResponseWrapper(OnCommittedResponseWrapper).doOnResponseCommitted() line: 227
	OnCommittedResponseWrapper$SaveContextServletOutputStream.close() line: 505
	LoggingBodyHttpServletResponse$ResponseServletOutputStream.close() line: 145 // other HttpServletResponse wrapper
	RestFilter.doFilter(ServletRequest, ServletResponse, FilterChain) line: 87 // other filter
	ApplicationFilterChain.internalDoFilter(ServletRequest, ServletResponse) line: 189
	ApplicationFilterChain.doFilter(ServletRequest, ServletResponse) line: 162
```
2. finally statement in SessionRepositoryFilter::doFilterInternal
```java
	SessionRepositoryFilter$SessionRepositoryRequestWrapper.commitSession() line: 217
	SessionRepositoryFilter<S>.doFilterInternal(HttpServletRequest, HttpServletResponse, FilterChain) line: 145
	SessionRepositoryFilter<S>(OncePerRequestFilter).doFilter(ServletRequest, ServletResponse, FilterChain) line: 82
	ApplicationFilterChain.internalDoFilter(ServletRequest, ServletResponse) line: 189
	ApplicationFilterChain.doFilter(ServletRequest, ServletResponse) line: 162
```

PR #2394, opened 2023-08-10
Introduce SessionRepositoryFilter::setCommitSessionOncePerRequest

quaff:
Fix https://github.com/spring-projects/spring-session/issues/1424
See https://github.com/spring-projects/spring-session/pull/1910


quaff:
Please review @rwinch @vpavic

marcusdacoregio:
I'm closing this since it has been discussed in the past in https://github.com/spring-projects/spring-session/pull/1910

quaff:
> I'm closing this since it has been discussed in the past in #1910

There is no discussion on this new option, previous discussion point out the change is not safe for some use case, this PR introduce an option default to `false` which will break nothing, allow user enable it if them want, I think this PR should be reopen, WDYT? @vpavic

PR #2395, opened 2023-08-10
Use spring-boot's @ServiceConnection

quaff:


marcusdacoregio:
Thanks, @quaff. This is now merged into `main`

Issue #2393, opened 2023-08-08
ReactiveRedisSessionRepository Block when creating new session


realzero0:
`java.util.UUID.randomUUID()` is blocking

it might have to be changed to non-blocking way..

https://github.com/reactor/BlockHound/issues/157

https://github.com/spring-projects/spring-session/blob/2d4233fcfdc462187966bb45b0b9cc2bb99cc9ff/spring-session-data-redis/src/main/java/org/springframework/session/data/redis/ReactiveRedisSessionRepository.java#L127

quaff:
You can implement your own [SessionIdGenerator](https://github.com/spring-projects/spring-session/blob/main/spring-session-core/src/main/java/org/springframework/session/SessionIdGenerator.java) after v3.2.0 released.

quaff:
@marcusdacoregio You missed https://github.com/spring-projects/spring-session/pull/2396

PR #2396, opened 2023-08-11
Avoid blocking on session id generation

quaff:
see https://github.com/spring-projects/spring-session/issues/2393


marcusdacoregio:
Thanks, @quaff, this is now merged into `main`.

PR #2399, opened 2023-08-18
Introduce SessionIdGenerator.DEFAULT

quaff:
organize code via centralizing the default SessionIdGenerator instance


quaff:
It's safe to merge this before merging https://github.com/spring-projects/spring-session/pull/2387. @marcusdacoregio

marcusdacoregio:
Hi, @quaff. We do not want to merge this now because there are no clear pros to do it.

Also, adding the constant to the interface is a pattern that we do not follow and it is a bit weird that the interface has a direct reference to one of its implementations.

I will close this for now but we can always revisit when we get more feedback. Thank you!

quaff:
> Hi, @quaff. We do not want to merge this now because there are no clear pros to do it.

The pro is make code more organized, changes will be little if we switch to another default implementation.

> Also, adding the constant to the interface is a pattern that we do not follow and it is a bit weird that the interface has a direct reference to one of its implementations.

Adding the constant to the interface is common pattern, for example:

https://github.com/spring-projects/spring-framework/blob/03650e7d58ebbd68e8cff22db379ec974b289423/spring-aop/src/main/java/org/springframework/aop/Pointcut.java#L51
https://github.com/spring-projects/spring-framework/blob/03650e7d58ebbd68e8cff22db379ec974b289423/spring-expression/src/main/java/org/springframework/expression/ParserContext.java#L62

or we could use static method instead of constant to take advantage of new feature of java language.

PR #2450, opened 2023-08-23
Polishing

quaff:
1. fix deprecations
2. fix generics
3. use lambda instead of anonymous class
4. mark some fields as final

### DIFF
--- a/spring-session-data-mongodb/src/integration-test/java/org/springframework/session/data/mongo/MongoDbDeleteJacksonSessionVerificationTest.java
+++ b/spring-session-data-mongodb/src/integration-test/java/org/springframework/session/data/mongo/MongoDbDeleteJacksonSessionVerificationTest.java
@@ -36,6 +36,7 @@ import org.springframework.data.mongodb.core.ReactiveMongoTemplate;
 import org.springframework.http.HttpHeaders;
 import org.springframework.http.MediaType;
 import org.springframework.http.ResponseEntity;
+import org.springframework.security.config.Customizer;
 import org.springframework.security.config.annotation.web.reactive.EnableWebFluxSecurity;
 import org.springframework.security.config.web.server.ServerHttpSecurity;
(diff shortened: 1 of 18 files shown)


spring-session-data-mongodb/src/integration-test/java/org/springframework/session/data/mongo/MongoDbDeleteJacksonSessionVerificationTest.java
```diff
@@ -142,14 +143,10 @@ static class SecurityConfig {
...
-					.formLogin() //
-					/**/.and() //
-					.csrf().disable() //
-					.authorizeExchange() //
-					.anyExchange().authenticated() //
-					/**/.and() //
+					.logout(Customizer.withDefaults())//
+					.formLogin(Customizer.withDefaults()) //
+					.csrf((csrf) -> csrf.disable()) //
+					.authorizeExchange((ae) -> ae.anyExchange().authenticated()) //
```
marcusdacoregio, line 149: We can probably remove those `//` too and wrap the block of code with `// @formatter:off` and `// @formatter:on`

```java
// @formatter:off
return http 
					.logout(Customizer.withDefaults())
					.formLogin(Customizer.withDefaults()) 
					.csrf((csrf) -> csrf.disable()) 
					.authorizeExchange((ae) -> ae.anyExchange().authenticated())
                     .build();
// @formatter:on
```

marcusdacoregio (approved):
Thank you so much for the contribution @quaff. I just left a minor feedback inline.
Feel free to apply it or not.

spring-session-data-mongodb/src/integration-test/java/org/springframework/session/data/mongo/MongoDbDeleteJacksonSessionVerificationTest.java
quaff: done.

PR #2711, opened 2024-01-03
Add Support for Authenticated Encrypted Sessions

marcusdacoregio:

### DIFF
--- a/spring-session-core/src/main/java/org/springframework/session/config/annotation/web/http/PersistentSessionConfiguration.java
+++ b/spring-session-core/src/main/java/org/springframework/session/config/annotation/web/http/PersistentSessionConfiguration.java
@@ -1,0 +1,104 @@
+/*
+ * Copyright 2014-2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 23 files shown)


spring-session-core/src/main/java/org/springframework/session/config/annotation/web/http/PersistentSessionConfiguration.java
```diff
@@ -0,0 +1,104 @@
...
+		PersistentSessionRepository repository = new PersistentSessionRepository(getSecretKey(), principalResolver,
+				principalRestorer);
+		sessionRepositoryCustomizers.orderedStream().forEach((customizer) -> customizer.customize(repository));
+		return repository;
+	}
+
+	@Autowired(required = false)
+	@Qualifier("springSessionPersistentSessionSecretKey")
+	public void setSecretKey(SecretKey secretKey) {
+		this.secretKey = secretKey;
```
rwinch, line 69: I think that using a qualifier is ok. However, you might consider providing a meta annotation so that users do not need to remember the qualifier name.

 I'm not sure but I think Qualifier only is used if there are multiple beans of the same type (you might double check that). If that is true, you might create a configuration object that could be provided that contains the secret key. That would also ensure that users don't accidentally using another SecretKey (which is a common type) with Spring Session.

rwinch (changes requested):
@marcusdacoregio I've provided feedback inline

PR #2828, opened 2024-03-01
Add H2 specific SessionRepositoryCustomizer

repolevedavaj:
This PR introduces a `SessionRepositoryCustomizer` for H2. The implementation is quite similar to the one for Oracle. The only difference is the needed datatype casts for the parameters.

### DIFF
--- a/spring-session-jdbc/src/integration-test/java/org/springframework/session/jdbc/H2JdbcIndexedSessionRepositoryCustomizerITests.java
+++ b/spring-session-jdbc/src/integration-test/java/org/springframework/session/jdbc/H2JdbcIndexedSessionRepositoryCustomizerITests.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2014-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


marcusdacoregio:
Hi, @repolevedavaj. Thanks for the PR.

Can you elaborate more why such customizer is needed? Does it also work if [H2 compability mode](https://www.h2database.com/html/features.html#compatibility) is set to Postgres, for example?

repolevedavaj:
> Hi, @repolevedavaj. Thanks for the PR.
> 
> Can you elaborate more why such customizer is needed? Does it also work if [H2 compability mode](https://www.h2database.com/html/features.html#compatibility) is set to Postgres, for example?

Hi @marcusdacoregio, I created the customizer as the existing ones were not working with a H2 database, even in compatibility mode. I did not test it specifically if it works in Postgres compatibility mode.

marcusdacoregio:
I think that there is uncertainty whether this customizer will work fine with any combination of H2's configurations and there is no easy way for us to test it, therefore I think it would be better to not add it now. If we receive more reports asking for an H2 customizer, we can combine the use-cases and provide one. Does that sound ok for you?

repolevedavaj:
Hi @marcusdacoregio, I see your point. According to H2, the used syntax is not depending on a specific mode, but I don't know if it behaves differently with in specific modes. For me its fine to close the PR as suggested.

Issue #2230, opened 2023-01-27
Clustered Redis server memory leak because FindByIndexNameSessionRepository.PRINCIPAL_NAME_INDEX_NAME key is not removed after session expired


sjzcao:
**Describe the bug**
spring session is configured using redis cluster servers and redis server memory usage keep increase because "*FindByIndexNameSessionRepository.PRINCIPAL_NAME_INDEX_NAME*" keys are not deleted after sessions are expired. 

**To Reproduce**
1/ prepare a local redis cluster. In my demo, there are 6 nodes. 3 master nodes and 3 slave nodes. (masters: localhost:7000/7001/7002, slaves: localhost:7003/7004/7005) 
2/ a sample spring security project (refer sample part) and config springSessionRedisConnectionFactory by adding all 6 nodes (or one node, does not matter)
3/ start the spring security project and login page http://127.0.0.1:8080
4/ open redis-cli and monitor key

**Expected behavior**
After all sessions expired, the related index keys "*FindByIndexNameSessionRepository.PRINCIPAL_NAME_INDEX_NAME*" should be deleted to release redis server memory.

**Analysis**
The root cause of the issue is seems that spring session only add del and expired event listener to one of the redis server, if the "*sessions:expires*" key is not created in the server to which the listener listen, spring session will miss the event and cause the index key (spring:session:index:org.springframework.session.FindByIndexNameSessionRepository.PRINCIPAL_NAME_INDEX_NAME:user) not removed.

Here is an example, in my demo environment port 7000,7001 and 7002 are master nodes and spring session add event listener to 7001:
127.0.0.1:7001> pubsub channels
1) "__keyevent@0__:expired"
2) "__keyevent@0__:del"

after log in my demo page, the expires key added to 7002:
127.0.0.1:7002> keys *
1) "spring:session:index:org.springframework.session.FindByIndexNameSessionRepository.PRINCIPAL_NAME_INDEX_NAME:user"
2) **"spring:session:sessions:expires:1e9938ed-d74c-4143-bb87-66d4e3cab304"**
3) "spring:session:sessions:20a65ccc-ce7c-408e-aa4b-a91e2f69cb6b"

When "spring:session:sessions:expires:1e9938ed-d74c-4143-bb87-66d4e3cab304" expired, spring session missed the event because the listener is only listening 7001.

**Sample**
https://github.com/sjzcao/spring-session-memory-issue.git

Issue #2614, opened 2023-11-07
PRINCIPAL_NAME_INDEX_NAME some times cant clean up


wash1983:
like this if not exists session, remove index when get it. is this ok?
```java
        @Override
	public Map<String, RedisSession> findByIndexNameAndIndexValue(String indexName, String indexValue) {
		if (!PRINCIPAL_NAME_INDEX_NAME.equals(indexName)) {
			return Collections.emptyMap();
		}
		String principalKey = getPrincipalKey(indexValue);
		Set<Object> sessionIds = this.sessionRedisOperations.boundSetOps(principalKey).members();
		Map<String, RedisSession> sessions = new HashMap<>(sessionIds.size());
		for (Object id : sessionIds) {
			RedisSession session = findById((String) id);
			if (session != null) {
				sessions.put(session.getId(), session);
			}else{
                             this.sessionRedisOperations.boundSetOps(principalKey).remove(session.getId());
                       }
		}
		return sessions;
	}
```

marcusdacoregio:
Hi, @wash1983. I did not understand what you are asking here, could you please re-elaborate the question?

wash1983:
In the context of spring session date redis, there is a problem where the sessions are not being cleared when the application server crashes, leading to an accumulation of the index which causes delays when using sessionManagement to restrict the number of user login sessions. My question is, is it possible to add validation in the`org.springframework.session.data.redis.RedisIndexedSessionRepository.findByIndexNameAndIndexValue(String, String)` method so that we can clean up the sessionId in PRINCIPAL_NAME_INDEX_NAME for the expired sessions? This way, even if the PRINCIPAL_NAME_INDEX_NAME is not cleaned up immediately when a session expires, it can still be cleaned up when accessed.
add code    }else{
                             this.sessionRedisOperations.boundSetOps(principalKey).remove(session.getId());
                       }

marcusdacoregio:
I don't believe that we should apply that workaround because the problem relies on the mechanisms used to store the expirations and clean them. I'm working on a new Redis indexed implementation for WebFlux and there is a [new mechanism that seems promising](https://github.com/spring-projects/spring-session/pull/2681/files#diff-635e667aa995606f42bf6d94ffde20107a2682894685777afafb21456a63f98c). We can then extend the current implementation to support specifying a different expiration policy.

marcusdacoregio:
[...]

PR #2869, opened 2024-03-19
Remove expired session from principal index

kinsersh:
Despite RedisIndexedSessionRepository's efforts to automatically remove expired sessions from its principal index, it is possible for expired sessions to remain in that index due to misses from at least these causes:
- application failure due to no event redelivery (see https://docs.spring.io/spring-data/redis/reference/redis/redis-repositories/expirations.html)
  > Redis Pub/Sub messages are not persistent. If a key expires while the application is down, the expiry event is not processed, which may lead to secondary indexes containing references to the expired object.
- application only gets events from one redis node in a cluster (see https://github.com/spring-projects/spring-data-redis/issues/1111)

The findByIndexNameAndIndexValue method in RedisIndexedSessionRepository reads every session in the principal index including the expired ones, only returning the live ones. As expired sessions stack up, it slows down the findByIndexNameAndIndexValue method. Even though RedisIndexedSessionRepository encounters expired sessions, it doesn't take the opportunity to delete them from the principal index. The effect is that expired sessions are read over and over, never getting cleaned up. This was observed negatively affecting Spring Authorization Server's token endpoint due to its reliance on findByIndexNameAndIndexValue via SessionRegistry's getAllSessions method (see https://github.com/spring-projects/spring-authorization-server/blob/29a8321ab2d65f2f479311c28dd72b3809353722/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeAuthenticationProvider.java#L323).

 This commit improves RedisIndexedSessionRepository to delete the expired session from the principal index when findByIndexNameAndIndexValue sees that a session is expired. This mitigates the impact of expired sessions in that they will only be read once by findByIndexNameAndIndexValue before being removed out of the index.

 Mitigates #2230.

### DIFF
--- a/spring-session-data-redis/src/main/java/org/springframework/session/data/redis/RedisIndexedSessionRepository.java
+++ b/spring-session-data-redis/src/main/java/org/springframework/session/data/redis/RedisIndexedSessionRepository.java
@@ -518,13 +518,17 @@ public class RedisIndexedSessionRepository
 			long prevMin = roundDownMinute(System.currentTimeMillis());
 			String expirationsPrefix = this.namespace + "expirations:";
 
-			List<String> expirationKeys = this.sessionRedisOperations.keys(expirationsPrefix + "*").stream()
+			List<String> expirationKeys = this.sessionRedisOperations.keys(expirationsPrefix + "*")
+				.stream()
 				.filter(expirationKey -> Long.parseLong(expirationKey.substring(expirationsPrefix.length())) < prevMin)
 				.toList();
 
 			orphanedSessionIds.stream()
-				.filter(orphanedSessionId -> expirationKeys.stream().noneMatch(expirationKey -> this.sessionRedisOperations.boundSetOps(expirationKey).isMember(orphanedSessionId)))
-				.forEach(orphanedSessionId -> this.sessionRedisOperations.boundSetOps(principalKey).remove(orphanedSessionId));
+				.filter(orphanedSessionId -> expirationKeys.stream()
+					.noneMatch(expirationKey -> this.sessionRedisOperations.boundSetOps(expirationKey)
+						.isMember(orphanedSessionId)))
+				.forEach(orphanedSessionId -> this.sessionRedisOperations.boundSetOps(principalKey)
+					.remove(orphanedSessionId));
 		}
 
 		return sessions;
--- a/spring-session-data-redis/src/test/java/org/springframework/session/data/redis/RedisIndexedSessionRepositoryTests.java
+++ b/spring-session-data-redis/src/test/java/org/springframework/session/data/redis/RedisIndexedSessionRepositoryTests.java
@@ -424,7 +424,8 @@ class RedisIndexedSessionRepositoryTests {
 				RedisSessionMapper.MAX_INACTIVE_INTERVAL_KEY, 1, RedisSessionMapper.LAST_ACCESSED_TIME_KEY,
 				Instant.now().minus(5, ChronoUnit.MINUTES).toEpochMilli());
 		given(this.boundHashOperations.entries()).willReturn(map);
-		given(this.redisOperations.keys("spring:session:expirations:*")).willReturn(Set.of("spring:session:expirations:1616594540000"));
+		given(this.redisOperations.keys("spring:session:expirations:*"))
+			.willReturn(Set.of("spring:session:expirations:1616594540000"));
 
 		this.redisRepository.findByIndexNameAndIndexValue(FindByIndexNameSessionRepository.PRINCIPAL_NAME_INDEX_NAME,
 				"principal");


marcusdacoregio:
Hi, @kinsersh. Thanks for the PR.

Have you already read [this issue](https://github.com/spring-projects/spring-session/issues/2614) and checked if the workaround satisfies your need for now?

kinsersh:
> Hi, @kinsersh. Thanks for the PR.
> 
> Have you already read [this issue](https://github.com/spring-projects/spring-session/issues/2614) and checked if the workaround satisfies your need for now?

I wasn’t aware of that issue. Funny that @wash1983 suggested the same workaround.

If the session still exists in redis, that workaround you note would allow `RedisIndexedSessionRepository`’s cleanup to do its job.

My remaining question is what to do if the key isn’t found in any `spring:session:expirations:*` entries? If not found, I assume that means redis likely already eliminated the `spring:session:sessions:<key>` and `spring:session:sessions:expires:<key>` entries, so there’d be no way to trigger the delete event on the expires key. Would you consider deleting the key out of the principal index where this scan comes up dry? The desire is to mitigate the redis memory leak and the slowdown to every call to `findByIndexNameAndIndexValue` due to it calling redis with HGETALL for stale sessions that aren’t cleaned out if the app missed the expired event.

marcusdacoregio:
Yes, I believe that the session can be deleted from the principal index when that happens.

For the reactive indexed implementation of Spring Session Redis, we adopted a different strategy to track session expirations, see the details [here](https://docs.spring.io/spring-session/reference/3.3/configuration/reactive-redis-indexed.html#how-spring-session-cleans-up-expired-sessions). Do you think that approach is better than the current one in the servlet implementation?

kinsersh:
> Yes, I believe that the session can be deleted from the principal index when that happens.
> 
> For the reactive indexed implementation of Spring Session Redis, we adopted a different strategy to track session expirations, see the details [here](https://docs.spring.io/spring-session/reference/3.3/configuration/reactive-redis-indexed.html#how-spring-session-cleans-up-expired-sessions). Do you think that approach is better than the current one in the servlet implementation?

Yes, I see a benefit with the new approach. If the app is unavailable for a time, instead of skipping some `spring:session:expirations:<expiration_minute>` entries, the reactive implementation will go through the oldest entries in `spring:session:sessions:expirations` the next time cleanup runs. I think the redis memory leak from orphaned `spring:session:expirations:<expiration_minute>` entries is resolved.

Redis memory leaks from `spring:session:sessions:index:PRINCIPAL_NAME_INDEX_NAME:<principal name>` and `spring:session:sessions:<session id>:idx` are possible if the app isn’t running between when redis attempts to deliver the expired event for `spring:session:sessions:expires:<session id>` and redis removes the session data at `spring:session:sessions:<session id>`.

As for handling the leak affecting `spring:session:sessions:index:PRINCIPAL_NAME_INDEX_NAME:<principal name>`, we can still clean up orphaned entries when they're encountered (caused by app downtime). One weakness of this approach is that the principal indexes that aren't visited aren't ever cleaned up.

If Spring Session supported a configurable max session expiration with a default of say 24 hours, we'd be able to set an upper EXPIRES value on `spring:session:sessions:<session id>:idx`. Similarly, an upper EXPIRES could be set on `spring:session:sessions:index:PRINCIPAL_NAME_INDEX_NAME:<principal name>` and re-extend with any new session that's created for a particular principal ensuring redis would eventually clean it up. This feature would both help with this Redis memory leak and better support [OWASP's recommendation on absolute session timeouts](https://cheatsheetseries.owasp.org/cheatsheets/Session_Management_Cheat_Sheet.html#absolute-timeout).

wash1983:
The exact cause has not been identified, but it has indeed been observed in production that the PRINCIPAL_NAME_INDEX_NAME contains a vast number of entries, reaching tens of thousands or even more, which results in significantly degraded performance for org.springframework.session.security.SpringSessionBackedSessionRegistry<? extends Session> when limiting the number of user sessions.

Currently, the temporary solution provided in the pull request (PR) is being employed to address the issue.

kinsersh:
@marcusdacoregio, I added https://github.com/spring-projects/spring-session/pull/2869/commits/833e1b67180be5dc8cfd962712d9d1bffd21b9d5 to this PR, implementing the what was discussed at https://github.com/spring-projects/spring-session/pull/2869#issuecomment-2016136675.

Penghui-liu:
> @marcusdacoregio，我将[833e1b6](https://github.com/spring-projects/spring-session/commit/833e1b67180be5dc8cfd962712d9d1bffd21b9d5)[添加到此 PR，查看#2869 (评论)](https://github.com/spring-projects/spring-session/pull/2869#issuecomment-2016136675)中讨论的内容。

I found a new problem. If session concurrency control is enabled, when a new user accesses, the old user's sessionAttr:org.springframework.session.security.SpringSessionBackedSessionInformation.EXPIRED is set to true, but the expiration time does not change, so it will not be deleted from the main index.
So in your changes, you also need to add verification of sessionAttr:org.springframework.session.security.SpringSessionBackedSessionInformation.EXPIRE

Issue #3097, opened 2024-07-15
Support persistent cookies for storing session ID


theigl:
**Expected Behavior**

We would like to store the session ID in a cookie with an expiration date that matches the session timeout. The cookie expiration date should be updated every time the user accesses the session.

**Current Behavior**

The session ID is stored in a session cookie by default. It is possible to override the cookie's `maxAge` in `CookieSerializer`, but there is no way to refresh the cookie whenever the session is accessed and the cookie would simply expire after `maxAge`, even if the session is still actively used.

**Context**

We are using [Capacitor](https://capacitorjs.com) to bundle our web application as a native app. Many users are closing the application instead of pausing it/moving it to the background, only to re-open it a minute later. Every time the app is closed, all session cookies are cleared and the user is associated with a new session. Using a persistent cookie that is extended on every session access, would be an elegant solution to this issue.

Possible workarounds:
- Set a very long expiration date on the session cookie (e.g. 24h) and hope that no user uses the app around the clock
- Manually extend the session cookie in response to user actions or  a globally in a filter

theigl:
I created a draft PR that shows what changes would be necessary to support my use case.

marcusdacoregio:
Hello @theigl. I'd like to know more about your use case and what you've tried.

Have you tried to do `defaultCookieSerializer.setCookieMaxAge(Integer.MAX_VALUE);`? This way the cookie expiration time is long and you rely only on the session expiration time.

> Manually extend the session cookie in response to user actions or a globally in a filter

This can probably be achieved with a custom filter, have you tried it? I'm unsure if you want to write the `Set-Cookie` header back on every request the client makes.

theigl:
Hi @marcusdacoregio!

> Have you tried to do defaultCookieSerializer.setCookieMaxAge(Integer.MAX_VALUE);? This way the cookie expiration time is long and you rely only on the session expiration time.

This approach would work, but it feels a bit wrong. My session timeout is set to 30 minutes, so setting a persistent cookie with a very long lifetime will lead to a lot of requests for sessions that definitely do not exist anymore.

> This can probably be achieved with a custom filter, have you tried it?

I did a quick prototype but soon realized that this is not trivial to implement. The filter has to be added *after* the session repository filter and it has to handle the case where the session is created or invalidated in the same request. There is no API for retrieving cookies from the `HttpServletResponse` so  I would need to set some attribute on the request to know that the session has been created or invalidated or wrap the response and expose the cookies that have been added.

[...]

PR #3098, opened 2024-07-15
#3097 Hook for extending session cookie on access

theigl:
This PR adds an additional hook to `HttpSessionIdResolver` that is called whenever an existing session is accessed. This would make it possible to store the session ID in a persistent cookie with an expiration date that is extended every time the session is accessed.

See https://github.com/spring-projects/spring-session/issues/3097.


theigl:
@marcusdacoregio: Could you please take a look at this PR and #3097? Is this something you would consider merging?

theigl:
Closing because there are alternative ways to solve this.

Issue #3185, opened 2024-09-02
Optimize the usage of JacksonMongoSessionConverter to prevent duplicate MongoSession Document saves when a custom ObjectMapper is provided.


xiaoquanidea:
While retaining the original instantiation method of JacksonMongoSessionConverter, it should also support the ability to pass in a custom ObjectMapper, with the framework providing basic custom configurations. This approach reduces the cost for users when using a custom ObjectMapper and prevents issues related to document duplication.

While retaining the original constructor, add a new constructor with an additional copyToUse parameter. If copyToUse is set to true, a copy of the ObjectMapper will be created using its copy method, followed by basic configuration. If the user-provided ObjectMapper has already registered some modules, users can adjust the parameters to ensure that module duplication issues are avoided.

1. If response.write is actively called in loginSuccessHandler, since the native servlet response is wrapped, calling the getWriter method on the wrapped response will return an instance of OnCommittedResponseWrapper.SaveContextPrintWriter. Before this instance calls the flush method, it will invoke the onResponseCommitted method, which will internally perform the first save of the MongoSession.

2. After the logic in point 1 is completed, the process will continue to the finally block of the SessionRepositoryFilter#doFilterInternal method (where the following code is executed in the finally block: wrappedRequest.commitSession()), and this method will perform the second save of the session.

If the _id property of the Document is the same, the second call to Mongo's save will perform an update operation. When using JdkMongoSessionConverter, there will be no issue of saving the session twice because this converter actively sets the _id. However, in JacksonMongoSessionConverter, if a custom ObjectMapper is provided but the ObjectMapper's PropertyNamingStrategy is not configured and Visibility is not set, it can lead to incorrect data being stored in the database.
<img width="743" alt="企业微信截图_40878484-86d6-4744-b735-ce1eb21fef1e" src="https://github.com/user-attachments/assets/1eb4605c-0066-4c68-a2d6-9087d20d2162">
<img width="960" alt="企业微信截图_75be489a-61ed-4b61-95f6-19cd37000073" src="https://github.com/user-attachments/assets/33f7f9f9-b32b-476d-937d-80430142f2bc">

<img width="742" alt="image" src="https://github.com/user-attachments/assets/ecdb5628-aa19-4724-a8a4-0b3df622dd38">

This might confuse those unfamiliar with the Spring Security framework, as they may wonder why there are two records in Mongo.

xiaoquanidea:
[...]

PR #3186, opened 2024-09-02
Optimize the usage of JacksonMongoSessionConverter to prevent duplica…

xiaoquanidea:
…te MongoSession Document saves when a custom ObjectMapper is provided.https://github.com/spring-projects/spring-session/issues/3185


xiaoquanidea:
@vpavic

PR #3187, opened 2024-09-03
Optimize the usage of JacksonMongoSessionConverter to prevent duplica…

xiaoquanidea:
…te MongoSession Document saves when a custom ObjectMapper is provided.https://github.com/spring-projects/spring-session/issues/3185

### DIFF
--- a/spring-session-data-mongodb/src/main/java/org/springframework/session/data/mongo/JacksonMongoSessionConverter.java
+++ b/spring-session-data-mongodb/src/main/java/org/springframework/session/data/mongo/JacksonMongoSessionConverter.java
@@ -16,11 +16,6 @@
 
 package org.springframework.session.data.mongo;
 
-import java.io.IOException;
-import java.util.Collections;
-import java.util.Date;
-import java.util.HashMap;
-
 import com.fasterxml.jackson.annotation.JsonAutoDetect;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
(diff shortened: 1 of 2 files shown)


xiaoquanidea:
@vpavic

spring-session-data-mongodb/src/main/java/org/springframework/session/data/mongo/JacksonMongoSessionConverter.java
```diff
@@ -16,11 +16,6 @@
 
 package org.springframework.session.data.mongo;
 
-import java.io.IOException;
```
rwinch, line 19: Please only make changes that are necessary for this PR. Note that this breaks checkstyle. You can run `./gradlew check --continue`

rwinch (changes requested):
Thanks for the PR. Generally, please only make the changes necessary (e.g. no changing of scopes or import ordering), ensure the build passes.

Also this PR should be based off of main since it is a new feature.

spring-session-data-mongodb/src/main/java/org/springframework/session/data/mongo/JacksonMongoSessionConverter.java
xiaoquanidea: Thank you for the reminder. I have resubmitted the code as requested.

xiaoquanidea:
> Thanks for the PR. Generally, please only make the changes necessary (e.g. no changing of scopes or import ordering), ensure the build passes.
> 
> Also this PR should be based off of main since it is a new feature.

Thank you for the reminder. I have resubmitted the code as requested.

PR #3316, opened 2025-01-07
#3299 fix jdbc session special characters problem

### LINKED ISSUE #3299: Spring Session JDBC JSON format invalid escape sequence
**Describe the bug**

I want to store spring security session via JDBC in a Postgres database, and followed [those](https://docs.spring.io/spring-session/reference/configuration/jdbc.html#customize-sql-queries) instructions.

When the stored principal has an Umlaut in its name attribute (e.g. "Rüdiger"), storing `SPRING_SECURITY_CONTEXT` fails with an error like this:

> [22P02] ERROR: invalid input syntax for type json Detail: Escape sequence "\3" is invalid. Where: JSON data, line 1: {"name":"R\3...

**To Reproduce**

The SQL for inserting/updating session attributes does not seem to work with those characters.

To reproduce, try this SQL containing the encode part from the suggested SQL in the [spring-session documentation](https://docs.spring.io/spring-session/reference/configuration/jdbc.html#customize-sql-queries):

```SQL
SELECT encode('{"name":"Rüdiger"}', 'escape')::jsonb;
```

**Expected behavior**

Storing non-ASCII characters in JSON fields should be possible.

**Sample**

Try the SQL above online here:

https://onecompiler.com/postgresql/4337htpct

mindhaq:
This fixes #3299, which prevents having session attributes with special characters to be stored in a JSONB column as JSON by using a different insertion method for the JSON blob stream.

See also the documentation of [convert_from](https://www.postgresql.org/docs/17/functions-binarystring.html#id-1.5.8.11.11.2.2.2.1.1.1) method.

### DIFF
--- a/spring-session-docs/modules/ROOT/pages/configuration/jdbc.adoc
+++ b/spring-session-docs/modules/ROOT/pages/configuration/jdbc.adoc
@@ -161,7 +161,7 @@ public class QueryCustomizer
 
     private static final String UPDATE_SESSION_ATTRIBUTE_QUERY = """
 		UPDATE %TABLE_NAME%_ATTRIBUTES
-		SET ATTRIBUTE_BYTES = encode(?, 'escape')::jsonb
+		SET ATTRIBUTE_BYTES = convert_from(?, 'UTF8')::jsonb
 		WHERE SESSION_PRIMARY_ID = ?
 		AND ATTRIBUTE_NAME = ?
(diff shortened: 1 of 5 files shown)


rwinch:

I did extract out the test for the username with special characters out so that it does not impact the sample application for users trying to log in (we want a consistent username/password for those)

PR #3382, opened 2025-04-16
Clean up @ExtendWith(SpringExtension.class) usage

quaff:
1. Use `@SpringJunitConfig` or `@SpringJUnitWebConfig` instead where feasible
2. Remove it if `@SpringBootTest` present


jzheaux:
Thanks for this helpful cleanup, @quaff! This will merge into `4.0.x` once the build completes.

PR #3499, opened 2025-11-17
SQL Server DDL should use NVARCHAR to prevent Implicit Conversion

Chienlin1014:
Description
Observation and Rationale:

We observed that key string columns in the Spring Session SQL Server DDL, specifically PRINCIPAL_NAME and ATTRIBUTE_NAME, are defined using the VARCHAR data type.

This creates a type mismatch because the Microsoft JDBC Driver, by default, sends string parameters (for lookups and constraints) to SQL Server using the Unicode (NVARCHAR) data type.

Potential Impact and Optimization:

This mismatch necessitates an Implicit Conversion by SQL Server whenever these columns are queried. This structural pattern compromises the database's ability to utilize the underlying indexes (SPRING_SESSION_IX3 and SPRING_SESSION_ATTRIBUTES_PK) optimally, often leading to slower Index Scan operations.

Proposed Solution:

We propose adjusting both the PRINCIPAL_NAME and ATTRIBUTE_NAME columns to NVARCHAR. This alignment eliminates the potential for implicit conversion, ensuring consistency with the MS JDBC driver and guaranteeing optimal index utilization for high-frequency session operations.

Precedent for Structural Alignment:

This DDL adjustment addresses a consistent pattern of performance friction across the Spring ecosystem on SQL Server. For reference, analogous fixes were recently merged in other core projects:

[Spring Batch PR #5065: Fix for SQL Server DDL column definitions](https://github.com/spring-projects/spring-batch/pull/5065)

Your guidance and review are highly appreciated.


spring-session-jdbc/src/main/resources/org/springframework/session/jdbc/schema-sqlserver.sql
```diff
@@ -5,7 +5,7 @@ CREATE TABLE SPRING_SESSION (
 	LAST_ACCESS_TIME BIGINT NOT NULL,
 	MAX_INACTIVE_INTERVAL INT NOT NULL,
 	EXPIRY_TIME BIGINT NOT NULL,
-	PRINCIPAL_NAME VARCHAR(100),
+	PRINCIPAL_NAME NVARCHAR(100),
```
ronodhirSoumik: question:: using `NVARCHAR `instead `VARCHAR` will lower max length capacity
For `VARCHAR: 8k Chars` but for `NVARCHAR: 4k Chars`
will that be a concern in this case (in future)?

jzheaux:
Thank you, @Chienlin1014! This is now merged into `4.0.x`

PR #3627, opened 2026-01-22
Add HsqldbJdbcIndexedSessionRepositoryCustomizer

### LINKED ISSUE #3626: Concurrent modification of session stored in hsqldb can result in HsqlException: integrity constraint violation: unique constraint or index violation ; SPRING_SESSION_ATTRIBUTES_PK table: SPRING_SESSION_ATTRIBUTES
**Describe the bug**
Running a spring-boot web application using spring-session-jdbc: when spring-session-jdbc is configured to store the session in an hsql database, if two concurrent http requests are handled from the same user, and attempt to insert the same attribute into the session, they can fail with the following exception message: 

```
HsqlException: integrity constraint violation: unique constraint or index violation ; SPRING_SESSION_ATTRIBUTES_PK table: SPRING_SESSION_ATTRIBUTES
```

**To Reproduce**
See _Describe the bug_

**Expected behavior**
Session attributes are inserted _or updated_ similar to how they would be with an in-memory session; or if using another relational database which already has upsert query configured for session attribute, e.g. postgres and using PostgreSqlJdbcIndexedSessionRepositoryCustomizer

**Sample**
In progress.

MFAshby:
Similar to other relational databases, this configures the session attribute create query to perform an upsert in order not to fail in the case of concurrent requests modifying the same session attributes

Fixes: https://github.com/spring-projects/spring-session/issues/3626

### DIFF
--- a/spring-session-jdbc/src/integration-test/java/org/springframework/session/jdbc/HsqldbJdbcIndexedSessionRepositoryCustomizerITests.java
+++ b/spring-session-jdbc/src/integration-test/java/org/springframework/session/jdbc/HsqldbJdbcIndexedSessionRepositoryCustomizerITests.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2014-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


jzheaux:
Thanks for this — filling in the Hsqldb customizer is a nice gap to close alongside the existing MariaDB/MySQL/Oracle/Postgres/SqlServer/Db2 customizers.

One thing to fix before this can be merged: `HsqldbJdbcIndexedSessionRepositoryCustomizerITests` currently `@Bean`-registers a `PostgreSqlJdbcIndexedSessionRepositoryCustomizer`, not the `HsqldbJdbcIndexedSessionRepositoryCustomizer` this PR adds — looks like a copy/paste leftover from an existing Postgres test. As written, the test never actually exercises the new class. Could you update it to register the Hsqldb customizer so the test proves the upsert behavior for the class it's named after?

Since this adds new public API, it'll also need a milestone/main-track decision from a maintainer rather than a backport — happy to help route that once the test is fixed.

PR #3638, opened 2026-01-29
Fixup JdbcIndexedSessionRepository - multiple sessions for ID

MFAshby:
SessionResultSetExtractor#extractData should not return a list containing multiple entries for the same JdbcSession This is confusing when debugging the code.

Add a warning when >1 session is found for the same ID. This is an indicator that data is corrupt; most of the schemas prevent at the database level by adding a unique constraint to the primary ID column. If the code encounters this scenario; it should fail fast, however this might be a breaking change. A WARN level logging helps developers initially spot this situation and remedy.


jzheaux:
Thanks for tracking this down, @MFAshby.

Could you add a regression test? Right now there's no test demonstrating the duplicate-entry behavior before the fix and the corrected behavior after.

Also, I believe this could apply to older branches. Are you able to rebase onto `4.0.x`?

PR #3688, opened 2026-02-23
Documentation and Spring Boot 4 updates and fixes

jkaving:
**Documentation Fixes**
Spring [now uses Antora](https://spring.io/blog/2023/05/11/spring-framework-reference-documentation-update) for documentation, which means:
- The single-page version of the documentation is no longer available
- The documentation has a version drop-down, so we should no longer link to a specific version.\
(This means we no longer need to parse versions, so it was possible to make a bit of cleanup in\
`spring-session-docs.gradle`).

There are currently a number of broken links because of this, and I've tried to fix all of them.

**Property Changes in Spring Boot**
The Spring Session Data Redis properties have been renamed in Spring Boot 4. See [here](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-4.0-Migration-Guide#spring-session) in the migration documentation.
This updates the documentation to use these new property names.

**Spring Boot Modular Design**
I've updated the samples and the documentation to use the starters now recommended by Spring Boot 4 (see [here](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-4.0-Migration-Guide#starters)).

**Bonus fix**
While testing my changes I had problems when running the `spring-session-sample-boot-jdbc` and trying to connect with the H2 console using the JDBC URL specified in the documentation.
That's because Spring Boot will give the database a unique (random) name by default.
I've added a property to avoid that and make the `jdbc:h2:mem:testdb` URL work.

_Note: I've tried to follow the contributing guidelines and based this on the oldest impacted maintenance branch, which I think is 4.0.x._


jzheaux:
Thanks for this thorough PR, @jkaving! It is now merged into `4.0.x` and will go out with the next maintenance release.

PR #3795, opened 2026-05-30
chore: update build configuration

AdhyaksaCyber:
A Remote Code Execution (RCE) vulnerability exists in the Gradle build configuration (`build.gradle`) of `spring-session`. The build script executes shell commands derived from environment variables and system properties without sufficient sanitization, allowing an attacker to execute arbitrary code on the GitHub Actions runner during the build process.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,13 +1,13 @@
 buildscript {
-	ext {
-		releaseBuild = version.endsWith('RELEASE')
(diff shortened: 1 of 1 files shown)


jzheaux:
@AdhyaksaCyber, if you feel there is a vulnerability, will you please report it to https://spring.io/security instead of opening a PR?

Issue #1715, opened 2020-10-20
redis keys not removed from PRINCIPLE_NAME_INDEX_NAME section when user session expires. (but removed when logout request comes)


emre06ylmz:
We use spring session with redis in our spring boot app. In our usecase, most of the users do not send logout request. If logout request comes from the user, spring removes the redis keys from PRINCIPLE_NAME_INDEX_NAME and the other sections. But, if logout request does not come, when user session expires, spring removes redis keys from all other sections except PRINCIPLE_NAME_INDEX_NAME.

After some time, number of user sessions in PRINCIPLE_NAME_INDEX_NAME section, will be pile up and some performance problems occours when spring read sessions from redis.
 
When we analyze the reqeust from dynatrace, we see HGETALL, QUIT, AUTH operations. (uploaded image)
![redis](https://user-images.githubusercontent.com/4029738/96590900-40e27a00-12ef-11eb-892e-200b88076947.png)

To sum up, we need that spring removes redis keys from PRINCIPLE_NAME_INDEX_NAME section such as when logout request comes.

More information about the problem is below.
------------------------------------------------------
The Project we talk about, is designed as a library and our domain teams add and use it in their spring boot projects. I will share the versions of components.
The problem is currently, if we send logout request the session key is removed from the redis. But if we do not send a logout request and wait for the expiration, the session key is not removed from the redis. In our applications most of the time there is no logout request so after some time redis operations slow down.

To sum up with the screen capture, user U039909 has 5 sessions, you can see right of the capture. if logout reqeust come, these sessions are removed from the index.But if there is no logout reqeust these session will exist in the PRINCIPLE_INDEX_NAME_PRICPLE.
![redis2](https://user-images.githubusercontent.com/4029738/96591701-35dc1980-12f0-11eb-80b1-337578be20e6.jpg)

To be honest we do not want to use redis keyspace notificaitons because of some cases, for example application can be down while notification fired or redis could be restarted any time. 
Do you have any other suggestion except using key notificaitions? 

project informations:
spring boot : 2.0.0.RELEASE
spring: 5.0.4.RELEASE

michal-kroliczek:
It was reported quite time ago and there is no reply? :(

I reported the same https://github.com/spring-projects/spring-session/issues/1779 but I suspect it is related to redis cloud deployments - at least that is explainable but I am not expert and it would be nice to have confirmation.

eleftherias:
[...]

PR #3798, opened 2026-06-02
Clean Redis principal index during expiration cleanup

yeomin4242:
When Redis expiration events are unavailable, the expires key may be gone while the session hash remains available in the grace period. Clean the principal index during the existing expiration cleanup flow when that session is already expired.

## Summary

- clean stale Redis principal index entries from the existing Redis expiration cleanup flow
- handle the case where `sessions:expires:<sessionId>` is already gone while the session hash is still available and expired
- add a Redis static master/replica integration test that reproduces missing expiration events with `RedisStaticMasterReplicaConfiguration`

Addresses gh-1715.

## Notes

This keeps the cleanup in `RedisIndexedSessionRepository`'s expiration cleanup path. It does not change `findByIndexNameAndIndexValue`, so index lookups do not gain cleanup side effects.

The cleanup only removes the principal index entry when the session hash can still be loaded and the session is already expired. If the session hash is no longer available, this change leaves the index entry untouched.

This PR does not change `SortedSetRedisSessionExpirationStore` behavior.

### DIFF
--- a/spring-session-data-redis/src/integration-test/java/org/springframework/session/data/redis/RedisIndexedSessionRepositoryStaticMasterReplicaITests.java
+++ b/spring-session-data-redis/src/integration-test/java/org/springframework/session/data/redis/RedisIndexedSessionRepositoryStaticMasterReplicaITests.java
@@ -1,0 +1,157 @@
+/*
+ * Copyright 2014-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.session.data.redis;
+
+import java.time.Duration;
+import java.util.UUID;
+
+import com.redis.testcontainers.RedisContainer;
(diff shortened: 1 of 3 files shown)


jzheaux:
Hi, @yeomin4242, thanks for the PR. I'm not yet convinced that what's needed is for expiration store to manually intervene. I've posted a theory and some code to #1715 and we'll see what comes of it. For now, I'm blocking this PR so it doesn't accidentally get merged while that investigation continues.

yeomin4242:
Thanks, @jzheaux. I verified the two-factory configuration from #1715 in a local Redis master/replica Testcontainers setup. With `RedisStaticMasterReplicaConfiguration` for session operations and a separate master-only `RedisStandaloneConfiguration` for `RedisMessageListenerContainer`, the expiration event was received and the principal index was cleaned.

I repeated the test after temporarily removing this PR's expiration-store change, and it still passed. I agree that the repository should not manually intervene here, so I have documented the result on #1715 and will close this PR.

yeomin4242:
Closing after verifying that the two-factory configuration resolves this case without a repository change.

yeomin4242:
Thanks for investigating this.

I tested the two-connection-factory configuration described in #1715 locally: RedisStaticMasterReplicaConfiguration with REPLICA_PREFERRED for RedisIndexedSessionRepository operations, and a RedisStandaloneConfiguration pointing to the master for RedisMessageListenerContainer. I also enabled Ex keyspace notifications and registered the standard Spring Session subscriptions.

With this setup, TTL expiration removed the PRINCIPAL_NAME_INDEX_NAME membership through the existing message-listener path. The test does not invoke cleanUpExpiredSessions(), so the expiration-store fallback proposed in this PR is not involved.

This suggests that separating the Pub/Sub connection addresses the stale-index scenario I reproduced. I agree that having the expiration store perform this cleanup is likely not the right direction.

Issue #3468, opened 2025-08-28
SortedSetRedisSessionExpirationStore touches Session itself instead of its expires key


mrfilipenko:
**Describe the bug**
`SortedSetRedisSessionExpirationStore` touches session itself instead of its expires key, so no Redis key event fires.

**Expected behavior**
Since `RedisIndexedSessionRepository#onMessage` listens for key events on pattern `this.namespace + "sessions:expires:"`, `SortedSetRedisSessionExpirationStore#cleanupExpiredSessions` should touch expires keys instead of sessions themselves just like `RedisIndexedSessionRepository.MinuteBasedRedisSessionExpirationStore` or `ReactiveRedisIndexedSessionRepository` + `SortedSetReactiveRedisSessionExpirationStore` does.

I think the bug arises from the fact that `SortedSetRedisSessionExpirationStore` does not store `expires:sessionId` format in the sorted set, like the `RedisIndexedSessionRepository.MinuteBasedRedisSessionExpirationStore`, but only the `sessionId`; `SortedSetReactiveRedisSessionExpirationStore` does not store  `expires:sessionId` either, just `sessionId`, but that is compensated in `ReactiveRedisIndexedSessionRepository#cleanUpExpiredSessions` where the `sessionId` is put into the expires key format.

lsh1215:
@marcusdacoregio I looked into this and the original reporter seems inactive. I'd like to pick it up and open a PR.

marcusdacoregio:
I don't maintain this project anymore. But I reckon @rwinch would love a PR for this

lsh1215:
Thanks for the heads up. I'll open a PR and tag @rwinch.

jzheaux:
Thanks, @mrfilipenko for the report and @lsh1215 for the PR. I'll go ahead and close this so we can continue any needed discussion at #3870

PR #3870, opened 2026-07-24
SortedSetRedisSessionExpirationStore should touch the expires key

lsh1215:
`SortedSetRedisSessionExpirationStore#cleanupExpiredSessions` touches the session key (`<namespace>:sessions:<id>`) instead of the expires shadow key (`<namespace>:sessions:expires:<id>`).

`RedisIndexedSessionRepository#onMessage` only handles keyspace events whose key starts with the `sessions:expires:` prefix. Because the store touches the session key, expiration does not produce an event that matches that prefix, so `SessionDeletedEvent` and `SessionExpiredEvent` are not published for sessions tracked by this store. The principal index cleanup and the removal from the expiration sorted set run from that same `onMessage` handler, so they do not happen either.

This affects `RedisIndexedSessionRepository` when it is configured with `SortedSetRedisSessionExpirationStore`. `MinuteBasedRedisSessionExpirationStore` and the reactive `SortedSetReactiveRedisSessionExpirationStore` both touch the expires key and are not affected.

### Change

`cleanupExpiredSessions` now touches `<namespace>:sessions:expires:<id>`, which matches the prefix that `onMessage` listens on and is consistent with the other two stores. The existing private helper was renamed from `getSessionKey` to `getExpiredKey` to reflect the key it builds and to match the naming used in `RedisIndexedSessionRepository` and `ReactiveRedisIndexedSessionRepository`.

### Testing

- Updated `SortedSetRedisSessionExpirationStoreTests#cleanupExpiredSessionsThenTouchExpiredSessions` to assert that the expires key is touched. It fails before the change and passes after.
- `./gradlew :spring-session-data-redis:test` passes.
- `./gradlew :spring-session-data-redis:integrationTest --tests "*SortedSetRedisSessionExpirationStoreITests"` passes.
- checkstyle and Spring Java Format checks pass for the changed files.

Closes gh-3468

cc @rwinch

### DIFF
--- a/spring-session-data-redis/src/main/java/org/springframework/session/data/redis/SortedSetRedisSessionExpirationStore.java
+++ b/spring-session-data-redis/src/main/java/org/springframework/session/data/redis/SortedSetRedisSessionExpirationStore.java
@@ -75,8 +75,8 @@ public class SortedSetRedisSessionExpirationStore implements RedisSessionExpirat
 
 	/**
 	 * Retrieves the sessions that are expected to be expired and invoke
-	 * {@link #touch(String)} on each of the session keys, resolved via
-	 * {@link #getSessionKey(String)}.
+	 * {@link #touch(String)} on each of the expired keys, resolved via
+	 * {@link #getExpiredKey(String)}.
 	 */
 	@Override
 	public void cleanupExpiredSessions() {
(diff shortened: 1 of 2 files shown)


lsh1215:
@rwinch flagging this as @marcusdacoregio suggested on #3468.

I've retargeted this from `3.5.x` to `4.0.x` — 3.5.x left OSS support on Jun 30, so my original base was wrong. Rebased cleanly onto 4.0.x; the change itself is unchanged (one commit, two files), and the bug is present on `main` too.

`SortedSetRedisSessionExpirationStore#cleanupExpiredSessions` touches the session key instead of the `sessions:expires:` shadow key, so expiration never produces a keyspace event matching the prefix `RedisIndexedSessionRepository#onMessage` listens for — no `SessionExpiredEvent`/`SessionDeletedEvent`, and the principal-index cleanup that runs from the same handler doesn't happen either.

jzheaux:
Thanks for this PR, @lsh1215! It will merge once the build completes.

PR #3884, opened 2026-08-01
Annotate DefaultCookieSerializer.setCookiePath as @Nullable

### LINKED ISSUE #3883: DefaultCookieSerializer.setCookiePath missing @Nullable on cookiePath parameter
**Describe the bug**
`DefaultCookieSerializer.setCookiePath` doesn't accept null values. 

This is in version 4.1.0.

`setCookiePath` is the setter for the `cookiePath` field, which is annotated with `@Nullable`, but it doesn't accept null values due to the missing annotation. 

The setter documentation specifically calls out nullable values, so it is considered a valid value, but the setter input parameter doesn't mark it as `@Nullable`. Therefore passing null breaks in Kotlin 2.3.1 (likely in anything >= Kotlin 2.1).

**To Reproduce**
```
class CustomCookieSerializer: DefaultCookieSerializer() {
    override fun writeCookieValue(cookieValue: CookieSerializer.CookieValue) {
        setCookiePath(null)
    }
}
```

**Expected behavior**
Able to call `setCookiePath(null)` without compilation errors

zanarellidev:
## Summary
`DefaultCookieSerializer.setCookiePath(String)` was missing `@Nullable` even though the field is `@Nullable` and the javadoc documents null as valid. Aligns with `setSameSite(@Nullable String)`.

Fixes #3883

## Test plan
- [x] `./gradlew :spring-session-core:test --tests DefaultCookieSerializerTests` (existing coverage of `setCookiePath(null)`)

### DIFF
--- a/spring-session-core/src/main/java/org/springframework/session/web/http/DefaultCookieSerializer.java
+++ b/spring-session-core/src/main/java/org/springframework/session/web/http/DefaultCookieSerializer.java
@@ -296,7 +296,7 @@ public class DefaultCookieSerializer implements CookieSerializer {
 	 * @param cookiePath the path of the Cookie. If null, the default of the context path
 	 * will be used.
 	 */
-	public void setCookiePath(String cookiePath) {
+	public void setCookiePath(@Nullable String cookiePath) {
 		this.cookiePath = cookiePath;
 	}
 


jzheaux:
Thanks for this correction, @zanarellidev. This will merge into `4.1.x` when the build completes.

Issue #1843, opened 2021-04-21
memory leak when session save model is "always"


gabrielsky:
**Describe the bug**
version:2.4.3
when a user after logging in , RedisIndexedSessionRepository will add sessionid to redis key:
"sessionAttr:org.springframework.session.FindByIndexNameSessionRepository.PRINCIPAL_NAME_INDEX_NAME" type of set,
the question is if you set save-mode to always, the sessionId will never be remove from this set after you call request.invalidate()
```
spring.session.redis.save-mode=always
```

**To Reproduce**
1. config
```
spring.session.redis.save-mode=always
```
2. login any user:test, just create sessionId
3. logout user:test,call request.invalidate()
4. look data in redis ,the sessionId still in "sessionAttr:org.springframework.session.FindByIndexNameSessionRepository.PRINCIPAL_NAME_INDEX_NAME:test"

**Expected behavior**
when call request.invalidate()
redis remove sessionid from redis set :"sessionAttr:org.springframework.session.FindByIndexNameSessionRepository.PRINCIPAL_NAME_INDEX_NAME:test"

** other **
after set ```spring.session.redis.save-mode=always```
in Reproduce3 put all attributes to  this.delta every RedisSession instance
```
if (this.isNew || (RedisIndexedSessionRepository.this.saveMode == SaveMode.ALWAYS)) {
				getAttributeNames().forEach((attributeName) -> this.delta.put(getSessionAttrNameKey(attributeName),
						cached.getAttribute(attributeName)));
			}
```
then clean sessionId from redis set
```
	private void cleanupPrincipalIndex(RedisSession session) {
		String sessionId = session.getId();
		Map<String, String> indexes = RedisIndexedSessionRepository.this.indexResolver.resolveIndexesFor(session);
		String principal = indexes.get(PRINCIPAL_NAME_INDEX_NAME);
		if (principal != null) {
			this.sessionRedisOperations.boundSetOps(getPrincipalKey(principal)).remove(sessionId);
		}
	}
```

but at last call saveDelta() re add sessionId
```
private void saveDelta() {
	if (this.delta.isEmpty()) {
		return;
	}
	String sessionId = getId();
	getSessionBoundHashOperations(sessionId).putAll(this.delta);
	String principalSessionKey = getSessionAttrNameKey(
			FindByIndexNameSessionRepository.PRINCIPAL_NAME_INDEX_NAME);
[...]

Issue #3183, opened 2024-09-01
Redis Spring Session - Indexes have a TTL of No Limit


dreamstar-enterprises:
Is it right behaviour for when sessions expire and get removed for the Indexed keys to remain their indefinitely?

<img width="1343" alt="Screenshot 2024-09-01 at 10 09 59" src="https://github.com/user-attachments/assets/6e640214-a0f9-45bd-985c-a0bfe7f2534c">

Would these not need to be removed eiher on this.sessions.deleteAndReturn, or by Redis based on the TTL or some other mechanism?

**Describe the bug**

**To Reproduce**
Steps to reproduce the behavior.

**Expected behavior**

**Sample**

A link to a GitHub repository with a [minimal, reproducible sample](https://stackoverflow.com/help/minimal-reproducible-example).

Reports that include a sample will take priority over reports that do not.
At times, we may require a sample, so it is good to try and include a sample up front.

amontenegro:
This issue was reported here as well: https://github.com/spring-projects/spring-session/issues/1843 and Im also waiting for a response. The only way I see to remove those entries is to manually remove it if they are not listed in the active list of sessions, which is not trivial.

dreamstar-enterprises:
Hi there, if it helps I can share my entire class / code for my work around for this. It's not pretty, but I think it works. I worked on this last year, so may need a day or so to find it. Let me know!

PR #3889, opened 2026-08-13
Prevent re-adding deleted sessions to Redis principal index

seonwooj0810:
Closes gh-1843

## Root cause

`RedisSession#saveDelta()` re-adds the session id to the resolved principal's Redis index set whenever the delta contains the principal/security-context attribute key — this branch does both a removal (of the id from whatever principal it was previously indexed under) and, if a principal still resolves, an unconditional re-add.

`RedisIndexedSessionRepository#deleteById()` calls `cleanupPrincipalIndex(session)` to remove the session id from its principal's index set, then sets `maxInactiveInterval` to `Duration.ZERO` and calls `save(session)`, which invokes `saveDelta()`.

Under `SaveMode.ALWAYS`, the `RedisSession` constructor unconditionally copies every current attribute (including the principal-index attribute) into `delta`, so `saveDelta()` always takes the principal-index branch when deleting such a session. Since `deleteById()` never removes the session's own attributes (it only zeroes the TTL), `resolveIndexesFor()` still resolves the same principal, and `saveDelta()` re-adds the id right back into the set that `cleanupPrincipalIndex()` had just removed it from a few lines earlier. The principal's Redis set (`...index:...PRINCIPAL_NAME_INDEX_NAME:<principal>`) then keeps growing with ids of sessions that were explicitly invalidated (e.g. on logout), leaking memory indefinitely.

This is the mechanism behind gh-1843's original report (`SaveMode.ALWAYS` + `request.invalidate()`) and matches the reporter's own code walkthrough in the issue.

## Scope vs. related issues/PRs

This repo has a cluster of principal-index-staleness reports, but they're two distinct mechanisms:
- gh-1843 / gh-3183 (duplicate) here, plus the closed gh-2616 and gh-2869, and the still-open #3798, are about sessions that expire **naturally** and never get cleaned out of the index because the Redis keyspace expiration event was missed (event redelivery gaps, single-node pub/sub in a cluster, etc.). That's a read-time/lazy-cleanup problem in `findByIndexNameAndIndexValue`/the expiration-cleanup flow, and is a legitimately different, still-open question — I'm not attempting to resolve it here, and #3798 remains orthogonal to this change (it touches the expiration-cleanup flow; this PR touches `saveDelta()`, called from `deleteById()`).
- This PR fixes the **explicit deletion** path: a session id is removed from the index and then immediately put back by the same call, regardless of any missed-event scenario. That's what gh-1843's original reporter (and gh-3183's duplicate) actually hit.

## Change

Guard the re-add in `saveDelta()` with `RedisSession#isExpired()`. At the point `deleteById()` calls `save(session)`, `maxInactiveInterval` has just been set to `Duration.ZERO` (not a negative "never expires" value), so `isExpired()` is already `true` there — no new field or flag needed. The unconditional removal from the previous principal's set is left untouched, since removing a session id from an index is always correct regardless of expiration state; only the re-add was wrong.

## Test evidence

Added `deleteWhenSaveModeAlwaysThenPrincipalIndexNotReAdded` to `RedisIndexedSessionRepositoryTests`, which sets `SaveMode.ALWAYS`, gives the session a principal-index attribute, calls `deleteById`, and asserts the principal set's `remove` still happens while `add` is never invoked. Before this change the test fails with `TooManyActualInvocations` on `add` (the id is re-added); after the change it passes.

## Verification done

1. No in-flight PR: `gh pr list --repo spring-projects/spring-session --search "1843"` / `--search "3183"` returned nothing; the only related open PR is #3798, which (per above) addresses a different code path.
2. No active self-claim on gh-1843 itself (last comment 2021); gh-3183 (the duplicate) has no claim either.
3. Code-focused: touches only `RedisIndexedSessionRepository.java` (1 line) and its test class.
4. Confirmed the bug is still present on current `3.5.x`/`main` by reading `saveDelta()` and `deleteById()` and reproducing it with the new test (fails without the fix, passes with it).
5. No parent epic.
6. Not `waiting-for-triage` (labeled `type: bug` + `in: redis`), so the triage-comment gate doesn't apply.

`./gradlew :spring-session-data-redis:test --tests '*RedisIndexedSessionRepositoryTests*'` → 57 tests, 0 failures. `./gradlew :spring-session-data-redis:checkFormatMain :spring-session-data-redis:checkFormatTest :spring-session-data-redis:checkstyleMain :spring-session-data-redis:checkstyleTest` → all pass. Docker-backed integration tests under `src/integration-test` were not run (not required for this change, which is a pure unit-testable Redis-key-management fix with no new integration surface).

This PR was prepared with AI assistance (Claude); I reviewed the root-cause analysis and diff myself before opening it.

### DIFF
--- a/spring-session-data-redis/src/main/java/org/springframework/session/data/redis/RedisIndexedSessionRepository.java
+++ b/spring-session-data-redis/src/main/java/org/springframework/session/data/redis/RedisIndexedSessionRepository.java
@@ -913,7 +913,7 @@ public class RedisIndexedSessionRepository
 				Map<String, String> indexes = RedisIndexedSessionRepository.this.indexResolver.resolveIndexesFor(this);
 				String principal = indexes.get(PRINCIPAL_NAME_INDEX_NAME);
 				this.originalPrincipalName = principal;
-				if (principal != null) {
+				if (principal != null && !isExpired()) {
 					String principalRedisKey = getPrincipalKey(principal);
 					RedisIndexedSessionRepository.this.sessionRedisOperations.boundSetOps(principalRedisKey)
 						.add(sessionId);
--- a/spring-session-data-redis/src/test/java/org/springframework/session/data/redis/RedisIndexedSessionRepositoryTests.java
+++ b/spring-session-data-redis/src/test/java/org/springframework/session/data/redis/RedisIndexedSessionRepositoryTests.java
@@ -334,6 +334,31 @@ class RedisIndexedSessionRepositoryTests {
 		verify(this.redisOperations, never()).boundValueOps(getKey("expires:" + id));
 	}
 
+	// gh-1843
+	@Test
+	void deleteWhenSaveModeAlwaysThenPrincipalIndexNotReAdded() {
+		String principalName = "principal";
+		MapSession expected = new MapSession();
+		expected.setLastAccessedTime(Instant.now().minusSeconds(60));
+		expected.setAttribute(FindByIndexNameSessionRepository.PRINCIPAL_NAME_INDEX_NAME, principalName);
+		given(this.redisOperations.<String, Object>boundHashOps(anyString())).willReturn(this.boundHashOperations);
+		given(this.redisOperations.boundSetOps(anyString())).willReturn(this.boundSetOperations);
+		Map<String, Object> map = map(
+				RedisIndexedSessionRepository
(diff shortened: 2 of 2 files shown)


seonwooj0810:
Gentle ping on this whenever someone has time to take a look. It addresses gh-1843: under `SaveMode.ALWAYS`, `deleteById()` cleans up the principal index but the subsequent `save()` re-adds the session id through `saveDelta()`, leaving a stale entry in the index. Happy to rework the approach or add more coverage if needed.

